PR #4, opened 2025-03-11
ci: add pytest to validate flashing a corrupt image

danieldegrasse:
Add a pytest to validate flashing a corrupt image. Note that this PR is blocked until we merge the following fixes:

https://github.com/tenstorrent/luwen/pull/20
https://github.com/tenstorrent/tt-flash/pull/24

Without these changes, tt-flash will hang when trying to flash the original FW while the recovery FW is running (as we don't have a telemetry table)

The PR currently has the pytest tagged so it shouldn't run (because if it does, the card on the CI system won't be flashable)

### DIFF
--- a/app/smc/pytest/recovery.py
+++ b/app/smc/pytest/recovery.py
@@ -1,0 +1,144 @@
+#!/bin/env python3
+
+# Copyright (c) 2025 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+import logging
+import os
+import time
+import sys
+
+import pyluwen
+
+from twister_harness import DeviceAdapter
+
+# Import tt_boot_fs utilities
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "..", "..", "scripts"))
+import tt_boot_fs
+
+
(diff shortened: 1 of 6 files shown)


nxuTT (approved):

app/smc/pytest/recovery.py
```diff
@@ -0,0 +1,144 @@
...
+import os
+import time
+import sys
+
+import pyluwen
+
+from twister_harness import DeviceAdapter
+
+# Import tt_boot_fs utilities
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "..", "..", "..", "scripts"))
```
cfriedt, line 16: Would be kind of preferable to use a `Path` object (requires `from pathlib import Path` too)

```suggestion
sys.path.insert(0, Path(__file__).parents[3] / "scripts")
```

cfriedt:
Looks like this is just waiting on https://github.com/tenstorrent/tt-flash/pull/24

cfriedt (approved):

danieldegrasse:
@cfriedt will send a follow up PR to change the tag on this testcase, currently twister won't run it in CI

PR #5, opened 2025-03-11
boards: tt_blackhole/bmc: move slot1 to flash1

cfriedt:
Move the slot1 partition to flash1 (external SPI).

Signed-off-by: Chris Friedt <cfriedt@tenstorrent.com>


cfriedt:
I think it would probably be useful to just remove the fwupdate altogether now that CI is passing.

So 2 things:

1. add a pytest that checks the firmware version, modifies it somehow, and then reads it back after flashing via tt_boot_fs. Implies that mcuboot must also have successfully updated the BMC application.
2. remove the fwupdate API altogether and switch to using mcuboot API calls directly

app/bmc/src/main.c
```diff
@@ -55,29 +55,24 @@ int update_fw(void)
 	k_busy_wait(1000);
 	gpio_pin_set_dt(&reset_spi, 0);
 
-	if (IS_ENABLED(CONFIG_TT_FWUPDATE)) {
+	if (false) {
```
cfriedt, line 58: The fwupdate library used `tt_boot_fs` checksums to determine if the image needed to be updated previously. However, with moving slot1 to external spi flash, mcuboot is able to determine if the images differ without any outside assistance. There is no need to have the application manually request an upgrade.

drivers/i2c/tt_stm32_i2c_byte_poll_api.c
```diff
@@ -31,6 +31,18 @@ LOG_MODULE_REGISTER(tt_stm32_i2c_api);
 
 #include "i2c-priv.h"
 
+#define LOG_ERR_RATELIMIT(ms, ...)                                                                 \
```
danieldegrasse, line 34: Should we move this change to a separate PR?

zephyr/patches/modules/hal/stm32/use-4k-page-size-for-stm32-flash.patch
danieldegrasse, line 1: Would it be worth patching the STM32 flash driver in Zephyr, and attempting to upstream the change we make? I doubt this specific change will go upstream

danieldegrasse (commented):
Awesome work here, excited to see this go forwards. A few things:
- have we verified that `tt-flash` can seamlessly update to a new firmware using this update method? My guess here is that we will need to update MCUBoot to enable this to work
- can we add a CI test (pytest?) verifying that the BMFW is upgraded? I'd like to use the unique test ID twister generates as part of the BMFW version field, and be sure that update completes.

drivers/i2c/tt_stm32_i2c_byte_poll_api.c
cfriedt: Could yeah, it's low-hanging fruit. There are a couple of commits that are kind of tangential to the main intent, just to help debug

zephyr/patches/modules/hal/stm32/use-4k-page-size-for-stm32-flash.patch
cfriedt: It's definitely important to unfork the stm drivers. I have another branch that should get that done, but it's important to ensure that we have stable testing on p100, p100a, and p150a (at least) before making that change, and then probably also important to have it go through a couple of nightly CI runs too.

cfriedt:
> Awesome work here, excited to see this go forwards. A few things:


> - have we verified that `tt-flash` can seamlessly update to a new firmware using this update method? My guess here is that we will need to update MCUBoot to enable this to work

I'm not at all confident that tt-flash can successfully do any of this yet. In fact, I think we should have 

1. an external SPI flash driver for stm32
2. possibly also a way to have the stm32 program flash via smbus

> - can we add a CI test (pytest?) verifying that the BMFW is upgraded? I'd like to use the unique test ID twister generates as part of the BMFW version field, and be sure that update completes.

Yep, that's planned (see earlier comments)

drivers/i2c/tt_stm32_i2c_byte_poll_api.c
cfriedt: I might not even need that change outside of this PR, so will likely just remove it later.

danieldegrasse:
> > Awesome work here, excited to see this go forwards. A few things:
> 
> > * have we verified that `tt-flash` can seamlessly update to a new firmware using this update method? My guess here is that we will need to update MCUBoot to enable this to work
> 
> I'm not at all confident that tt-flash can successfully do any of this yet. In fact, I think we should have
> 
>     1. an external SPI flash driver for stm32 in OpenOCD
> 
>     2. possibly also a way to have the stm32 program flash via smbus
> 
> 
> > * can we add a CI test (pytest?) verifying that the BMFW is upgraded? I'd like to use the unique test ID twister generates as part of the BMFW version field, and be sure that update completes.
> 
> Yep, that's planned (see earlier comments)

Sounds good. I think to get this to a mergable state we need:
- some form of CI test verifying that BMFW upgrades work with this flash layout
- a test verifying the update process from legacy BMFW to the new revision. This needs to be defined because we will need to update MCUBoot on existing cards for this approach to work

cfriedt:
Def need a reliable way to reset as well.

cfriedt:
@DRvader @afongTT @danieldegrasse  - a few PRs will have to be kind of coordinated for the next big release because the bootloader will change. e.g.

* bmc slot1 partition
* mcuboot for smc
* #182
* others..

I think we should be able to do it with a tt-flash, but there might have to be some tooling needed?

PR #6, opened 2025-03-11
Add i2c auto reset

palexsonTT:
This PR adds the following:

- a cm2bm message allowing an auto reset timeout to be set per chip in BMC FW
- a cm2bm message updating telemetry heartbeat for each chip in BMC FW
- auto reset timer per chip in BMC FW
- reset reason in bmStaticInfo

Still to do:

- Add default auto reset timeout to SPI and mechanism to change it during runtime
- Add reset reason to telemetry table
- Create some test cases. Flash FW to SMC that won't respond to BMC FW, create common hang scenarios like accessing harvested cores, accessing invalid noc nodes, out of bounds accesses on real noc nodes

### DIFF
--- a/lib/tenstorrent/bh_arc/cm2bm_msg.c
+++ b/lib/tenstorrent/bh_arc/cm2bm_msg.c
@@ -125,6 +125,15 @@ void UpdateAutoRstTimeoutRequest(uint32_t timeout)
 	EnqueueCm2BmMsg(&msg);
 }
 
+void UpdateTelemHeartbeatRequest(uint32_t heartbeat)
+{
+	Cm2BmMsg msg = {
+		.msg_id = kCm2BmMsgTelemHeartbeatUpdate,
+		.data = heartbeat /* in ms */
+	};
+	EnqueueCm2BmMsg(&msg);
+}
+
 /* Report the current message and automatically acknowledge it. */
 int32_t ResetBoardByte(uint8_t *data, uint8_t size)
 {
--- a/lib/tenstorrent/bh_arc/cm2bm_msg.c
+++ b/lib/tenstorrent/bh_arc/cm2bm_msg.c
@@ -116,6 +116,15 @@ void UpdateFanSpeedRequest(uint32_t fan_speed)
 	EnqueueCm2BmMsg(&msg);
 }
 
+void UpdateAutoRstTimeoutRequest(uint32_t timeout)
+{
+	Cm2BmMsg msg = {
+		.msg_id = kCm2BmMsgIdAutoRstTimeoutUpdate,
+		.data = timeout /* in ms */
+	};
+	EnqueueCm2BmMsg(&msg);
+}
+
 /* Report the current message and automatically acknowledge it. */
 int32_t ResetBoardByte(uint8_t *data, uint8_t size)
 {
--- a/app/bmc/src/main.c
+++ b/app/bmc/src/main.c
@@ -37,6 +37,9 @@ static const struct gpio_dt_spec board_fault_led =
 	GPIO_DT_SPEC_GET_OR(DT_PATH(board_fault_led), gpios, {0});
 static const struct device *const ina228 = DEVICE_DT_GET_OR_NULL(DT_NODELABEL(ina228));
 
+// static struct k_timer auto_reset_timer;
+static int auto_reset_timeout = 0;
+
 int update_fw(void)
 {
 	/* To get here we are already running known good fw */
@@ -110,9 +113,17 @@ void process_cm2bm_message(struct bh_chip *chip)
 			bharc_smbus_word_data_write(&chip->config.arc, 0x21, 0xA5A5);
 			break;
 		case 0x3:
+			/* Update fan PWM */
 			if (IS_ENABLED(CONFIG_TT_FAN_CTRL)) {
 				set_fan_speed((uint8_t)message.data & 0xFF);
 			}
+			break;
+		case 0x4:
+			/* Update auto reset timeout */
+			auto_reset_timeout = message.data;
+			// if (auto_reset_timeout == 0) {
+			// 	k_timer_stop(&auto_reset_timer);
+			// }
 			break;
 		}
 	}
--- a/app/bmc/src/main.c
+++ b/app/bmc/src/main.c
@@ -37,9 +37,6 @@ static const struct gpio_dt_spec board_fault_led =
 	GPIO_DT_SPEC_GET_OR(DT_PATH(board_fault_led), gpios, {0});
 static const struct device *const ina228 = DEVICE_DT_GET_OR_NULL(DT_NODELABEL(ina228));
 
-// static struct k_timer auto_reset_timer;
-static int auto_reset_timeout = 0;
-
 int update_fw(void)
 {
 	/* To get here we are already running known good fw */
@@ -120,10 +117,22 @@ void process_cm2bm_message(struct bh_chip *chip)
 			break;
 		case 0x4:
 			/* Update auto reset timeout */
-			auto_reset_timeout = message.data;
-			// if (auto_reset_timeout == 0) {
-			// 	k_timer_stop(&auto_reset_timer);
-			// }
+			chip->data.auto_reset_timeout = message.data;
+			if (chip->data.auto_reset_timeout == 0) {
+				k_timer_stop(&chip->auto_reset_timer);
+			}
+			break;
+		case 0x5:
+			/* Update telemetry heartbeat */
+			if (chip->data.telemetry_heartbeat != message.data) {
+				/* Telemetry heartbeat is moving */
+				chip->data.telemetry_heartbeat = message.data;
+				if (chip->data.auto_reset_timeout != 0) {
+					k_timer_start(&chip->auto_reset_timer,
+						      K_MSEC(chip->data.auto_reset_timeout),
+						      K_NO_WAIT);
+				}
+			}
 			break;
 		}
 	}
@@ -307,8 +316,10 @@ int main(void)
 	}
 
 	/* No mechanism for getting bl version... yet */
-	bmStaticInfo static_info =
-		(bmStaticInfo){.version = 1, .bl_version = 0, .app_version = APPVERSION};
+	bmStaticInfo static_info = (bmStaticInfo){.version = 1,
+						  .bl_version = 0,
+						  .app_version = APPVERSION,
+						  .last_reset_was_automatic = false};
 
 	while (1) {
 		k_sleep(K_MSEC(20));
@@ -317,8 +328,11 @@ int main(void)
 		 */
 		ARRAY_FOR_EACH_PTR(BH_CHIPS, chip) {
 			if (chip->data.arc_just_reset) {
+				static_info.last_reset_was_automatic =
+					chip->data.last_reset_was_automatic;
 				if (bh_chip_set_static_info(chip, &static_info) == 0) {
 					chip->data.arc_just_reset = false;
+					chip->data.last_reset_was_automatic = false;
 				}
 				/* TODO: we don't have to read this per chip */
 				uint16_t max_pwr = detect_max_pwr();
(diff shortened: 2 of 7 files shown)


danieldegrasse (commented):
Thanks for the work on this- a few general thoughts:
- I think the timeout should be *opt-out*- IE on boot, the BMFW expects the CMFW to start sending heartbeat messages within a predefined timeout, or the BH will get reset. This will help us cover cases of bad CMFW flashes in the future (once we have MCUBoot)
- Pairing with this, we should have a way to disable this feature. IMO, the best option is board level init code for the SMC that turns off the timer, which we can turn off with a Kconfig. This sort of mirrors the way that we use watchdogs on a chip- the watchdog is usually disabled during boot, but you can keep it enabled by setting a Kconfig

So to summarize, I think we need the following:
- make the timeout opt out, IE always enabled by default on the BMFW side
- add code on the CMFW side to disable the timeout by default, and allow the user to opt-out of that using a Kconfig. We can then enable that KConfig in the SMC app.

cfriedt:
Technically, we could write a driver that implements the [watchdog API](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/drivers/watchdog.h) for smc, but it would really just periodically message the bmc.

If the device has `status = disabled`, we could send a message from smc to bmc to disable the timeout on the bmc.

palexsonTT:
@danieldegrasse Can you let me know if this would match the kind of behaviour you're suggesting?

- BMC initializes with a nonzero auto reset timeout value and starts the timer.
- SMC has a Kconfig option (e.g. `CONFIG_TT_AUTO_RESET`). If this is disabled, then during SMC init, it would send a message to BMC updating auto reset timeout to 0 and wouldn't send telemetry heartbeat updates. If the auto reset timeout update doesn't work, then BMC would issue the auto reset.

danieldegrasse:
> @danieldegrasse Can you let me know if this would match the kind of behaviour you're suggesting?
> 
>     * BMC initializes with a nonzero auto reset timeout value and starts the timer.
> 
>     * SMC has a Kconfig option (e.g. `CONFIG_TT_AUTO_RESET`). If this is disabled, then during SMC init, it would send a message to BMC updating auto reset timeout to 0 and wouldn't send telemetry heartbeat updates. If the auto reset timeout update doesn't work, then BMC would issue the auto reset.

Yeah, this is exactly what I'm thinking. Basically that the BMC acts like an external watchdog the SMC must disable at boot. We should provide a Kconfig to switch off the auto reset timer in the BMC firmware as well, otherwise we won't be able to debug the SMC or run some CI tests (basically we'd flash the modified BMC firmware while debugging or running tests)

lib/tenstorrent/bh_arc/cm2bm_msg.c
```diff
@@ -116,6 +116,22 @@ void UpdateFanSpeedRequest(uint32_t fan_speed)
...
+}
+
+void UpdateTelemHeartbeatRequest(uint32_t heartbeat)
+{
+	Cm2BmMsg msg = {
+		.msg_id = kCm2BmMsgTelemHeartbeatUpdate, .data = heartbeat /* in ms */
+	};
+	EnqueueCm2BmMsg(&msg);
+}
+
```
cfriedt, line 134: @palexsonTT - just trying to look at this through the lens of the watchdog API and driver model.

Assume, for now, we have a driver that implements the watchdog API, in such a way that it pretends that the stm32 i2c master is an off-chip WDT. At the top of this file, we would have something like.

```
static const struct device *const wdt0 = DEVICE_DT_GET_OR(DT_NODELABEL(wdt0), NULL);
```

If `wdt0` has `status = "disabled"` in Devicetree, then it will be `NULL`.

In Zephyr, there is basically 1 function call that we need to use to service a watchdog device (outside of initialization).

[wdt_feed(wdt)](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/drivers/watchdog.h#L240)

Then there are 3 function calls needed to set up a watchdog, reinitialize it, or disable it altogether (disabling would be handy when trying to JTAG the ARC).

1. [wdt_install_timeout](https://github.com/zephyrproject-rtos/zephyr/blob/59b62243fa6b47f1102ff6254ffac058591b73db/include/zephyr/drivers/watchdog.h#L218)
2. [wdt_setup](https://github.com/zephyrproject-rtos/zephyr/blob/59b62243fa6b47f1102ff6254ffac058591b73db/include/zephyr/drivers/watchdog.h#L162)
3. [wdt_disable](https://github.com/zephyrproject-rtos/zephyr/blob/59b62243fa6b47f1102ff6254ffac058591b73db/include/zephyr/drivers/watchdog.h#L186)

If I were to request changes to just this code so that it could use Zephyr's Watchdog API, then I would say, it looks like calling `UpdateAutoResetTimeoutRequest(0)` should disable the watchdog, and when called with a non-zero value, `UpdateAutoResetTimeoutRequest(N)` seems like it should disable the watchdog, install a timeout, and then set up the watchdog again.

`UpdateTelemHeartbeatRequest()` looks to me like it is really just feeding the watchdog, keeping the system from restarting. From that perspective, I don't think it even needs a parameter. It should just call `wdt_feed(wdt0)`.

The initial setup of `wdt0` would be completed as part of Zephyr's system initialization, so it would be done in the driver itself.

I think either @danieldegrasse or I could help you with the details. Commit-wise, 
1. add devicetree bindings for `tenstorrent,bh-watchdog`
2. add a driver for that which implements Zephyr's watchdog API (with complementary BMC code, which looks done already)
3. add a DT node to the SMC devicetree
4. add configs to `test-conf/tests/` that will support it with the `tests/drivers/watchdog/` testsuites
5. enable the watchdog DT node in the `app/smc` app.

Feel free to drop something in my calendar if you would like me to help out, or same with Daniel.

The one thing to keep in mind, is that on the SMC side, if it does not have the WDT activated, then some code in `main()` will likely need to send one message to the BMC to disable that (it should be enabled by default though on the BMC side).

palexsonTT: Okay, I'll give it a shot - and probably put something in your calendar early next week to ask questions!

afongTT:
> * I think the timeout should be _opt-out_- IE on boot, the BMFW expects the CMFW to start sending heartbeat messages within a predefined timeout, or the BH will get reset. This will help us cover cases of bad CMFW flashes in the future (once we have MCUBoot)
> * Pairing with this, we should have a way to disable this feature. IMO, the best option is board level init code for the SMC that turns off the timer, which we can turn off with a Kconfig. This sort of mirrors the way that we use watchdogs on a chip- the watchdog is usually disabled during boot, but you can keep it enabled by setting a Kconfig

I strongly prefer this to be an opt-in watchdog timer that is opted in by default by the KMD driver. This is how we did it for WH and it gives the end user some chance of getting out of a boot loop from a bad CMFW flash (e.g. they could change the KMD setting to disable the wdt). They have external mechanisms to trigger BMFW and CMFW reset (i.e. power cycling / rebooting the host system), so preventing boot loops is more important to me than having BMFW / CMFW reset automatically.

cfriedt:
> preventing boot loops is more important to me than having BMFW / CMFW reset automatically.

BMC resetting BH is critical for recovery in the case of flashing bad FW to BH (when we eventually have mcuboot managing CMFW updates).

Until then, I think it's fine to just disable the watchdog entirely, but we should write the driver in any case.

danieldegrasse:
> > * I think the timeout should be _opt-out_- IE on boot, the BMFW expects the CMFW to start sending heartbeat messages within a predefined timeout, or the BH will get reset. This will help us cover cases of bad CMFW flashes in the future (once we have MCUBoot)
> > * Pairing with this, we should have a way to disable this feature. IMO, the best option is board level init code for the SMC that turns off the timer, which we can turn off with a Kconfig. This sort of mirrors the way that we use watchdogs on a chip- the watchdog is usually disabled during boot, but you can keep it enabled by setting a Kconfig
> 
> I strongly prefer this to be an opt-in watchdog timer that is opted in by default by the KMD driver. This is how we did it for WH and it gives the end user some chance of getting out of a boot loop from a bad CMFW flash (e.g. they could change the KMD setting to disable the wdt). They have external mechanisms to trigger BMFW and CMFW reset (i.e. power cycling / rebooting the host system), so preventing boot loops is more important to me than having BMFW / CMFW reset automatically.

Once we support MCUBoot in the CMFW, the watchdog timer will be a critical part of the update strategy- we need a way to reset a hung CMFW automatically so that MCUBoot will switch to the working fallback image. Also, I'm unclear how disabling the WDT would allow a user to recover from a bad CMFW flash if the card is failing to come up on the PCIe bus? In my view once the card is up on the PCIe bus, recovery becomes pretty straightforward- until then, a user is going to need JTAG to recover the card, and the best option is almost certainly to reboot the CMFW and allow MCUBoot to swap back to a working image.

app/smc/src/main.c
```diff
@@ -66,8 +70,15 @@ int main(void)
 		}
 	}
 
+	if (!DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(wdt0)) ||
+	    (get_fw_table()->chip_limits.auto_reset_timeout == 0)) {
```
cfriedt, line 74: @palexsonTT - I guess there needs to be a `bool wdt0_disabled_by_fwtable = false`, and then inside of the `!...RECOVERY..` part, update it.

The condition at line 74 can be simplified to check the book variable.

```shell
/home/runner/work/tt-zephyr-platforms/tt-zephyr-platforms/zephyr-sdk/arc-zephyr-elf/bin/../lib/gcc/arc-zephyr-elf/12.2.0/../../../../arc-zephyr-elf/bin/ld.bfd: /home/runner/work/tt-zephyr-platforms/tt-zephyr-platforms/tt-zephyr-platforms/app/smc/src/main.c:74: undefined reference to `get_fw_table'
```

cfriedt:
@palexsonTT / @kikkiakki - can you please rebase? Sorry - there were some changes that just went in that introduced conflicts.

afongTT:
> Once we support MCUBoot in the CMFW, the watchdog timer will be a critical part of the update strategy- we need a way to reset a hung CMFW automatically so that MCUBoot will switch to the working fallback image. Also, I'm unclear how disabling the WDT would allow a user to recover from a bad CMFW flash if the card is failing to come up on the PCIe bus? In my view once the card is up on the PCIe bus, recovery becomes pretty straightforward- until then, a user is going to need JTAG to recover the card, and the best option is almost certainly to reboot the CMFW and allow MCUBoot to swap back to a working image.

My concern about recovery with a WDT is for the case where CMFW finishes PCIe init, but doesn't kick the watchdog for some other reason. This would get us into a watchdog loop which would have been otherwise recoverable over PCIe.

I guess with a proper MCUBoot A/B scheme, we should never get stuck in the scenario I just described.

Is my understanding correct that even if we don't enable the WDT by default right now, we can still opt into the WDT later on while CMFW is running? (e.g. if we added a message from host)

cfriedt:
This is a tricky one:
* when watchdog is disabled on SMC, SMC will need to somehow inform BMC to _not_ trigger a reset
* we will definitely need tests (system tests with pytest)
* tests will need to cover both cases - when the watchdog is enabled _and_ when the watchdog is disabled
* from the host, how do we determine that the SMC has reset due to WD timeout?
  * this will be a crucial bit necessary for both tests
  * there must be some way to query SMC uptime from the host
* WD timeout will probably need some tuning to account for the time it takes to reflash, rescan pcie, etc

danieldegrasse:
Latest update includes the following:
- opt-in watchdog (enabled by sending an ARC message to set the watchdog timeout duration)
- pytest to verify watchdog resets the SMC correctly
- DMC support for recording PC if watchdog expires, for debugging fault conditions

cfriedt (approved):
This looks fine to me and I am ok getting this in the release. If the watchdog isn't enabled, it's business as usual for firmware, but at least that simplifies testing a bit.

Before the 18.6.0 release on June 30 we should do a bit more extensive testing so that we can enable it by default and so that we have all of the other bases covered (to avoid a possible WDT loop). Whenever we do that, I think we should consider a relatively long default timeout value.

PR #8, opened 2025-03-11
doc: add release process documentation

cfriedt:
Add a rough document describing the steps in the release process.

This is a mashup of a couple of different release process docs.

https://docs.mcuboot.com/release.html
https://docs.zephyrproject.org/latest/project/release_process.html#tagging

### DIFF
--- a/doc/release.md
+++ b/doc/release.md
@@ -1,0 +1,108 @@
+# Release Process
+
+This page describes the release process used by Tenstorrent firmware built upon the Zephyr
+Real-Time Operating System (a.k.a. `tt-zephyr-platforms`).
+
+## Version numbering
+
+`tt-zephyr-platforms` firmware uses
+[semantic versioning](https://en.wikipedia.org/wiki/Software_versioning) where version numbers
+follow a `MAJOR.MINOR.PATCH` format.
+
+1. `MAJOR` version when there are incompatible API changes.
+2. `MINOR` version when new functionalities were added in a
+   backward-compatible manner.
+3. `PATCH` version when there are backward-compatible bug fixes.
+4. `EXTRAVERSION` release-candidate suffix (e.g. `rc1`).
+
+We add pre-release tags using the format `MAJOR.MINOR.PATCH-EXTRAVERSION`.
+
+> [!NOTE]
+> Current firmware bundle files include a fourth version field for experimental builds. For
+> official releases, the fourth version field is always zero.
+
+## Release candidates
+
+Before each release, tags are made (see below) for at least one release candidate (`a.b.c-rc1`,
+followed by `a.b.c-rc2` and the subsequent release candidates, followed by the official `a.b.c`
+release). The intent is to freeze the code and allow testing.
+
+During the time between the `rc1` and the final release, the only changes that should be merged into
+the `main` branch are those to fix bugs found in the release candidate, to correct documentation,
+or to make other cosmetic fixes.
+
+## Release Procedure
+
+The following steps are required to be followed by firmware release engineers when creating a new
+`tt-zephyr-platforms` release.
+
+### Release Checklist
+
+Refer to the previous release checklist for steps required before tagging release candidates and
+final releases.
+
+### Tagging
+
+> [!NOTE]
+> This section uses a fake release version, `v1.2.3`, or `v1.2.3-rc1` as an example. Replace with
+> the appropriate release candidate or final release version.
+
+> [!IMPORTANT]
+> Any changes to `app/*/VERSION` files is required to take place prior to the release process.
+
+1. Update the version variables in the bundle version file `VERSION` file located in the root of
+    the Git repository to match the version for this release. The `EXTRAVERSION` variable is used
+    to identify release candidates. It is left empty for final releases.
+
+    ```
+    EXTRAVERSION = rc1
+    ```
+
+2. Post a PR with the updated `VERSION` file using `release: 1.2.3-rc1` or `release: 1.2.3` as
+    the commit subject. Merge the PR after CI is successful.
+
+3. Tag and push the version, using an annotated tag:
+
+    ```shell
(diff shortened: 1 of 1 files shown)


doc/release.md
```diff
@@ -0,0 +1,100 @@
...
+## Version numbering
+
+`tt-zephyr-platforms` firmware uses
+[semantic versioning](https://en.wikipedia.org/wiki/Software_versioning) where version numbers
+follow a `MAJOR.MINOR.PATCH` format.
+
+1. `MAJOR` version when there are incompatible API changes.
+2. `MINOR` version when new functionalities were added in a
+   backward-compatible manner.
+3. `PATCH` version when there are backward-compatible bug fixes.
```
TTDRosen, line 15: nit: I think the EXTRAVERSION field should be introduced here, so it's clear how rc1 gets added to the version instead of requiring the reader to hit the longer explication later.

doc/release.md
```diff
@@ -0,0 +1,100 @@
...
+    the commit subject. Merge the PR after CI is successful.
+
+3. Tag and push the version, using an annotated tag:
+
+    ```shell
+    git tag -s -m "tt-zephyr-platforms 1.2.3-rc1" v1.2.3-rc1
+    ```
+
+4. Verify that the tag has been signed correctly, git show for the tag must contain a signature
+    (look for the `BEGIN PGP SIGNATURE` or `BEGIN SSH SIGNATURE` marker in the output):
```
TTDRosen, line 67: nit: I think how to add the signiture if it doesn't exist should be included in a link here

cfriedt: Like instructions for setting up PGP keys?

TTDRosen: Yeah, I was thinking of just linking to the docs https://docs.github.com/en/authentication/managing-commit-signature-verification/telling-git-about-your-signing-key or something better. Just to give a resolution if someone doesn't have it setup.

doc/release.md
cfriedt: Fixed!

doc/release.md
cfriedt: Fixed!

doc/release.md
```diff
@@ -0,0 +1,103 @@
...
+6. Find the generated
+    [draft release](https://github.com/tenstorrent/tt-zephyr-platforms/releases), edit the release
+    notes, and publish the release.
+
+7. Announce the release via official Tenstorrent channels and provide a link to the
+    GitHub release page.
+
+### Publishing a Combined Firmware Bundle to `tt-firmware`
+
+Find the `.fwbundle` file and combine it with closed-source firmware versions for older Tenstorrent
```
danieldegrasse, line 93: Long term, would it make sense to put the closed source firmware versions as binaries in this repo? That way we can generate releases directly from here. Another option would be to use west's blobs support to pull in the binaries from another URL. Either way, I think it would make sense to get to the point where we can create finalized FW bundles directly from this repo.

doc/release.md
```diff
@@ -0,0 +1,103 @@
...
+products.
+
+```shell
+scripts/tt_boot_fs.py fwbundle combine -o fw_pack-MAJOR.MINOR.PATCH.0.fwbundle \
```
danieldegrasse, line 97: I'm assuming this is a planned addition to the `tt_boot_fs` script?

doc/release.md
cfriedt: I agree, it would be nice to release final firmware bundles here for all products, but our official releases are still published on tt-firmware.

Ideally, long term, we should have open source firmware for older products in this repo as well.

Short term, a bit of automation and maybe a deploy hook. Our older FW does not get updated very frequently, so it shouldn't be terrible to manually bundle them either.

doc/release.md
cfriedt: Yes - it doesn't exist right now, but it should be (at least theoretically) pretty straight forward.

doc/release.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Release Checklist
+
+Refer to the previous release checklist for steps required before tagging release candidates and
+final releases.
+
+### Tagging
+
+> [!NOTE]
+> This section uses a fake release version, `v1.2.3`, or `v1.2.3-rc1` as an example. Replease with
```
afongTT, line 47:
```suggestion
> This section uses a fake release version, `v1.2.3`, or `v1.2.3-rc1` as an example. Replace with
```

doc/release.md
```diff
@@ -0,0 +1,103 @@
...
+> the appropriate release candidate or final release version.
+
+1. Update the version variables in the bundle version file `VERSION` file located in the root of
+    the Git repository to match the version for this release. The `EXTRAVERSION` variable is used
+    to identify release candidates. It is left empty for final releases.
```
afongTT, line 52: Curious if anyone has an optinion on what to do for SMC / BMC VERSION files?

cfriedt: I think we should keep them for consistance.

afongTT: Should we up rev them both at the same time the rest of this release process is happening?

I do prefer having individual versions for those FWs, but I'm not sure how / if we would ever release / up rev them out of sync.

cfriedt: Good point.

It might be quite rare, but it's possible that code in this repository might not have changed but closed-source firmware for older products might have changed. So in that case, BMC and SMC firmware versions would presumably remain the same.

doc/release.md
cfriedt: Fixed in https://github.com/tenstorrent/tt-zephyr-platforms/pull/21

doc/release.md
cfriedt: Added an "IMPORTANT" note that individual app version must be updated prior to the release process.

cfriedt:
We can always update this doc as needed. No sense in waiting to merge.

PR #3, opened 2025-03-11
lib: tenstorrent: bh_arc: remove spi controller code, use flash driver

danieldegrasse:
Remove the SPI eeprom driver code, and use the dw_flash driver. This
simply removes some duplicate code, and relies directly on the dw flash
driver for flash operations.

PR #10, opened 2025-03-11
drivers: spi_dw_flash: read performance optimizations

danieldegrasse:
When reading in 4 byte aligned increments, we can optimize the performance of the SPI DW driver by using a data frame size of 32 bytes (this optimization is also possible for 16 byte frames if desired).

Implement this performance optimization within the SPI DW flash driver. In addition, raise the clock frequency of the SPI controller, as we can now read faster without FIFO overruns.

Add a flash performance sample, to help validate the performance of the SPI DW driver (and test for FIFO overflows)

Output of flash performance sample without this change (SPI clocked at 2MHz to pass read without FIFO overflow)
```
*** Booting Zephyr OS build v4.1.0-rc3 ***
Erasing 8 pages at 0x2000000
Erase took 116 ms
Write of 32768 bytes took 257 ms
Read of 32768 bytes took 15 ms
Flash performance test complete
```

Output of flash performance sample with this change (SPI clocked at 9MHz):
```
*** Booting Zephyr OS build v4.1.0-rc3 ***
Erasing 8 pages at 0x2000000
Erase took 116 ms
Write of 32768 bytes took 222 ms
Read of 32768 bytes took 3 ms
Flash performance test complete
```


danieldegrasse:
This optimization is primarily useful on the read side, since we will need fast read support for MCUBoot to function well

samples/drivers/flash/performance/sample.yaml
```diff
@@ -0,0 +1,12 @@
...
+  name: Flash Performance Test
+common:
+  tags: smoke
+  integration_platforms:
+    - tt_blackhole@p100/tt_blackhole/smc
+  harness: console
+  harness_config:
+    type: one_line
+    regex:
+      - "Flash performance test complete"
```
cfriedt, line 12: Maybe something like this?
```suggestion
      - "Flash performance test complete"
tests:
  sample.drivers.flash.performance: {}
```

danieldegrasse: It does seem like we need that- added.

cfriedt (approved):

nxuTT (approved):
I'm good with #10, as long as you retest #3.

danieldegrasse:
> I'm good with https://github.com/tenstorrent/tt-zephyr-platforms/pull/10, as long as you retest https://github.com/tenstorrent/tt-zephyr-platforms/pull/3. We used to have issues making octal mode DDR/DQS works with frame size 32, that why we only use frame size 8. But you might be able to fix it.

Great- I'll merge this first, as it won't affect the core app. Then can retest #3 and validate octal/quad mode in the bootrom there.

PR #11, opened 2025-03-11
tenstorrent: bh_chip: add pgood fault interrupts

palexsonTT:
Add an interrupt to the pgood gpio to follow the board power fault procedure when pgood drops.


lib/tenstorrent/bh_chip/bh_chip.c
```diff
@@ -147,7 +147,55 @@ int therm_trip_gpio_setup(struct bh_chip *chip)
...
+
+void pgood_fault_detected(const struct device *dev, struct gpio_callback *cb, uint32_t pins)
+{
+	static const struct gpio_dt_spec board_fault_led =
+		GPIO_DT_SPEC_GET_OR(DT_PATH(board_fault_led), gpios, {0});
+	struct bh_chip *chip = CONTAINER_OF(cb, struct bh_chip, pgood_cb);
+
+	/* Assert board fault */
+	gpio_pin_set_dt(&board_fault_led, 1);
+	/* Report over SMBus - to add later */
```
danieldegrasse, line 163: Do we want to merge this without SMBUS reporting, or is this PR more a WIP right now?

lib/tenstorrent/bh_chip/bh_chip.c
```diff
@@ -147,7 +147,55 @@ int therm_trip_gpio_setup(struct bh_chip *chip)
...
+	/* Assert ASIC reset */
+	bh_chip_assert_asic_reset(chip);
+	/* Once board fault comes back up */
+	while (!gpio_pin_get_dt(&board_fault_led)) {
+	}
```
cfriedt, line 168: Bit of an ask here, but can we avoid busy-looping inside of an ISR? This is a dangerously slippery slope, because it means that no threads will be scheduled at all and that the BMC will be unresponsive (outside of nested interrupts with higher priority).

If anything, what I would suggest doing is issuing events to some queue and then having a state machine e.g. in the `main` thread, dequeue events and act upon them.

https://docs.zephyrproject.org/latest/kernel/services/synchronization/events.html

cfriedt (changes requested):
We should definitely avoid busy-looping in ISR context.

lib/tenstorrent/bh_chip/bh_chip.c
palexsonTT: I'll look into it, thanks Chris!

cfriedt:
@palexsonTT - this looks like it should compile. You might just need to edit 

```
Exception: bmfw padto value 45056 is < the binary size 46458
```

You would need to increase this line for the appropriate board revisions.

https://github.com/tenstorrent/tt-zephyr-platforms/blob/31a7a5abff816f6b474cfa7867506d348a54890d/boards/tenstorrent/tt_blackhole/bootfs/p100-bootfs.yaml#L44

lib/tenstorrent/bh_chip/bh_chip.c
```diff
@@ -145,9 +147,82 @@ int therm_trip_gpio_setup(struct bh_chip *chip)
...
+}
+
+void pgood_fault_work_handler(struct k_work *work)
+{
+	struct bh_chip_data *data = CONTAINER_OF(work, struct bh_chip_data, pgood_fault_worker);
+	struct bh_chip *chip = CONTAINER_OF(data, struct bh_chip, data);
+	static const struct gpio_dt_spec board_fault_led =
+		GPIO_DT_SPEC_GET_OR(DT_PATH(board_fault_led), gpios, {0});
+
+	k_sem_reset(&chip->data.pgood_high_sem);
```
cfriedt, line 164: What about doing this in the callback:

* try to take a semaphore with `K_NO_WAIT`
* if the return value is 0 then all is good
* otherwise if it's `EAGAIN`, just return - it means a reset is already in progress
* schedule delayed work for 50ms
* in the delayed work handler, give the semaphore and do the pgood check below where you assert the board fault, etc

lib/tenstorrent/bh_chip/bh_chip.c
```diff
@@ -132,7 +133,51 @@ int therm_trip_gpio_setup(struct bh_chip *chip)
...
+	chip->data.pgood_fall_triggered = true;
+	bm_event_post(WAKE_BM_MAIN_LOOP);
+}
+
+void pgood_rise_detected(const struct device *dev, struct gpio_callback *cb, uint32_t pins)
+{
+	struct bh_chip *chip = CONTAINER_OF(cb, struct bh_chip, pgood_rise_cb);
+
+	chip->data.pgood_rise_triggered = true;
+	bm_event_post(WAKE_BM_MAIN_LOOP);
```
palexsonTT, line 154: Is `bh_chip_cancel_bus_transfer_set(chip);` needed here and `bh_chip_cancel_bus_transfer_clear(chip);` in the handler?

cfriedt: @nxuTT / @TTDRosen / @danieldegrasse - do you have opinions?

lib/tenstorrent/bh_chip/bh_chip.c
```diff
@@ -132,7 +133,51 @@ int therm_trip_gpio_setup(struct bh_chip *chip)
...
+			   BIT(chip->config.pgood.pin));
+	ret = gpio_add_callback_dt(&chip->config.pgood, &chip->pgood_fall_cb);
+
+	ret = gpio_pin_interrupt_configure_dt(&chip->config.pgood, GPIO_INT_EDGE_TO_ACTIVE);
+	if (ret != 0) {
+		return ret;
+	}
+	gpio_init_callback(&chip->pgood_rise_cb, pgood_rise_detected,
+			   BIT(chip->config.pgood.pin));
+	ret = gpio_add_callback_dt(&chip->config.pgood, &chip->pgood_fall_cb);
```
cfriedt, line 180: Just thinking it would be a bit safer to initialize callbacks and add them to the driver prior to enabling interrupts. Also, need to configure interrupts once, since the second configuration would overwrite the first.



```suggestion
	LOG_MODULE_REGISTER(bh_chip, CONFIG_TT_BH_CHIP_LOG_LEVEL);
	/* ^^ at the top of this file, and would need a logging entry in lib/tenstorrent/bh_chip/Kconfig */

	gpio_init_callback(&chip->pgood_fall_cb, pgood_fall_detected,
			   BIT(chip->config.pgood.pin));
	ret = gpio_add_callback_dt(&chip->config.pgood, &chip->pgood_fall_cb);
	if (ret != 0) {
		LOG_ERR("%s() failed: %d", "gpio_add_callback_dt", ret);
		return ret;
	}

	gpio_init_callback(&chip->pgood_rise_cb, pgood_rise_detected,
			   BIT(chip->config.pgood.pin));
	ret = gpio_add_callback_dt(&chip->config.pgood, &chip->pgood_fall_cb);
	if (ret != 0) {
		LOG_ERR("%s() failed: %d", "gpio_pin_interrupt_configure_dt", ret);
		return ret;
	}
	
	ret = gpio_pin_interrupt_configure_dt(&chip->config.pgood, GPIO_INT_EDGE_BOTH);
	if (ret != 0) {
		LOG_ERR("%s() failed: %d", "gpio_pin_interrupt_configure_dt", ret);
		return ret;
	}
```

include/tenstorrent/bh_chip.h
```diff
@@ -55,12 +55,19 @@ struct bh_chip_data {
 
 	/* notify the main thread to handle therm trip */
 	bool therm_trip_triggered;
+
+	/* notify the main thread to handle pgood events */
+	bool pgood_fall_triggered;
+	bool pgood_rise_triggered;
```
danieldegrasse, line 61: Nit- these two should probably be `volatile` since we are writing to them in an ISR. Unlikely to break anything though

lib/tenstorrent/bh_chip/bh_chip.c
```diff
@@ -126,13 +127,47 @@ int therm_trip_gpio_setup(struct bh_chip *chip)
...
+
+	return ret;
+}
+
+void pgood_change_detected(const struct device *dev, struct gpio_callback *cb, uint32_t pins)
+{
+	struct bh_chip *chip = CONTAINER_OF(cb, struct bh_chip, pgood_cb);
+
+	/* Sample PGOOD to see if it rose or fell */
+	/* TODO: is this reliable enough? */
```
danieldegrasse, line 154: I think this ok- we could also set the interrupt only for PGOOD falling, then set an interrupt for PGOOD rising only once PGOOD had actually fallen

danieldegrasse (approved):
Changes look go to me. Longer term I think we should consider splitting the BMFW tasks out into workqueue items to handle each event- as this stands it's fine though.

danieldegrasse:
@cfriedt can you take one more look since you engaged here before?

cfriedt:
> @cfriedt can you take one more look since you engaged here before?

Yeah, absolutely.

It's definitely good.

I would say the only thing missing here are tests.

The code in `main()` is not really designed to be testable though, so it might be a bit of a steep ask, and this is one of those fault or error conditions that is difficult to test using hardware, but super easy to test in simulation (e.g. with gpio_emul). It needs to be in a library though, so that, e.g. the loop in `main()` can be tested in isolation.

What would you rather do, @palexsonTT ?

kikkiakki:
> What would you rather do, @palexsonTT ? Commit it and let someone else fix / test the code, or write tests for the unit?
> 
> I would say, the code you added is fine, but this whole main.c is kind of in rough shape, so there is no shame in kicking the can.

I may as well clean it up and write some tests because it's already on my plate.

cfriedt (approved):
LGTM - thanks for the extra effort of adding tests @palexsonTT @kikkiakki  🙏

PR #15, opened 2025-03-12
tenstorrent: bh_arc: add message to re-initialize Tensix programming

afongTT:
This message must be called by our test infra after it resets all the Tensix.


alewycky-tenstorrent (approved):

lib/tenstorrent/bh_arc/init.c
```diff
@@ -264,6 +265,30 @@ static void EthInit(void)
...
+		EnableTensixCG();
+	}
+	if (get_fw_table()->feature_enable.noc_translation_en) {
+		InitNocTranslationFromHarvesting();
+	}
+
+	return 0;
+}
+
+REGISTER_MESSAGE(MSG_TYPE_REINIT_TENSIX, ReinitTensix);
```
cfriedt, line 290: The build fails when building recovery firmware due to missing symbols. Repro with
```shell
west build -p auto -b tt_blackhole/tt_blackhole/smc app/smc -- -DCONFIG_TT_SMC_RECOVERY=y
```

This should fix it.
```suggestion
#ifndef CONFIG_TT_SMC_RECOVERY
/**
 * @brief Redo Tensix init that gets cleared on Tensix reset
 *
 * This includes all NOC programming and any programming within the tile.
 */
static uint8_t ReinitTensix(uint32_t msg_code, const struct request *req, struct response *rsp)
{
	ClearNocTranslation();
	/* We technically don't have to re-program the entire NOC (only the Tensix NOC portions),
	 * but it's simpler to reuse the same functions to re-program all of it.
	 */
	NocInit();
	if (get_fw_table()->feature_enable.cg_en) {
		EnableTensixCG();
	}
	if (get_fw_table()->feature_enable.noc_translation_en) {
		InitNocTranslationFromHarvesting();
	}

	return 0;
}

REGISTER_MESSAGE(MSG_TYPE_REINIT_TENSIX, ReinitTensix);
#endif
```

cfriedt (commented):
minor fix should do it.

lib/tenstorrent/bh_arc/init.c
afongTT: Thanks, trying again.

afongTT:
Moved to draft because I found that this unsets AXI_ENABLE in GDDR tiles.

cfriedt:
> Moved to draft because I found that this breaks DRAM test.

Is that test in our CI? Would be good as a regression check (and I think we'll need it for bist)

afongTT:
> > Moved to draft because I found that this breaks DRAM test.
> 
> Is that test in our CI? Would be good as a regression check (and I think we'll need it for bist)

We don't have the test in this CI. It would be good to have some version of the test in this repo's CI, but it might be hard to move over as-is.

afongTT:
> Moved to draft because I found that this breaks DRAM test.

Fixed the issue causing breakage in the "clear translate_en" commit.

cfriedt:
@afongTT - probably best to squash the last two commits, for bisectability (although I think all tests on GH were almost passing before too)

afongTT:
> @afongTT - probably best to squash the last two commits, for bisectability (although I think all tests on GH were almost passing before too)

Squashed it

alewycky-tenstorrent (approved):

cfriedt (approved):

PR #16, opened 2025-03-12
ci: enable matrix for p100a support

danieldegrasse:
Enable matrix for P100a support in CI.


danieldegrasse:
This PR also includes a fix for the SPI DW Flash driver on P100a, since the CI being turned on exposed a bug

cfriedt:
Something like this would be kind of cool too, because then we can build the smc app for other `BOARD`s as well, or skip it if it doesn't apply.

```shell
          case "${{ matrix.board }}" in
            p100) BOARD=tt_blackhole@p100/tt_blackhole/smc;;
            p100a) BOARD=tt_blackhole@p100a/tt_blackhole/smc;;
            p150a) BOARD=tt_blackhole@p150a/tt_blackhole/smc;;
            *) echo "Skipping: ${{ matrix.board }}"; exit 0;
          esac
```

cfriedt (approved):

PR #19, opened 2025-03-13
ci: simplify prepare-zephyr reusable workflow

cfriedt:
* move manifest group filter to manifest file
* remove west update step, since it duplicates zephyr action
* move local python deps to scripts/requirements.txt


TTDRosen (approved):
lgtm :)

PR #21, opened 2025-03-13
scripts: add manual release script

cfriedt:
* scripts: tt_boot_fs: add support for combining fwbundle files
* add manual release script

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,4 @@ Ruff.txt
 SphinxLint.txt
 TextEncoding.txt
 YAMLLint.txt
+*.fwbundle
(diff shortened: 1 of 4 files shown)


TTDRosen (commented):
So much easier 🎉

cfriedt:
I might add the "combine" subcommand to `tt_boot_fs.py` too. That would be handy.

danieldegrasse (approved):

TTDRosen (approved):
We will have to remember to add p150{b,c} before this can fully replace the standard flow

cfriedt:
> We will have to remember to add p150{b,c} before this can fully replace the standard flow

The manual release script should mainly be used to package up release candidates.

For final releases, we should have CI generate a `.fwbundle` and other artifacts when a tag is pushed.

PR #22, opened 2025-03-14
tenstorrent: qsfp_dd: enable active qsfp-dd on p150a

palexsonTT:
Add qsfp_dd library that enables I2C communication with GPIO expanders connected to several control and status pins on QSFP-DD modules.

Enable QSFP-DD high power mode on P150A.

### DIFF
--- a/include/tenstorrent/qsfp_dd.h
+++ b/include/tenstorrent/qsfp_dd.h
@@ -1,0 +1,18 @@
+/*
+ * Copyright (c) 2025 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define GPIO_XP_REG_INPUT_PORT 0x0 /* Read input values */
+#define GPIO_XP_REG_OUTPUT_PORT 0x1 /* Read/write output values */
+#define GPIO_XP_REG_POLARITY_INV 0x2
+#define GPIO_XP_REG_CONFIG 0x3 /* Configure GPIOs, output = 0 and input = 1 */
+
+/**
+ * @brief Set up GPIOs on GPIO expander to enable high power mode on QSFP module
+ *
+ * @retval 0 on success.
+ * @retval -EIO	if an I/O error occurs.
+ */
+int enable_active_qsfp_dd(void);
--- a/lib/tenstorrent/qsfp_dd/qsfp_dd.c
+++ b/lib/tenstorrent/qsfp_dd/qsfp_dd.c
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2025 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <tenstorrent/qsfp_dd.h>
+
+#include <zephyr/drivers/gpio.h>
+#include <zephyr/drivers/i2c.h>
+
+static const struct device *const i2c3 = DEVICE_DT_GET_OR_NULL(DT_NODELABEL(i2c3));
+static const struct gpio_dt_spec mcu_conn_i2c_en =
+	GPIO_DT_SPEC_GET_OR(DT_PATH(mcu_conn_i2c_en), gpios, {0});
+
+uint8_t gpio_xp_addrs[] = {0x38, 0x39, 0x3a, 0x3b};
+
+int gpio_xp_set_config(uint8_t addr, uint8_t val)
+{
+	uint8_t cmd_data[2] = {GPIO_XP_REG_CONFIG, val};
+
+	return i2c_write(i2c3, cmd_data, 2, addr);
+}
+
+int gpio_xp_set_output(uint8_t addr, uint8_t val)
+{
+	uint8_t cmd_data[2] = {GPIO_XP_REG_OUTPUT_PORT, val};
+
+	return i2c_write(i2c3, cmd_data, 2, addr);
+}
+
+int enable_active_qsfp_dd(void)
+{
+	int ret;
+
+	/* Enable communication with GPIO expander */
+	ret = gpio_pin_configure_dt(&mcu_conn_i2c_en, GPIO_OUTPUT_ACTIVE);
+	if (ret != 0) {
+		return ret;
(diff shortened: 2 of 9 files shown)


lib/tenstorrent/qsfp_dd/qsfp_dd.c
cfriedt, line 1: This looks good as a proof of concept. 

Can we convert it to a GPIO driver? They are are pretty simple to make. An example is [gpio_pcal64xxa.c](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/gpio/gpio_pcal64xxa.c).

```shell
zephyr % find . -name '*expander*'
./dts/bindings/test/vnd,gpio-expander-common.yaml
./dts/bindings/test/vnd,gpio-expander-spi.yaml
./dts/bindings/test/vnd,gpio-expander-i2c.yaml
./boards/nordic/nrf9151dk/dts/nrf9151dk_leds_on_io_expander.dtsi
./boards/nordic/nrf9151dk/dts/nrf9151dk_buttons_on_io_expander.dtsi
./boards/nordic/nrf9160dk/dts/nrf9160/nrf9160dk_buttons_on_io_expander.dtsi
./boards/nordic/nrf9160dk/dts/nrf9160/nrf9160dk_leds_on_io_expander.dtsi
./boards/nordic/nrf9160dk/dts/nrf52840/nrf9160dk_buttons_on_io_expander.dtsi
./boards/nordic/nrf9160dk/dts/nrf52840/nrf9160dk_leds_on_io_expander.dtsi
./boards/nordic/nrf9161dk/dts/nrf9161dk_buttons_on_io_expander.dtsi
./boards/nordic/nrf9161dk/dts/nrf9161dk_leds_on_io_expander.dtsi
```

cfriedt (changes requested):
Looks good - I think we should aim to add Zephyr drivers going forward rather than library drivers though.

lib/tenstorrent/qsfp_dd/qsfp_dd.c
palexsonTT: 🫡 I will driverize. I think it should be super simple

cfriedt:
Just wanted to follow-up with this, and apparently the `ti,tca9554a` is basically identical to the `nxp,pca9554a` and the latter already has a Zephyr driver upstream. So writing a driver is not needed. All that is needed is adding the nodes to Devicetree.

lib/tenstorrent/qsfp_dd/qsfp_dd.c
cfriedt: @palexsonTT - it was significantly simpler. The `nxp,pca95xx` will do the trick.

Added that in #172, so this PR effectively only needs to add code in `app/bmc/src/main.c`.

You will likely need to check which board revision we are working with though, because the p100 has some differences in terms of what is connected to which expander and which pin.

My guess is that it's safe to call something like `enable_active_qsfp_dd()` if working with `p100a`, `p150*`, `p300*`.

cfriedt:
@palexsonTT @kikkiakki - you should be good to rebase this on `main` shortly and use `gpioxN` 

See #172

kikkiakki:
Going to utilize the gpioxN devices for this, but also disable by default for now because p150 cards officially support only passive QSFP.

boards/tenstorrent/tt_blackhole/support/bmc-reset-vars.cfg
cfriedt, line 1: woops - this file shouldn't be checked-in to git history

include/tenstorrent/qsfp_dd.h
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (c) 2025 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#define GPIO_XP_REG_INPUT_PORT 0x0 /* Read input values */
+#define GPIO_XP_REG_OUTPUT_PORT 0x1 /* Read/write output values */
+#define GPIO_XP_REG_POLARITY_INV 0x2
+#define GPIO_XP_REG_CONFIG 0x3 /* Configure GPIOs, output = 0 and input = 1 */
```
cfriedt, line 10: Likely these can be removed

boards/tenstorrent/tt_blackhole/support/bmc-reset-vars.cfg
kikkiakki: oops, it snuck in there

lib/tenstorrent/qsfp_dd/qsfp_dd.c
```diff
@@ -0,0 +1,62 @@
...
+		ret = gpio_xp_set_config(gpio_xp_addrs[i], 0xf1);
+		if (ret != 0) {
+			return ret;
+		}
+		/* Set RST (P1) output to high */
+		ret = gpio_xp_set_output(gpio_xp_addrs[i], 0x2);
+		if (ret != 0) {
+			return ret;
+		}
+	}
```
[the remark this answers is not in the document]
cfriedt: And wow - I'm just looking at the schematic, and it looks like the qsfp is connected to the ARC, but the gpio to enable talking to the qsfp from the arc is attached to the stm32... like why?? Lol ... ¯\_(ツ)_/¯

include/tenstorrent/qsfp_dd.h
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * @brief Set up GPIOs on GPIO expander to enable high power mode on QSFP module
+ *
+ * @retval 0 on success.
+ * @retval -EIO	if an I/O error occurs.
+ */
+int enable_active_qsfp_dd(void);
```
cfriedt, line 18: A corresponding `int disable_active_qsfp_dd();` would be good too.

cfriedt (commented):
Some comments. And a lot of questions but more from the PCB side. Even the circuit to enable active QSFP does not really make a lot of sense.

Maybe we should shelve this one for a while. If it isn't needed, then maybe just close the PR.

PR #25, opened 2025-03-15
drivers: uart_tt_virt: remove to add cleanly later

cfriedt:
Remove the uart_tt_virt driver. This needs to be re-added again cleanly.

Typically, the way that this should be done is:

1st commit dt bindings
2nd commit driver
3rd commit test that exercises driver in CI

Occasionally, the 1st and 2nd commits can be combined.


cfriedt:
re-added in #26

PR #26, opened 2025-03-15
drivers: serial: vuart: add virtual uart driver and tests

cfriedt:
* dts: bindings: serial: add binding for `tenstorrent,vuart` driver
* drivers: serial: add tenstorrent virtual uart driver
* tests: drivers: uart: elementary: support for `tenstorrent,vuart`
* ci: enable continuous test for `tenstorrent,vuart`
* scripts: tt-console: add a simple c app for vuart console
* ci: build-native: verify that `tt-console` builds
* app: smc: populate vuart addresses if device is configured
* app: smc: add e2e test case for virtual uart

Aside from tests run in CI, it should be possible to verify functionality with
```shell
# Build console app
gcc -O0 -g -Wall -Wextra -Werror -std=gnu11 -I include -o scripts/tt-console/tt-console scripts/tt-console/console.c
# Build firmware
west build --sysbuild -p -b tt_blackhole@p100/tt_blackhole/smc app/smc -- -DEXTRA_DTC_OVERLAY_FILE="vuart.overlay" -DEXTRA_CONF_FILE="vuart.conf"
# Flash device
./scripts/bmc-reset.py && west flash && ./scripts/rescan-pcie.sh
 # Run console app
./scripts/tt-console/tt-console
```

```shell
./scripts/tt-console/tt-console
Press Ctrl-a,x to quit
         .:.                 .:
      .:-----:..             :+++-.
   .:------------:.          :++++++=:
 :------------------:..      :+++++++++
 :----------------------:.   :+++++++++
 :-------------------------:.:+++++++++
 :--------:  .:-----------:. :+++++++++
 :--------:     .:----:.     :+++++++++
 .:-------:         .        :++++++++-
    .:----:                  :++++=:.
        .::                  :+=:
          .:.               ::
          .===-:        .-===-
          .=======:. :-======-
          .==================-
          .==================-
           ==================:
            :-==========-:.
                .:====-.

*** Booting tt_blackhole with Zephyr OS v4.1.0 ***
Tenstorrent Blackhole CMFW 0.9.1
```

It's also fairly easy to duplicate this with `pytest` now too:
```shell
# Build firmware
west build --sysbuild -p -b tt_blackhole@p100/tt_blackhole/smc app/smc -- -DEXTRA_DTC_OVERLAY_FILE="vuart.overlay" -DEXTRA_CONF_FILE="vuart.conf"
# Flash device
./scripts/bmc-reset.py && west flash && ./scripts/rescan-pcie.sh
# Run Pytest
pytest -s ./app/smc/pytest/vuart.py 
```

```shell
pytest -s ./app/smc/pytest/vuart.py 
================================================================================ test session starts =================================================================================
platform linux -- Python 3.10.16, pytest-8.3.4, pluggy-1.5.0
rootdir: /home/ubuntu/cfriedt/tt-zephyr-platforms-work/tt-zephyr-platforms.git
collected 2 items                                                                                                                                                                    

app/smc/pytest/vuart.py .
console output successfully captured:
         .:.                 .:
      .:-----:..             :+++-.
   .:------------:.          :++++++=:
 :------------------:..      :+++++++++
 :----------------------:.   :+++++++++
 :-------------------------:.:+++++++++
 :--------:  .:-----------:. :+++++++++
 :--------:     .:----:.     :+++++++++
 .:-------:         .        :++++++++-
    .:----:                  :++++=:.
        .::                  :+=:
          .:.               ::
          .===-:        .-===-
          .=======:. :-======-
          .==================-
          .==================-
           ==================:
            :-==========-:.
                .:====-.

*** Booting tt_blackhole with Zephyr OS v4.1.0 ***
Tenstorrent Blackhole CMFW 0.9.1

.

================================================================================= 2 passed in 0.56s ==================================================================================
```

Of course, the test can also be run via twister (although there is less illustrative output).
```shell
twister -i -p tt_blackhole/tt_blackhole/smc --device-testing --tag e2e --hardware-map /opt/tenstorrent/twister/hw-map.yml --west-flash -T app/smc
```

Next steps (separate PR):

1. copy `tt-console` to `/opt/tenstorrent/bin` on CI machines
2. modify tt_blackhole smc boards to use the virtual uart instead of rtt or ram console
3. modify `app/smc/prj.conf` to use the virtual uart instead of rtt or ram console
4. update `hw-map.yml` to use `tt-console` as the `serial_pty` instead of `rtt`
5. ensure that all tests capture serial output via pcie
6. Update README to use tt-console instead of RTT for SMC

At that point, it would be good to demonstrate the Zephyr shell working over PCIe virtual uart. It would be good as well, to consider switching to VirtIO instead of a custom virtual uart implementation.


cfriedt:
Getting `vuart.overlay: No such file or directory` when running tests in CI.

Would be kind of great if twister searched the app directory for overlay files (normally it does this, I have no idea why id doesn't in our case).

cfriedt:
`../tt-zephyr-platforms/app/smc/vuart.overlay: No such file or directory`

Assumed that because we have `working-directory: zephyr` in `run-e2e-tests` that the `vuart.overlay` file would be relative to that directory, but of course not.

This is kind of ridiculous. Time to play the "what's my PWD" game!

cfriedt:
* only check for a timeout when `timeout_abs_ms` is non-zero

include/tenstorrent/uart_tt_virt.h
```diff
@@ -0,0 +1,259 @@
...
+	return tt_vuart_buf_size(head, tail) == 0;
+}
+
+/**
+ * @brief Determine if the given buffer is full.
+ *
+ * @param head Head index of the buffer
+ * @param tail Tail index of the buffer
+ * @param cap Capacity of the buffer
+ */
```
joelsmithTT, line 128: @return `true` if ... ?

scripts/tt-console/console.c
```diff
@@ -0,0 +1,918 @@
...
+
+	return tt_vuart_buf_size(vuart->rx_head, vuart->rx_tail);
+}
+
+static inline void vuart_putc(struct console *cons, int ch)
+{
+	volatile struct tt_vuart *const vuart = cons->vuart;
+	volatile char *const rx_buf = (volatile char *)&vuart->buf[vuart->tx_cap];
+
+	if (tt_vuart_buf_space(vuart->rx_cap, vuart->rx_head, vuart->rx_tail) > 0) {
```
joelsmithTT, line 623: Seems mismatched with `tt_vuart_buf_space(uint32_t head, uint32_t tail, uint32_t cap)` - head, tail, cap vs cap, head, tail.

scripts/tt-console/console.c
```diff
@@ -0,0 +1,918 @@
...
+	if (vuart->magic != cons->magic) {
+		return 0;
+	}
+
+	return tt_vuart_buf_size(vuart->rx_head, vuart->rx_tail);
+}
+
+static inline void vuart_putc(struct console *cons, int ch)
+{
+	volatile struct tt_vuart *const vuart = cons->vuart;
```
joelsmithTT, line 620: Want a magic check here like `vuart_getc` and `vuart_space` do?

joelsmithTT (commented):
Nice work, looks very thorough.  I'm not familiar with Zephyr enough to remark on anything Zephyr-specific, but left a few comments for your consideration.

scripts/tt-console/console.c
cfriedt: Good catch. I originally had it as cap, head, tail, and then decided to reorder to keep arguments in the same registers.

cfriedt: Fixed!

include/tenstorrent/uart_tt_virt.h
```diff
@@ -0,0 +1,259 @@
...
+static inline size_t tt_vuart_inst(volatile const struct tt_vuart *vuart)
+{
+	return vuart->version >> 24;
+}
+
+/**
+ * @brief Determine the size of the given buffer.
+ *
+ * This function initializes the virtual UART buffer descriptor with the provided buffer sizes.
```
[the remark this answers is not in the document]
cfriedt: Fixed!

dts/bindings/serial/tenstorrent,vuart.yml
```diff
@@ -0,0 +1,99 @@
...
+  - The remote processor updates the head index of the transmit buffer and the tail index of the
+    receive buffer.
+  - The local processor updates the head index of the receive buffer and the tail index of the
+    transmit buffer.
+  - Accesses are atomic with respect to other threads on the same processor. The local and
+    remote processors access independent indices, and indices are monotonically increasing,
+    generally mitigating the need for locks.
+  - Caching is disabled for the virtual uart memory region.
+  - The remote processor polls the transmit tail index to determine if there is data to read.
+  - The local processor polls the receive tail index to determine if there is data to read.
```
cfriedt, line 45: It might suffice to remove these lines

drivers/serial/uart_tt_virt.c
```diff
@@ -0,0 +1,418 @@
...
+static int uart_tt_virt_irq_rx_ready(const struct device *dev);
+static int uart_tt_virt_irq_tx_ready(const struct device *dev);
+#endif
+
+#ifdef CONFIG_UART_USE_RUNTIME_CONFIGURE
+static int uart_tt_virt_config_get(const struct device *dev, struct uart_config *cfg)
+{
+	struct uart_tt_virt_data *data = dev->data;
+
+	memcpy(cfg, &data->cfg, sizeof(*cfg));
```
alewycky-tenstorrent, line 60: Please use = to copy structures.

drivers/serial/uart_tt_virt.c
```diff
@@ -0,0 +1,418 @@
...
+	      (cfg->data_bits <= UART_CFG_DATA_BITS_8))) {
+		return -EINVAL;
+	}
+
+	if (!((cfg->flow_ctrl >= UART_CFG_FLOW_CTRL_NONE) &&
+	      (cfg->flow_ctrl <= UART_CFG_FLOW_CTRL_RTS_CTS))) {
+		return -EINVAL;
+	}
+
+	memcpy(&data->cfg, cfg, sizeof(*cfg));
```
alewycky-tenstorrent, line 91: Please use = to copy structures.

app/smc/src/main.c
```diff
@@ -55,7 +56,7 @@ int main(void)
 	}
 
 	while (1) {
-		k_yield();
+		k_msleep(1000);
```
alewycky-tenstorrent, line 59: This change must be 1. moved to a separate commit and 2. explained.

app/smc/src/main.c
```diff
@@ -75,3 +76,17 @@ static int _InitFW(void)
...
+void uart_tt_virt_init_callback(const struct device *dev, size_t inst)
+{
+	/*
+	 * Currently, there is only 1 scratch ram word set aside for virtual uarts.
+	 * We may want to expand on that at some point. Like 1 or 2 for various cores
+	 * with specific offsets for each core.
+	 */
+	sys_write32((uint32_t)(uintptr_t)uart_tt_virt_get(dev), STATUS_FW_VUART_REG_ADDR);
+}
+#endif
```
alewycky-tenstorrent, line 92: Please put the FW and test changes in separate commits. This, the main change and status_reg.h

app/smc/pytest/vuart.py
```diff
@@ -0,0 +1,47 @@
...
+
+def test_compile_tt_console(tt_console):
+    assert os.access(tt_console, os.X_OK)
+
+
+def test_boot_banner(tt_console, arc_chip):
+    arc_chip.axi_read32(ARC_STATUS)
+
+    # run 'tt-console' in quiet mode, and timeout after 500ms
+    cmd = f"{str(tt_console)} -q -w 500"
```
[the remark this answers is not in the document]
cfriedt: For some reason, I thought that `Path` objects might produce something different, but you're right. Thanks!

include/tenstorrent/uart_tt_virt.h
```diff
@@ -0,0 +1,259 @@
...
+			cap = vuart->rx_cap;
+			offs = vuart->tx_cap;
+		} else if (role == TT_VUART_ROLE_HOST) {
+			headp = (volatile atomic_uint *)&vuart->tx_head;
+			head = vuart->tx_head;
+			tail = vuart->tx_tail;
+			cap = vuart->tx_cap;
+			offs = 0;
+		} else {
+			/* assert? */
```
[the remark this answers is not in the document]
cfriedt: Good call. I'll add the assert too.

include/tenstorrent/uart_tt_virt.h
```diff
@@ -0,0 +1,259 @@
...
+ * @param dev Pointer to the device
+ * @return Pointer to the virtual UART buffer descriptor
+ */
+static inline volatile struct tt_vuart *uart_tt_virt_get(const struct device *dev)
+{
+	struct _tt_vuart_config {
+		void *ptr;
+	} *const config = (struct _tt_vuart_config *)dev->config;
+
+	return (volatile struct tt_vuart *)config->ptr;
```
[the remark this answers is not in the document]
cfriedt: This technically doesn't _need_ to be inline, so maybe I'll just convert it to a regular function.

scripts/tt-console/console.c
cfriedt: Missed this one yesterday, let me get to it.

lib/tenstorrent/bh_arc/status_reg.h
```diff
@@ -32,6 +32,7 @@
 #define STATUS_MSG_Q_ERR_FLAGS_REG_ADDR      RESET_UNIT_SCRATCH_RAM_REG_ADDR(9)
 #define STATUS_GDDR_AXI_EN_FLAGS_REG_ADDR    RESET_UNIT_SCRATCH_RAM_REG_ADDR(10)
 #define STATUS_MSG_Q_INFO_REG_ADDR           RESET_UNIT_SCRATCH_RAM_REG_ADDR(11)
+#define STATUS_FW_VUART_REG_ADDR             RESET_UNIT_SCRATCH_RAM_REG_ADDR(42)
```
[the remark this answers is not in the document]
cfriedt: Fixed! 
I reserved 40 and 41 for virtual uarts on arc for blackhole

dts/bindings/serial/tenstorrent,vuart.yml
```diff
@@ -0,0 +1,99 @@
...
+  magic:
+    type: int
+    default: 0x775e21a1
+    description: |
+      Magic number with which to identify the virtual UART device in-memory and determine when the
+      virtual uart device is ready for use.
+
+      The default value is a hexadecimal representation of the string "TTSeRial".
+
+  rx-cap:
```
danieldegrasse, line 68: Should these be stored as Kconfigs? If we need this configuration to be possible to do per instance, I think it makes sense to support this as a DT property. Otherwise I think it should be a Kconfig, like it is for RTT with SEGGER

dts/bindings/serial/tenstorrent,vuart.yml
```diff
@@ -0,0 +1,99 @@
...
+      at one time for reception via the virtual UART device.
+
+  tx-cap:
+    type: int
+    default: 3036
+    description: |
+      Transmit buffer capacity in bytes. This is the maximum amount of data that can be queued
+      at one time for transmission via the virtual UART device.
+
+  version:
```
danieldegrasse, line 82: Would there be value in some kind of `channel` ID as well? So we can support multiple vuarts

include/tenstorrent/uart_tt_virt.h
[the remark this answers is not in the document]
cfriedt: Fixed!

scripts/tt-console/console.c
cfriedt: Fixed!

include/tenstorrent/uart_tt_virt.h
```diff
@@ -0,0 +1,259 @@
...
+			cap = vuart->tx_cap;
+			offs = 0;
+		} else if (role == TT_VUART_ROLE_HOST) {
+			tailp = (volatile atomic_uint *)&vuart->rx_tail;
+			tail = vuart->rx_tail;
+			head = vuart->rx_head;
+			cap = vuart->rx_cap;
+			offs = vuart->tx_cap;
+		} else {
+			/* assert? */
```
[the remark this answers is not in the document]
cfriedt: Fixed!

include/tenstorrent/uart_tt_virt.h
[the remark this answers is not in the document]
cfriedt: Fixed!

drivers/serial/uart_tt_virt.c
cfriedt: Fixed!

drivers/serial/uart_tt_virt.c
cfriedt: Fixed!

drivers/serial/uart_tt_virt.c
```diff
@@ -0,0 +1,418 @@
...
+	struct uart_tt_virt_data *const data = dev->data;
+	volatile struct tt_vuart *vuart = config->vuart;
+
+	K_SPINLOCK(&data->rx_lock) {
+		data->rx_irq_en = true;
+		submit = !tt_vuart_buf_empty(vuart->rx_head, vuart->rx_tail);
+	}
+
+	if (submit) {
+		(void)k_work_submit(&data->irq_work);
```
danieldegrasse, line 236: Won't this only run the IRQ worker once, and then stop? Or is there somewhere where we reschedule the IRQ worker to poll periodically?

scripts/tt-console/console.c
```diff
@@ -0,0 +1,918 @@
...
+}
+
+static void console_init(struct console *cons)
+{
+	*cons = (struct console){
+		.dev_name = TT_DEVICE,
+		.fd = -1,
+		.addr = UART_TT_VIRT_DISCOVERY_ADDR,
+		.magic = UART_TT_VIRT_MAGIC,
+		.pci_device_id = BH_SCRAPPY_PCI_DEVICE_ID,
```
danieldegrasse, line 234: I think this needs to be a runtime arg to work on all boards and not just scrappy

app/smc/pytest/vuart.py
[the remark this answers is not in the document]
cfriedt: Fixed!

.github/workflows/build-native.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  build-tt-console:
+    runs-on: ubuntu-22.04
+    name: Build tt-console
+    steps:
+    - name: Checkout the code
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Run the script
```
danieldegrasse, line 14: Can we upload an artifact here so that those who are lazy (me) don't have to build the console?

app/smc/src/main.c
cfriedt: I'll leave it out for now (assuming that it's fine). However, `k_yield()` by itself is equivalent to busy-looping if there are no other active threads. Normally, the main thread should run to completion when there is no work left to do, but in our case, the ARC dies for some reason.

That's something we need to fix.

app/smc/src/main.c
cfriedt: Fixed!

app/smc/src/main.c
danieldegrasse: Agree here- is there something in the way the UART driver polls that warrants this?

danieldegrasse (commented):
I like the general approach here, but why not use SEGGER's RTT for the R/W protocol? the reason I suggest this is that it saves us from maintaining a downstream driver, and as I understand it RTT is just a circular buffer in memory, so we should be able to use it here as well.

The other advantage to this (IMO) is that we get extra features RTT offers, like the ability to support multiple channels. We can still pass the RTT buffer address to the C application on the host using the scratch register

dts/bindings/serial/tenstorrent,vuart.yml
cfriedt: Per device, uarts can have different sized buffers, so I think it's fine in DT.

dts/bindings/serial/tenstorrent,vuart.yml
cfriedt: That's captured by the `INST` identifier. Since each channel still needs metadata, it's simple enough to just use separate instances.

drivers/serial/uart_tt_virt.c
cfriedt: It's only intended to run the IRQ worker once.

scripts/tt-console/console.c
cfriedt: It is a runtime arg.

```cpp
"-i <pci_device_id> : pci device id (default: %04x)\n"
```

.github/workflows/build-native.yml
cfriedt: I think we'll just deploy this on all CI machines as a static binary. It may change at any time (e.g. after TLB kernel changes are incorporated), so it's probably best not to rely on a precompiled binary for now.

cfriedt:
> I like the general approach here, but why not use SEGGER's RTT for the R/W protocol? the reason I suggest this is that it saves us from maintaining a downstream driver, and as I understand it RTT is just a circular buffer in memory, so we should be able to use it here as well.

We could, but I think the eventual goal is to use VirtIO serial devices rather than Segger's protocol. This is really more of a stopgap solution until we have support for VirtIO.

> The other advantage to this (IMO) is that we get extra features RTT offers, like the ability to support multiple channels. We can still pass the RTT buffer address to the C application on the host using the scratch register

This already does support multiple channels (just as separate vuart instances).

app/smc/src/main.c
cfriedt: Actually, it seems that this is necessary, for the reasons I mentioned above.

cfriedt: Yes - so adding the `k_msleep(1000)` fixed the previous failed test run.

@alewycky-tenstorrent @danieldegrasse - I would be happy to submit that commit as a separate bugfix PR if that's preferred.

danieldegrasse (approved):

cfriedt:
Just adding an attachment here showing about 30 kiB of log messages over the course of almost 6 minutes.

[blah.txt](https://github.com/user-attachments/files/19492384/blah.txt)

PR #28, opened 2025-03-17
ci: bump metal revision to v0.56.0

danieldegrasse:
Bump metal revision to v0.56.0- this is needed to support P100a cards
properly.


cfriedt:
Probably can close this in favour of #35

danieldegrasse:
Yeah, we need to. This won't work because the P100a in CI has some tensix columns harvested.

PR #29, opened 2025-03-18
zephyr: blobs: Update GDDR FW to v2.6.

abuckTT:
- Expands GDDR telemetry table to include:
- - Speed
- - Training status
- - Error status
- - FW version
- - DRAM vendor
- - Number of corrected EDC errors
- - Number of uncorrected EDC errors
- Fix bug for incorrect timing parameter for Micron E die revision


cfriedt (approved):

cfriedt:
@abuckTT - in a follow up pr would be good to include some tests from the arc. Bist but also even some simple ztest that might verify some states in the mrisc fw

abuckTT:
@cfriedt agreed I need to add a check that training was at least successful. Also need to add a check that telemetry table versions are compatible between mrisc fw and smc. Let me create a ticket for that.

PR #30, opened 2025-03-19
Increase VR switching freq

nxuTT:
modify serdes vr rails switching frequency to maximum on p150, p300, and Galaxy
We skip serdes_vdd on p300 left, because left and right share serdes_vdd, only right has i2c connection to serdes_vdd.

### DIFF
--- a/lib/tenstorrent/bh_arc/regulator.c
+++ b/lib/tenstorrent/bh_arc/regulator.c
@@ -24,6 +24,8 @@
 #define PMBUS_MST_ID 1
 
 /* PMBus Spec constants */
+#define MFR_CTRL_OPS                   0xD2
+#define MFR_CTRL_OPS_DATA_BYTE_SIZE    1
 #define VOUT_COMMAND                   0x21
 #define VOUT_COMMAND_DATA_BYTE_SIZE    2
 #define VOUT_SCALE_LOOP                0x29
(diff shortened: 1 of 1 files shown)


lib/tenstorrent/bh_arc/regulator.c
```diff
@@ -266,6 +271,28 @@ void RegulatorInit(PcbType board_type)
...
+			}
+
+			I2CInit(I2CMst, *addr_ptr, I2CFastMode, PMBUS_MST_ID);
+			I2CWriteBytes(PMBUS_MST_ID, MFR_CTRL_OPS, PMBUS_CMD_BYTE_SIZE,
+				      &mfr_ctrl_ops, MFR_CTRL_OPS_DATA_BYTE_SIZE);
+		}
+	}
+
+	/* 100us to flush the tx of i2c */
+	WaitUs(100);
```
afongTT, line 295: Doesn't the `I2CWriteBytes` routine already wait for all TX done before returning?

bingliTT:
Tested on P300 in-house, confirmed VR change is working

lib/tenstorrent/bh_arc/regulator.c
nxuTT: Yes, it should be flushed out from TX FIFO, but just not sure if the changes got applied immediately. I removed the delay anyways.

afongTT (approved):

PR #31, opened 2025-03-19
ci: workflows: use docker container to run metal workflow

danieldegrasse:
Use a docker container to run the metal tests. This should enable us to run metal tests on a per-PR basis, rather than nightly


danieldegrasse:
Closing in favor of #35

PR #32, opened 2025-03-21
Checks GDDR training success

abuckTT:
Will mark hw_init_status to kHwInitError if GDDR training failed. Adds logging to indicate what problems may have occurred.


lib/tenstorrent/bh_arc/init.c
```diff
@@ -128,39 +157,62 @@ static void InitMrisc(void)
 	}
 	for (uint8_t gddr_inst = 0; gddr_inst < 8; gddr_inst++) {
 		if ((dram_mask >> gddr_inst) & 1) {
```
alewycky-tenstorrent, line 159: While you're here, please change to IS_BIT_SET.

lib/tenstorrent/bh_arc/init.c
```diff
@@ -128,39 +157,62 @@ static void InitMrisc(void)
...
+			}
+			MriscRegWrite32(gddr_inst, MRISC_INIT_STATUS, MRISC_INIT_BEFORE);
 			ReleaseMriscReset(gddr_inst);
 		}
 	}
+	bool all_passed = true;
 
-	/* TODO: Check for MRISC FW success / failure */
+	for (uint8_t gddr_inst = 0; gddr_inst < 8; gddr_inst++) {
+		if ((dram_mask >> gddr_inst) & 1) {
```
alewycky-tenstorrent, line 205: Please use IS_BIT_SET.

afongTT:
I was wondering if it makes sense to do these potentially long checks later in main, or possibly scheduled?

One benefit is that you kick off all GDDR and ethernet training before you start polling for them to complete.

abuckTT:
> I was wondering if it makes sense to do these potentially long checks later in main, or possibly scheduled?
> 
> One benefit is that you kick off all GDDR and ethernet training before you start polling for them to complete.

I'd still like it to be part of InitHW() so that we set boot_status0.f.hw_init_status to error if issues. However, I could put the check at the end of the rest of the init sequence. Then we can hide some of the latency.

abuckTT:
Also measured init time. In passing cases, it currently takes ~5ms.

.gitignore
```diff
@@ -107,3 +107,5 @@ SphinxLint.txt
 TextEncoding.txt
 YAMLLint.txt
 *.fwbundle
+
+scripts/tt-console/tt-console
```
danieldegrasse, line 111: Please split this change into a distinct commit- ideally we'd have it in another PR, but I'm ok with it here since it's a small change

abuckTT:
TODO in subsequent PR: 
Need to replace all instances of using `tile_enable.gddr_enabled` with `GetDramMask()`. This is needed because fw_table has a DRAM mask that can disable an instance.

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -34,6 +35,24 @@ uint32_t MriscL1Read32(uint8_t gddr_inst, uint32_t addr)
 	return NOC2AXIRead32(0, MRISC_SETUP_TLB, MRISC_L1_ADDR + addr);
 }
 
+uint32_t MriscRegRead32(uint8_t gddr_inst, uint32_t addr)
+{
+	uint8_t x, y;
+
+	GetGddrNocCoords(gddr_inst, MRISC_FW_NOC2AXI_PORT, 0, &x, &y);
+	NOC2AXITlbSetup(0, MRISC_SETUP_TLB, x, y, MRISC_REG_ADDR + 0xff000000);
```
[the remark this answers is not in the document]
abuckTT: Fixed

lib/tenstorrent/bh_arc/init.c
```diff
@@ -432,7 +488,11 @@ static int InitHW(void)
 	} else {
 		boot_status0.f.fw_id = FW_ID_SMC_NORMAL;
 	}
-	boot_status0.f.hw_init_status = kHwInitDone;
+	if (init_errors) {
+		boot_status0.f.hw_init_status = kHwInitError;
+	} else {
+		boot_status0.f.hw_init_status = kHwInitDone;
+	}
```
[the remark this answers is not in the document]
abuckTT: Fixed

lib/tenstorrent/bh_arc/init.c
```diff
@@ -103,7 +106,34 @@ static void DeassertRiscvResets(void)
...
+		poll_val = MriscRegRead32(gddr_inst, MRISC_INIT_STATUS);
+		if (poll_val == MRISC_INIT_FINISHED || poll_val == MRISC_INIT_FAILED) {
+			break;
+		}
+		if (k_uptime_delta(&start_time) > MRISC_INIT_TIMEOUT) {
+			LOG_ERR("Timeout after %lld ms waiting for GDDR instance %d to "
+				"initialize.\n",
+				k_uptime_delta(&start_time), gddr_inst);
+			return -ETIME;
+		}
```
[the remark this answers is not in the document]
abuckTT: Reduced timeout to 1s for all controllers. Measured it took about 5ms to run, but I just added big buffer.

lib/tenstorrent/bh_arc/init.c
```diff
@@ -103,7 +106,34 @@ static void DeassertRiscvResets(void)
...
+		k_msleep(1);
+	}
+	if (poll_val == MRISC_INIT_FINISHED) {
+		return EXIT_SUCCESS;
+	}
+	LOG_ERR("GDDR instance %d failed to initialize. "
+		"Something went wrong.\n",
+		gddr_inst);
```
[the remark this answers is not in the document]
abuckTT: Added. Will just print the post code after timeout or if failure.

lib/tenstorrent/bh_arc/init.c
```diff
@@ -128,39 +157,62 @@ static void InitMrisc(void)
...
 	if (SetGddrMemClk(gddr_speed / GDDR_SPEED_TO_MEMCLK_RATIO)) {
-		/* Error */
-		/* TODO: Handle this more gracefully */
-		return;
+		LOG_ERR("Failed to set GDDR memory clock to requested: %d Mbps\n", gddr_speed);
+		return -EIO;
 	}
 
 	for (uint8_t gddr_inst = 0; gddr_inst < 8; gddr_inst++) {
 		if ((dram_mask >> gddr_inst) & 1) {
```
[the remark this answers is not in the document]
abuckTT: Fixed

lib/tenstorrent/bh_arc/gddr.c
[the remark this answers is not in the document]
danieldegrasse: Agree on `0xff000000`- at a minimum we need to define some kind of constant for this rather than using a magic number

lib/tenstorrent/bh_arc/init.c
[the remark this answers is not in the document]
danieldegrasse: Can we add a post code to MRISC if there isn't one? At a minimum I think we should distinguish between "init failed" and "init timeout"

lib/tenstorrent/bh_arc/init.c
```diff
@@ -104,64 +107,103 @@ static void DeassertRiscvResets(void)
...
+	if (poll_val == MRISC_INIT_FINISHED) {
+		return EXIT_SUCCESS;
+	}
+	uint32_t post_code = MriscRegRead32(gddr_inst, MRISC_POST_CODE);
+
+	if (timedout) {
+		LOG_ERR("Timeout after %d ms waiting for GDDR instance %d to "
+			"initialize. Post code: 0x%x\n",
+			MRISC_INIT_TIMEOUT, gddr_inst, post_code);
+		return -ETIME;
```
danieldegrasse, line 135: Nit but I would prefer `-ETIMEDOUT` here

lib/tenstorrent/bh_arc/init.c
```diff
@@ -104,64 +107,103 @@ static void DeassertRiscvResets(void)
...
-	}
-	for (uint8_t gddr_inst = 0; gddr_inst < 8; gddr_inst++) {
-		if ((dram_mask >> gddr_inst) & 1) {
-			LoadMriscFw(gddr_inst, large_sram_buffer, fw_size);
+		LOG_ERR("Failed to load MRISC FW from file system to ARC.\n");
+		return -EIO;
+	}
+	uint32_t dram_mask = GetDramMask();
+
+	for (uint8_t gddr_inst = 0; gddr_inst < NUM_GDDR; gddr_inst++) {
```
danieldegrasse, line 161: Not something we have to do now, but should we make `NUM_GDDR` a Kconfig? Generally wondering if this would be useful for the future, in case we reuse GDDR code on other platforms

lib/tenstorrent/bh_arc/init.c
```diff
@@ -427,6 +473,21 @@ static int InitHW(void)
...
+	SetPostCode(POST_CODE_SRC_CMFW, POST_CODE_ARC_INIT_STEPE);
+	/* Check GDDR training status. */
+	if (!IS_ENABLED(CONFIG_TT_SMC_RECOVERY)) {
+		k_timepoint_t timeout = sys_timepoint_calc(K_MSEC(MRISC_INIT_TIMEOUT));
+
+		for (uint8_t gddr_inst = 0; gddr_inst < NUM_GDDR; gddr_inst++) {
+			if (IS_BIT_SET(GetDramMask(), gddr_inst)) {
+				if (CheckGddrTraining(gddr_inst, timeout)) {
+					LOG_ERR("GDDR instance %d failed training.\n", gddr_inst);
+					init_errors = true;
```
danieldegrasse, line 485: Should we break out here if the return value was a timeout? I guess it doesn't matter too much since the additional calls will complete pretty quickly.

lib/tenstorrent/bh_arc/init.c
```diff
@@ -427,6 +473,21 @@ static int InitHW(void)
...
 
+	SetPostCode(POST_CODE_SRC_CMFW, POST_CODE_ARC_INIT_STEPE);
+	/* Check GDDR training status. */
+	if (!IS_ENABLED(CONFIG_TT_SMC_RECOVERY)) {
+		k_timepoint_t timeout = sys_timepoint_calc(K_MSEC(MRISC_INIT_TIMEOUT));
+
+		for (uint8_t gddr_inst = 0; gddr_inst < NUM_GDDR; gddr_inst++) {
+			if (IS_BIT_SET(GetDramMask(), gddr_inst)) {
+				if (CheckGddrTraining(gddr_inst, timeout)) {
+					LOG_ERR("GDDR instance %d failed training.\n", gddr_inst);
```
danieldegrasse, line 484: Let's include the return value of `CheckGddrTraining` in the log here so it is obvious if the failure was a timeout or not

lib/tenstorrent/bh_arc/init.c
alewycky-tenstorrent: In the case of timeout every remaining still gets one chance to report that it's ready.

alewycky-tenstorrent:
> TODO in subsequent PR: Need to replace all instances of using `tile_enable.gddr_enabled` with `GetDramMask()`. This is needed because fw_table has a DRAM mask that can disable an instance.

I don't think we should do this here, but rather integrate dram_mask into harvesting so that it shows up in tile_enable.gddr_enabled. @afongTT

lib/tenstorrent/bh_arc/init.c
alewycky-tenstorrent: It's possible but very unlikely that we produce a BH-like chip with different physical layout.

Changing physical GDDR count would change the NOC coords of the GDDR controllers, so we'd need to update the mapping function as well.

.gitignore
abuckTT: Just split it into its own PR https://github.com/tenstorrent/tt-zephyr-platforms/pull/38

afongTT:
> > TODO in subsequent PR: Need to replace all instances of using `tile_enable.gddr_enabled` with `GetDramMask()`. This is needed because fw_table has a DRAM mask that can disable an instance.
> 
> I don't think we should do this here, but rather integrate dram_mask into harvesting so that it shows up in tile_enable.gddr_enabled. @afongTT

Yes, makes sense to me.

Unless there's any reason to have that setting mean "leave tile clock enabled, but don't load GDDR FW". But I doubt that would be a real use case.

lib/tenstorrent/bh_arc/init.c
```diff
@@ -36,8 +36,11 @@
 #include <tenstorrent/post_code.h>
 #include <tenstorrent/tt_boot_fs.h>
 #include <zephyr/init.h>
+#include <zephyr/logging/log.h>
 #include <zephyr/kernel.h>
 
+LOG_MODULE_REGISTER(InitHW, LOG_LEVEL_DBG);
```
alewycky-tenstorrent, line 42: Should we be setting LOG_LEVEL_DBG by default in production? If you don't pass a logging level to LOG_MODULE_REGISTER, you get CONFIG_LOG_DEFAULT_LEVEL (which defaults to INFO).

danieldegrasse: Good point here. We should probably define a Kconfig for the log level used by the `BH_ARC` libraries. See here for an example of how to source a Kconfig template that will define Kconfigs like `CONFIG_XX_LOG_LEVEL_DBG, CONFIG_XXX_LOG_LEVEL_ERR, ...`: https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/sdhc/Kconfig#L56

abuckTT: In main.c, the main log is also set to LOG_LEVEL_DBG rather than INFO (through the Kconfig stuff). Should I also change it to INFO?

danieldegrasse: Let's reuse `CONFIG_TT_APP_LOG_LEVEL` here- if you want to create a separate commit changing this: https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/app/smc/prj.conf#L11 to `CONFIG_TT_APP_LOG_LEVEL_INF` I think that would be good as well.

abuckTT: Fixed

PR #33, opened 2025-03-21
tenstorrent: bh_arc: init boot_freq even when aiclk_ppm is disabled

afongTT:
This was causing an issue where unforcing the frequency with aiclk_ppm disabled caused CMFW to attempt setting an AICLK frequency of 0, which hangs the chip.

### DIFF
--- a/lib/tenstorrent/bh_arc/aiclk_ppm.c
+++ b/lib/tenstorrent/bh_arc/aiclk_ppm.c
@@ -111,6 +111,12 @@ uint32_t GetMaxAiclkForVoltage(uint32_t voltage)
 	return low_freq - 1;
 }
 
+void InitArbMaxVoltage(void)
+{
+	/* ArbMaxVoltage is statically set to the frequency of the maximum voltage */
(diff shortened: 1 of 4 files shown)


danieldegrasse (approved):

danieldegrasse:
Not blocking for this PR, but could we add a testcase where we try forcing the AICLK to a variety of values (as well as unforcing it) and verify the chip still functions?

afongTT:
> Not blocking for this PR, but could we add a testcase where we try forcing the AICLK to a variety of values (as well as unforcing it) and verify the chip still functions?

I agree, that seems like a test we should have.

alewycky-tenstorrent:
Can we give the other aiclk_ppm fields valid values too? Actually why not just call InitAiclkPPM? Except maybe the max voltage limit arbiter, the rest all seem correct.

afongTT:
> Can we give the other aiclk_ppm fields valid values too? Actually why not just call InitAiclkPPM? Except maybe the max voltage limit arbiter, the rest all seem correct.

Updated to call InitAiclkPPM right after InitPLLs. Does this work better?

alewycky-tenstorrent (approved):

PR #34, opened 2025-03-22
ina228: fix input current formatting

palexsonTT:
Send input current data from BMC to SMC in signed int 16.16 format. Provide input current as a float to the board power throttler.

### DIFF
--- a/app/bmc/src/main.c
+++ b/app/bmc/src/main.c
@@ -124,8 +124,9 @@ void ina228_current_update(void)
 
 	sensor_sample_fetch_chan(ina228, SENSOR_CHAN_CURRENT);
 	sensor_channel_get(ina228, SENSOR_CHAN_CURRENT, &current_sensor_val);
-	uint32_t current = (current_sensor_val.val1 << 16) |
-			   ((current_sensor_val.val2 / 100) & 0xFFFF); /* Truncate */
+
+	int32_t current =
+		(current_sensor_val.val1 << 16) + (current_sensor_val.val2 * 65536ULL) / 1000000ULL;
 
 	ARRAY_FOR_EACH_PTR(BH_CHIPS, chip) {
 		bh_chip_set_input_current(chip, &current);
--- a/lib/tenstorrent/bh_arc/cm2bm_msg.c
+++ b/lib/tenstorrent/bh_arc/cm2bm_msg.c
@@ -28,7 +28,7 @@ typedef struct {
 
 static Cm2BmMsgState cm2bm_msg_state;
 static bool bmfw_ping_valid;
-static uint32_t current;
+static int32_t current;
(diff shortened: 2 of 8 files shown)


afongTT:
> Is negative current ever something that we would anticipate seeing? Even with a transient, it seems strange.
No, I think the use of signed 16.16 fixed point is strictly to match the format we have in SMC's telemetry table.

> If using floats can be avoided (even in the power throttler) we really should.
I prefer to use floats for our power throttling in SMC, since the fixed point math gets really messy. We also have an FPU on ARC, and we should generally have an FPU on all SMCs for this reason.

I agree we should try to avoid conversion to float within BMC FW though.

app/bmc/src/main.c
```diff
@@ -124,8 +124,9 @@ void ina228_current_update(void)
 
 	sensor_sample_fetch_chan(ina228, SENSOR_CHAN_CURRENT);
 	sensor_channel_get(ina228, SENSOR_CHAN_CURRENT, &current_sensor_val);
-	uint32_t current = (current_sensor_val.val1 << 16) |
-			   ((current_sensor_val.val2 / 100) & 0xFFFF); /* Truncate */
+
+	float current_flt = current_sensor_val.val1 + current_sensor_val.val2 / 1000000.0;
+	int32_t current = current_flt * 65536;
```
afongTT, line 129: I think you can avoid the floating point calculation by doing something like:

`uint32_t current = (current_sensor_val.val1 << 16) + current_sensor_val.val2 / (1000000/65536)`

lib/tenstorrent/bh_arc/cm2bm_msg.c
```diff
@@ -228,17 +228,22 @@ int32_t Bm2CmSendCurrentHandler(const uint8_t *data, uint8_t size)
...
 /* TODO: Put these somewhere else? */
-uint32_t GetInputCurrent(void)
+int32_t GetInputCurrent(void)
 {
 	return current;
 }
 
+float GetInputCurrentFloat(void)
+{
+	return current / 65536.0;
```
afongTT, line 244: Can you create a function in telemetry.h that handles this conversion?

e.g.
```c
float ConvertTelemetryToFloat(int32_t value)
{
	/* Convert signed int 16.16 format to float */
	if (value == INT32_MIN) {
		return FLT_MAX;
	} else {
		return value / 65536.0F;
	}
}
```

cfriedt:
Avoiding floating point in bare-metal code (e.g. drivers) is pretty common (both Linux and Zephyr avoid using floats).

It's fine to convert to float at the output (essentially the presentation layer), but I wouldn't advise using down at the driver layer.

afongTT (approved):

cfriedt (approved):
Looks better to me. Thanks!

This is one area where I wish Zephyr would be a bit better. Even a macro or two to convert from sensor_value to arbitrary Qm.n, or vice versa, would be pretty great.

PR #35, opened 2025-03-24
ci: workflows: use docker container to run metal workflow

danieldegrasse:
Use a docker container to run the metal tests. This should enable us to run metal tests on a per-PR basis, rather than nightly


danieldegrasse:
@afongTT or @TTDRosen, could either of you take a look at this and approve if you're ok with it? My plan is to merge this, than update the metal revision in use within #39, so that we test metal with harvesting on there.

.github/workflows/hardware.yml
```diff
@@ -395,3 +395,153 @@ jobs:
...
+        env:
+          ARCH_NAME: blackhole
+          TT_METAL_HOME: /tt-metal
+          PYTHONPATH: /tt-metal
+        run: |
+          # Source metal env
+          source /metal-env/bin/activate
+          # Run metal tests
+          # Skipped because the P100a has harvested columns
+          # tests/scripts/run_cpp_unit_tests.sh
```
afongTT, line 536: Is this something they plan to fix?

e.g. Maybe with the harvesting PR?

afongTT (approved):

.github/workflows/hardware.yml
```diff
@@ -395,3 +395,153 @@ jobs:
...
+        uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            ci:
+              - 'submanifests/ci.yaml'
+
+      - name: Update Metal
+        id: update_metal
+        if: steps.filter.outputs.ci == 'true'
```
afongTT, line 508: I guess this is the rule that triggers a rebuild of Metal.
How does that get back into the base container?

cfriedt (commented):
Would be great to make this far less verbose and re-use some of the prepare-zephyr workflow or the upstream workflow.

What is the main motivators for adding our own docker image and where is the dockerfile for that image?

I have some concerns about rebuilding metal on every PR. Can that step be cached or is it already cached? How much overhead does it add?

Is there a way that we can automate checking out the appropriate version of metal for building against firmware. E.g. adding a `zephyr/module.yml` file to metal and pulling it in with a specific revision in `west.yml`?

submanifests/ci.yaml
```diff
@@ -0,0 +1,26 @@
...
+  projects:
+    - name: openocd
+      revision: 720c2b6023433ae4d11d929a74c71e96f5d8159f
+      path: ci/openocd
+      remote: danieldegrasse
+      submodules: true
+      groups:
+        - ci
+    - name: tt-metal
+      revision: v0.56.0
```
cfriedt, line 21: +1

This is good. It will allow us to update more easily without breaking anything. Can we simplify cloning?

https://github.com/zephyrproject-rtos/example-application/blob/main/.github/workflows/build.yml#L27

Metal will need to be able to test for certain features before enabling them though, since they do not have a similar means of tracking changes.

.github/workflows/hardware.yml
danieldegrasse: Yes, it is- the P100a we have in CI can't run these tests until we have harvesting enabled.

In our harvesting PR we should turn this test back on, and make sure it passes

.github/workflows/hardware.yml
danieldegrasse: We'll rebuild the base container when a branch is merged to `main` with changes to the manifest- that way if we ever bump the openocd revision or metal revision the container gets rebuilt. See the PR for that feature here: https://github.com/tenstorrent/tt-zephyr-platforms/pull/40

danieldegrasse:
> Would be great to make this far less verbose and re-use some of the prepare-zephyr workflow or the upstream workflow.

> https://github.com/zephyrproject-rtos/example-application/blob/main/.github/workflows/build.yml#L27

As far as I can tell, the prepare-zephyr workflow does not rebase the branch within the PR onto `main`, does it? I think that is a critical step we should be doing to get accurate test results per PR.

I am happy to coalesce this workflow into a reusable file though, if you'd like

> What are the main motivators for adding our own docker image and where is the dockerfile for that image?

We can include a prebuilt copy of metal that we control the revision of. The image is right here: https://github.com/tenstorrent/tt-zephyr-platforms/pkgs/container/tt-zephyr-platforms%2Ftt-zephyr-platforms%2Fubuntu-22-04-ci-image

PR #40 includes an action for rebuilding the docker image when a PR is merged to main with changes to the manifest

> I have some concerns about rebuilding metal on every PR. Can that step be cached or is it already cached? How much overhead does it add per-pr in terms of time?

I do too- we will not rebuild metal on every PR. The docker container *is* the build cache- we have a prebuilt copy of metal in the container, and only rebuild it if the manifest file for `tt-metal` has changed. Rebuilding has significant overhead (15-20 minutes), so we will only do it when updating the docker container or metal revision within a PR.

Github caching would be an option here, but I think the amount of data we'd cache is too large. With `du`, the metal build folder is `3.8G`, and the metal python env is `2.8G`. Github actions limits the cache size per repo to 10G, so we could only cache one full metal build at a time

submanifests/ci.yaml
danieldegrasse: The method I want to use here is sequencing like this:
- metal makes a PR on their side to enable a feature
- We create a PR bumping the metal revision, and turning on whatever feature they want to test
- Our CI runs metal, and validates that tests pass with the new feature enabled

.github/workflows/hardware.yml
```diff
@@ -395,3 +395,153 @@ jobs:
...
+        shell: bash
+        run: |
+          # Clone the new metal repo
+          west config manifest.group-filter -- +ci
+          west update
+          rm -r /metal-env
+          rm -r /tt-metal
```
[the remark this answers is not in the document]
danieldegrasse: These files are already in the container, so when the metal revision is bumped we remove them then replace them.

submanifests/ci.yaml
```diff
@@ -0,0 +1,26 @@
...
+      url-base: https://github.com/danieldegrasse
+    - name: tenstorrent
+      url-base: https://github.com/tenstorrent
+  defaults:
+    remote: tenstorrent
+  group-filter:
+    - -ci
+  # zephyr-keep-sorted-start re(^\s+\- name:)
+  projects:
+    - name: openocd
```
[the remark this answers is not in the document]
danieldegrasse: This is a good point- I tried to use upstream openocd (with these patches merged), and it does not build currently. So as far as a path forwards, do you think the best option is to try to merge these changes into Zephyr's fork of openocd?

cfriedt:
@danieldegrasse - is there some kind of "install" stage for metal? Like "make install" to some directory? I don't think we would need all of the build files. Probably just some `.so` files. If metal uses a common build method for python packages, we can just add them to scripts/requirements.txt for certain versions. IIRC, the zephyr workflow for installing python packages also caches them.

It is a bit annoying that it would transfer data over the network for binary artifacts, so if we can find out the minimum of what is needed to install on a host system, that can be compressed to reduce bandwidth.

cfriedt:
@danieldegrasse - is this one still useful? Could use a rebase or maybe put it back into draft if it's not ready yet.

danieldegrasse:
> @danieldegrasse - is this one still useful? Could use a rebase or maybe put it back into draft if it's not ready yet.

Nah, I'd rather get #99 working

PR #37, opened 2025-03-24
boards: fw_table: update soft harvesting according to product specs

afongTT:
This makes sure that all boards of the same type look the same, despite different harvesting configurations.

Note that this setting doesn't actually have any effect until feature_enable.harvesting_en is enabled in the fw_tables.

### DIFF
--- a/boards/tenstorrent/tt_blackhole/spirom_data_tables/P100/fw_table.txt
+++ b/boards/tenstorrent/tt_blackhole/spirom_data_tables/P100/fw_table.txt
@@ -64,3 +64,9 @@ eth_property_table {
   eth_disable_mask: 0x3fff
   eth_disable_mask_en: true
 }
+
+product_spec_harvesting {
+  dram_disable_count: 0
+  eth_disabled: false
+  tensix_col_disable_count: 0
+}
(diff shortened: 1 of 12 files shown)


boards/tenstorrent/tt_blackhole/spirom_data_tables/P100/fw_table.txt
```diff
@@ -64,3 +64,9 @@ eth_property_table {
   eth_disable_mask: 0x3fff
   eth_disable_mask_en: true
 }
+
+product_spec_harvesting {
+  dram_disable_count: 0
+  eth_disabled: true
+  tensix_col_disable_count: 0
+}
```
alewycky-tenstorrent, line 72: Let me make sure I understand P100 being 0/false/0. It's no longer a product and all the boards we ever made are fully-functional chips? So we leave them as unharvested engineering boards?

afongTT: Yes, that was exactly my line of thinking.
And I re-confirmed with @tcolacoTT all p100 were built with fully-functional chips.

afongTT: Oh, I should change eth_disabled back to false.

alewycky-tenstorrent (approved):

PR #39, opened 2025-03-24
boards: fw_table: enable harvesting and NOC translation

afongTT:
Enable harvesting and NOC translation in the FW table.

### DIFF
--- a/boards/tenstorrent/tt_blackhole/spirom_data_tables/GALAXY/fw_table.txt
+++ b/boards/tenstorrent/tt_blackhole/spirom_data_tables/GALAXY/fw_table.txt
@@ -17,12 +17,13 @@ chip_limits {
 
 feature_enable {
   cg_en: true
-  noc_translation_en: false
+  noc_translation_en: true
   ddr_train_en: true
   aiclk_ppm_en: false
   watchdog_en: false
   smbus_en: false
   fan_ctrl_en: false
+  harvesting_en: true
 }
 
 fan_table {
--- a/boards/tenstorrent/tt_blackhole/spirom_data_tables/P100/fw_table.txt
+++ b/boards/tenstorrent/tt_blackhole/spirom_data_tables/P100/fw_table.txt
@@ -18,12 +18,13 @@ chip_limits {
 
 feature_enable {
   cg_en: true
-  noc_translation_en: false
+  noc_translation_en: true
   ddr_train_en: true
   aiclk_ppm_en: true
   watchdog_en: false
   smbus_en: false
   fan_ctrl_en: false
+  harvesting_en: true
 }
 
 fan_table {
--- a/boards/tenstorrent/tt_blackhole/spirom_data_tables/P100A/fw_table.txt
+++ b/boards/tenstorrent/tt_blackhole/spirom_data_tables/P100A/fw_table.txt
@@ -18,12 +18,13 @@ chip_limits {
 
 feature_enable {
   cg_en: true
-  noc_translation_en: false
+  noc_translation_en: true
   ddr_train_en: true
   aiclk_ppm_en: true
   watchdog_en: false
   smbus_en: false
   fan_ctrl_en: true
+  harvesting_en: true
 }
 
 fan_table {
--- a/boards/tenstorrent/tt_blackhole/spirom_data_tables/P150A/fw_table.txt
+++ b/boards/tenstorrent/tt_blackhole/spirom_data_tables/P150A/fw_table.txt
@@ -18,12 +18,13 @@ chip_limits {
 
 feature_enable {
   cg_en: true
-  noc_translation_en: false
+  noc_translation_en: true
   ddr_train_en: true
   aiclk_ppm_en: true
   watchdog_en: false
   smbus_en: false
   fan_ctrl_en: true
+  harvesting_en: true
 }
 
 fan_table {
--- a/boards/tenstorrent/tt_blackhole/spirom_data_tables/P150B/fw_table.txt
+++ b/boards/tenstorrent/tt_blackhole/spirom_data_tables/P150B/fw_table.txt
@@ -18,12 +18,13 @@ chip_limits {
 
 feature_enable {
   cg_en: true
-  noc_translation_en: false
+  noc_translation_en: true
   ddr_train_en: true
   aiclk_ppm_en: true
   watchdog_en: false
   smbus_en: false
   fan_ctrl_en: false
+  harvesting_en: true
 }
 
 fan_table {
--- a/boards/tenstorrent/tt_blackhole/spirom_data_tables/P150C/fw_table.txt
+++ b/boards/tenstorrent/tt_blackhole/spirom_data_tables/P150C/fw_table.txt
@@ -18,12 +18,13 @@ chip_limits {
 
 feature_enable {
   cg_en: true
-  noc_translation_en: false
+  noc_translation_en: true
   ddr_train_en: true
   aiclk_ppm_en: true
   watchdog_en: false
   smbus_en: false
   fan_ctrl_en: false
+  harvesting_en: true
 }
 
 fan_table {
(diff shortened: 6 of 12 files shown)


danieldegrasse:
Talking with @sbansalTT I think we would need to manually flash the `tt_boot_fs.bin` binary in order to actually touch the CMFWCFG tag, because `tt-flash` won't, right? If `tt-flash` will overwrite this than yes, this will work to turn on harvesting

Long term I would like to either:
- flash the CMFWCFG tag as part of CI and validate that the board works with the change
- use `tt-mod` in CI to write a temporary new CMFW configuration

danieldegrasse:
> Yes, tt-flash overwrites CMFWCFG by default.

Ah perfect, than yes- this should work for testing

danieldegrasse:
> Ah perfect, than yes- this should work for testing

Interesting note- it seems that with harvesting enabled, the SMC E2E test (which loads to CSM and does not flash the CMFW config) fails on the P100a- noticed it on #35 after CI ran on this PR. Any idea why that would happen?

afongTT:
> > Ah perfect, than yes- this should work for testing
> 
> Interesting note- it seems that with harvesting enabled, the SMC E2E test (which loads to CSM and does not flash the CMFW config) fails on the P100a- noticed it on #35 on the P100a after CI ran on this PR. #35 passed after I flashed a known good FW onto the P100A card. Any idea why that would happen?

It's likely the FW does not boot correctly when the NOC is not in a clean state. Especially since the `init()` routine expects NOC translation to be disabled, and enables it at the end.

afongTT:
> @danieldegrasse can you use this for making sure tt-metal can handle our breaking change?

I guess we're not waiting to see if tt-metal can handle this and releasing it to them as an RC? Moved out of draft.

cfriedt:
> It's likely the FW does not boot correctly when the NOC is not in a clean state. Especially since the `init()` routine expects NOC translation to be disabled, and enables it at the end.

@afongTT - is there any way we can detect if the NOC is not in a clean state at boot and just soft (or even hard) reset?

I feel like that could solve some subclass of problems. It might double the setup time in the worst scenario, and enumeration might not happen, but maybe it would miss enumeration anyway (or maybe we could still enumerate 🤷🏼‍♂️)

We have reset drivers for every IP block, but also can tell the BMC to do a hard reset if needed.

cfriedt:
If @palexsonTT 's auto reset / wdt feature goes in, and the NOC is hung, would that prevent ARC from feeding the wdt?

Seems like it could actionthe hard reset issue pretty nicely (almost as well as an on-chip WDT 😏)

cfriedt:
@afongTT - we should likely make a PR to Metal to support these firmware features (if there isn't already a PR).

Once that is done, we should
1. update the git ref for Metal that this repo points to (something like `refs/pull/<pr_number>` instead of `v0.55.0`)
2. verify that CI passes using the new Metal PR
3. merge the PR to Metal
4. update the git ref for Metal in this repo with the post-merge commit
5. merge this commit

That's how Zephyr repos typically manage changes that span 2 repositories in a non-breaking way.

The only caveat is that the change to Metal needs to work even before this PR is merged.

afongTT:
> > It's likely the FW does not boot correctly when the NOC is not in a clean state. Especially since the `init()` routine expects NOC translation to be disabled, and enables it at the end.
> 
> @afongTT - is there any way we can detect if the NOC is not in a clean state at boot and just soft (or even hard) reset?
> 
> I feel like that could solve some subclass of problems. It might double the setup time in the worst scenario, and enumeration might not happen, but maybe it would miss enumeration anyway (or maybe we could still enumerate 🤷🏼‍♂️)
> 
> We have reset drivers for every IP block, but also can tell the BMC to do a hard reset if needed.

The only reason we have unclean state at boot is because we don't have BH ASIC reset via a pin on any board other than p100. This causes our SMC E2E test to run only having reset ARC. I think our best way to handle this is to create a mechanism to direct BMC to do a BH ASIC reset without the JTAG portions.

> If @palexsonTT 's auto reset / wdt feature goes in, and the NOC is hung, would that prevent ARC from feeding the wdt?
> 
> Seems like it could action the hard reset pretty nicely (almost as well as an on-chip WDT 😏)

This reset would reload to the FW that is in the SPI, and not the code that you wanted to run in the smc-e2e test.

cfriedt:
> The only reason we have unclean state at boot is because we don't have BH ASIC reset via a pin on any board other than p100. This causes our SMC E2E test to run only having reset ARC. I think our best way to handle this is to create a mechanism to direct BMC to do a BH ASIC reset without the JTAG portions.

Wouldn't a BH ASIC reset be like a power-cycle / also require the JTAG workaround to be done again?

> > If @palexsonTT 's auto reset / wdt feature goes in, and the NOC is hung, would that prevent ARC from feeding the wdt?
> > Seems like it could action the hard reset pretty nicely (almost as well as an on-chip WDT 😏)
> 
> This reset would reload to the FW that is in the SPI, and not the code that you wanted to run in the smc-e2e test.

I think that's what we want, no? If the smc-e2e test isn't run because of a WDT timeout or if the wrong firmware runs, the test should fail.

cfriedt:
It would be good to have some tests in CI to verify that harvesting is working too (e.g. ZTests). Ideally, any new features we add would include test code to ensure that there is continuous test coverage.

Adding a new testcase to https://github.com/tenstorrent/tt-zephyr-platforms/tree/main/tests/lib/tenstorrent/bh_arc/src would be a good step.

We could use https://github.com/tenstorrent/tt-zephyr-platforms/pull/79 to ensure that tests are run with harvesting and NOC translation options built directly into the binary. That way, there would be certainty that we aren't accidentally relying on stale firmware tables from a previous flash operation.

afongTT:
> It would be good to have some tests in CI to verify that harvesting is working too (e.g. ZTests). Ideally, any new features we add would include test code to ensure that there is continuous test coverage.

I don't have any CI test cases for harvesting right now, but I have manually tested it on our internal infrastructure. Right now, I don't have a good idea of what it would even mean to unit test harvesting (other than checking that we harvested the right tiles). Perhaps it would need to fall into some kind of end to end test.

afongTT:
This change breaks subsequent p100a smoke-e2e-test runs due to the issue I previously mentioned:
> The only reason we have unclean state at boot is because we don't have BH ASIC reset via a pin on any board other than p100. This causes our SMC E2E test to run only having reset ARC. I think our best way to handle this is to create a mechanism to direct BMC to do a BH ASIC reset without the JTAG portions.

The specific issue in this case is (likely) that NOC translation stays enabled, and our HwInit assumes NOC translation is disabled. There may be a way to work around this for CI, but this is going to keep taking us down the rabbit hole of trying to restore the chip to a clean state during init, which is normally handled by the ASIC reset pin.

Due to the urgency of needing this change (which has been delayed for many weeks at this point), I advocate for merging it anyway and kicking off this release. There will be a lot of manual hand-holding of this release.

cfriedt:
> This change breaks subsequent p100a smoke-e2e-test runs due to the issue I previously mentioned:

> The specific issue in this case is (likely) that NOC translation stays enabled, and our HwInit assumes NOC translation is disabled. There may be a way to work around this for CI, but this is going to keep taking us down the rabbit hole of trying to restore the chip to a clean state during init, which is normally handled by the ASIC reset pin.

Can we disable NOC translation on init until it's safe to enable it? That's not really a workaround, it's more about  deterministic init.

```cpp
static int disable_noc(void)
{
...
}

SYS_INIT(disable_noc, EARLY, 0);
```

> Due to the urgency of needing this change (which has been delayed for many weeks at this point), **I advocate for merging it anyway** and kicking off this release. There will be a lot of manual hand-holding of this release.

There are many changes that are delayed for months. Let's do releases the right way.

One way to _not_ do things right is asking to cut a release on the same day.

I don't think anything should be tagged until
a) on-diff CI is green on all platforms
b) hardware-long is green on all platforms

Why do we need a "release" at all here? Why not just roll a firmware bundle manually?

> [!NOTE]  
> Anyone can roll a firmware bundle at any time using `./scripts/manual-release.sh`. The name of the script is possibly a misnomer - perhaps a better name is just `build-fwbundle.sh`.

cfriedt:
This seems perfectly fine if we're just talking about an RC - thanks for clarifying. Final release, I think we should def have CI passing.

1. Approve / merge 
2. Create PR to print git hash of build @cfriedt
3. Bump versions (separate prs) @afongTT 
4. Tag @cfriedt

cfriedt (approved):

PR #40, opened 2025-03-24
support building docker container within CI

danieldegrasse:
Add a job to build the docker container within CI. This should allow us to rebuild the metal container whenever an update is merged that changes the metal revision


cfriedt:
@danieldegrasse - this should be fine to merge soon, I would guess.

PR #41, opened 2025-03-25
logging: Change default log from DBG to INFO

abuckTT:
Per https://github.com/tenstorrent/tt-zephyr-platforms/pull/32#discussion_r2012486302
Change default log level to INFO rather than DBG for production.


cfriedt (approved):

cfriedt:
Does not hurt, but I wonder if this is something that should be added to the `release.yml` CI workflow, because for just about every other use case, we probably do want debug level output.

E.g. with PCIe uart we have plenty of bandwidth and would like to have automated crash dumps for post-mortem analysis.

For the `release.yml` CI workflow, we would just add command line switches like `CONFIG_FOO_LOG_LEVEL_DBG=n CONFIG_FOO_LOG_LEVEL_INFO=y`

PR #42, opened 2025-03-25
bh_arc: Check GDDR telemetry table version

abuckTT:
This change will check the GDDR telemetry table version and ensure it matches the gddr_telemetry_table.h version. If not, it will log warning message and keep all the GDDR telemetry fields as 0. 
Justification is in case we ever have a mismatch in MRISC and SMC FWs, we don't want to make assumptions that may no longer hold.

### DIFF
--- a/lib/tenstorrent/bh_arc/telemetry.c
+++ b/lib/tenstorrent/bh_arc/telemetry.c
@@ -25,6 +25,9 @@
 
 #include <tenstorrent/post_code.h>
 #include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+
+LOG_MODULE_REGISTER(telemetry, CONFIG_TT_APP_LOG_LEVEL);
 
 struct telemetry_entry {
 	uint16_t tag;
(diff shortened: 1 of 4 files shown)


lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -82,7 +85,11 @@ static void UpdateGddrTelemetry(void)
 		gddr_telemetry_table_t gddr_telemetry;
 		/* Harvested instances should read 0b00 for status. */
 		if (IS_BIT_SET(tile_enable.gddr_enabled, i)) {
-			read_gddr_telemetry_table(i, &gddr_telemetry);
+			if (read_gddr_telemetry_table(i, &gddr_telemetry)) {
```
cfriedt, line 88:
```suggestion
			if (read_gddr_telemetry_table(i, &gddr_telemetry) < 0) {
```

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -156,9 +163,13 @@ static void write_static_telemetry(uint32_t app_version)
 		/* Use first available instance. */
 		uint32_t gddr_inst = find_lsb_set(tile_enable.gddr_enabled) - 1;
 
-		read_gddr_telemetry_table(gddr_inst, &gddr_telemetry);
-		telemetry[GDDR_FW_VERSION] = (gddr_telemetry.mrisc_fw_version_major << 16) |
-			gddr_telemetry.mrisc_fw_version_minor;
+		if (read_gddr_telemetry_table(gddr_inst, &gddr_telemetry)) {
```
cfriedt, line 166: Same here
```suggestion
		if (read_gddr_telemetry_table(gddr_inst, &gddr_telemetry) < 0) {
```

lib/tenstorrent/bh_arc/telemetry.c
abuckTT: Noted and fixed. Also fixed in fan_ctrl.c

cfriedt (approved):

PR #49, opened 2025-03-26
bh_arc: regulator: init programming with readback verification

afongTT:
Read back regulator initialization programming values to ensure they are written properly. Retry once (per CMD) if programming fails. Aggregate all failures and return error status from `RegulatorInit()`.

### DIFF
--- a/lib/tenstorrent/bh_arc/dw_apb_i2c.c
+++ b/lib/tenstorrent/bh_arc/dw_apb_i2c.c
@@ -85,6 +85,7 @@
 /* IC abort source */
 /* starting from bit21, bit20-0 is reserved. */
 #define IC_ABRT_A3_STATE (0x1 << 21)
+#define IC_VERIFY_FAIL (0x1 << 22)
 
 #define GET_I2C_OFFSET(REG_NAME) DW_APB_I2C_##REG_NAME##_REG_OFFSET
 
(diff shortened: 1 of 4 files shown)


afongTT:
In another PR, will handle the case where RegulatorInit() returns with an error status.

tcolacoTT:
@afongTT  which PR relates to this? "In another PR, will handle the case where RegulatorInit() returns with an error status."

afongTT:
> @afongTT which PR relates to this? "In another PR, will handle the case where RegulatorInit() returns with an error status."

That PR doesn't exist yet.

alewycky-tenstorrent:
Now I understand your question about naming these. The PMBus spec gives all these commands names, even if many are useless (e.g. USER_DATA_00, MFR_SPECIFIC_C4). You could rename the structures e.g. vcore_data_USER_DATA_00 and then also have `#define PMBUS_CMD_USER_DATA_00 0xB0` and REGULATOR_DATA could pick up the command code from there.

afongTT:
> Now I understand your question about naming these. The PMBus spec gives all these commands names, even if many are useless (e.g. USER_DATA_00, MFR_SPECIFIC_C4). You could rename the structures e.g. vcore_data_USER_DATA_00 and then also have `#define PMBUS_CMD_USER_DATA_00 0xB0` and REGULATOR_DATA could pick up the command code from there.

Should I do this? It doesn't seem to add much value for most commands.

cfriedt:
We should really start moving a lot of this to Zephyr drivers.

Basically all of the issues I'm seeing these days are from the bh_arc library.

alewycky-tenstorrent:
> > Now I understand your question about naming these. The PMBus spec gives all these commands names, even if many are useless (e.g. USER_DATA_00, MFR_SPECIFIC_C4). You could rename the structures e.g. vcore_data_USER_DATA_00 and then also have `#define PMBUS_CMD_USER_DATA_00 0xB0` and REGULATOR_DATA could pick up the command code from there.
> 
> Should I do this? It doesn't seem to add much value for most commands.

Actually no, it only helps with 2 commands.

alewycky-tenstorrent (approved):

PR #51, opened 2025-03-27
bh_arc: do not enable aiclk_ppm if regulator init fails

afongTT:
It is potentially unsafe to enable aiclk_ppm if the regulator init fails. Report this scenario via LOG_ERR and into a status register.

This needs #49 to be merged first.

### DIFF
--- a/app/smc/src/main.c
+++ b/app/smc/src/main.c
@@ -9,6 +9,8 @@
 #include "init_common.h"
 #include "smbus_target.h"
 #include "telemetry.h"
+#include "status_reg.h"
+#include "reg.h"
 
 #include <stdint.h>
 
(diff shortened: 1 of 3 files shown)


cfriedt:
suggestion: start writing Zephyr drivers.

alewycky-tenstorrent:
Can you have InitHW return something useful instead of 0?
There are two recipients for this sort of info: FW application and SW tools. RU registers are not our primary interface to either of these.

alewycky-tenstorrent:
> suggestion: start writing Zephyr drivers.

How does Zephyr propagate a device init failure like this?

cfriedt:
> > suggestion: start writing Zephyr drivers.
> 
> How does Zephyr propagate a device init failure like this?

Normally, there would be some log messages, but additionally, API calls that pass in uninitialized device pointers will fail.

Ultimately, the application should ensure that all devices that are necessary for use are initialized (so e.g. some kind of check in a built-in self test).

afongTT:
> Ultimately, the application should ensure that all devices that are necessary for use are initialized (so e.g. some kind of check in a built-in self test).
Would our intention be that if we fail a self-test, we go into some kind of recovery mode? e.g. a mode where we can still update the FW, other functionality is shut down / in a safe mode?

afongTT:
> Can you have InitHW return something useful instead of 0? There are two recipients for this sort of info: FW application and SW tools. RU registers are not our primary interface to either of these.

It doesn't seem like Zephyr does much with the result of a SYS_INIT function:
https://github.com/zephyrproject-rtos/zephyr/blob/7823374e872145b5bd018bfe447839eb36042611/kernel/init.c#L371

It looks like it just saves the results to trace (when tracing is enabled)?

cfriedt (approved):

cfriedt:
@afongTT - it looks like this needs a rebase.

danieldegrasse:
Rebased and force pushed to attempt to resolve the CI failure seen on the VUART sample

cfriedt (approved):

PR #53, opened 2025-03-27
bh_arc: fix gddr soft harvesting logic

afongTT:
We want to soft harvest one GDDR when the actual GDDR count is greater than the desired GDDR count.

### DIFF
--- a/lib/tenstorrent/bh_arc/harvesting.c
+++ b/lib/tenstorrent/bh_arc/harvesting.c
@@ -212,8 +212,10 @@ void CalculateHarvesting(void)
 			tile_enable.eth_enabled = 0;
 		}
 
-		if (8 - get_fw_table()->product_spec_harvesting.dram_disable_count >
-			POPCOUNT(tile_enable.gddr_enabled)) {
+		uint8_t desired_gddr_count =
+			8 - get_fw_table()->product_spec_harvesting.dram_disable_count;
+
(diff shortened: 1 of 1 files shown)


lib/tenstorrent/bh_arc/harvesting.c
```diff
@@ -212,8 +212,10 @@ void CalculateHarvesting(void)
 			tile_enable.eth_enabled = 0;
 		}
 
-		if (8 - get_fw_table()->product_spec_harvesting.dram_disable_count >
-			POPCOUNT(tile_enable.gddr_enabled)) {
+		uint8_t desired_gddr_count =
+			8 - get_fw_table()->product_spec_harvesting.dram_disable_count;
+
+		if (POPCOUNT(tile_enable.gddr_enabled) > desired_gddr_count) {
```
danieldegrasse, line 218: Should we be changing the harvesting logic here in case GDDR3 was harvested during SLT or ATE? Maybe something for another PR though.

danieldegrasse (approved):

lib/tenstorrent/bh_arc/harvesting.c
afongTT: Yeah, I guess it's unwritten, but I only intended to ever use dram_disable_count = 0 or 1. But perhaps I should've done a more generic solution here.

PR #55, opened 2025-03-28
ci: ensure workflows run after push to main

cfriedt:
We ideally would prefer to see runner history on `main` instead of just on PR history. This will allow us to more easily determine the last successful CI run, and thus help narrow-down PRs that cause regressions, without having to do a full git bisect.

Additionally, run checks whenever PRs (backported bugfixes) have been made or merged to version branches.

Additionally, disable fast-failing for hardware tests so that we can see which platforms pass or fail tests independently. Without this, if one platform fails tests, the job is cancelled for all platforms.

### DIFF
--- a/.github/workflows/compliance.yml
+++ b/.github/workflows/compliance.yml
@@ -10,6 +10,7 @@ on:
       - edited
     branches:
       - main
+      - v*-branch
 
 jobs:
   check_compliance:
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -9,6 +9,7 @@ jobs:
   hardware-metal-test:
     if: github.repository_owner == 'tenstorrent'
     strategy:
+      fail-fast: false
       matrix:
         config:
           - board: p100
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -5,6 +5,10 @@ on:
     secrets:
       SIGNATURE_KEY:
         required: true
+  push:
+    branches:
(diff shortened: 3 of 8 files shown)


afongTT (dismissed):
Seems like a good thing to have.

.github/workflows/compliance.yml
danieldegrasse, line 1: I think we can skip this check (and maybe the license/copyright check) on main- all those are based on checking the diff with a PR, right? I think the builds and HW tests definitely should run on main though

.github/workflows/hardware-long.yml
```diff
@@ -2,6 +2,10 @@ name: Hardware Long Tests
 
 # Run long tests once nightly, at 00:00
 on:
+  push:
```
danieldegrasse, line 5: This test takes ~15 minutes, sometimes more. Do we want to run it on every merge?

cfriedt: Woops - good catch. I think just release branches.

danieldegrasse: Schedule should be against tip of tree

.github/workflows/compliance.yml
cfriedt: Sure

.github/workflows/hardware-long.yml
cfriedt: Fixed. Yeah, I didn't realize that it only runs against the default branch.

alewycky-tenstorrent (approved):

PR #56, opened 2025-03-28
bh_arc: Implement voltage and margin frequency

alexanderlayTT:
This updates voltage frequency curve to use voltage and frequency margins from firmware table.
Sets placeholder voltage margin as 50mV for each board type

### DIFF
--- a/lib/tenstorrent/bh_arc/vf_curve.c
+++ b/lib/tenstorrent/bh_arc/vf_curve.c
@@ -6,10 +6,23 @@
 
 #include <zephyr/sys/util.h>
 #include "vf_curve.h"
+#include "fw_table.h"
+
+/* Bounds checks for frequency and voltage margin */
+#define FREQ_MARGIN_MAX    300.0F
+#define FREQ_MARGIN_MIN    -300.0F
+#define VOLTAGE_MARGIN_MAX 150.0F
+#define VOLTAGE_MARGIN_MIN -150.0F
+
+static float freq_margin_mhz;
+static float voltage_margin_mv;
 
 void InitVFCurve(void)
 {
-	/* Do nothing for now */
+	freq_margin_mhz = CLAMP(get_fw_table()->chip_limits.frequency_margin, FREQ_MARGIN_MIN,
+				FREQ_MARGIN_MAX);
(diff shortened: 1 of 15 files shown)


lib/tenstorrent/bh_arc/vf_curve.c
```diff
@@ -20,7 +21,12 @@ void InitVFCurve(void)
  */
 float VFCurve(float freq_mhz)
 {
-	float voltage_mv = 0.00031395F * freq_mhz * freq_mhz - 0.43953F * freq_mhz + 828.83F;
+	float freq_margin_mhz = get_fw_table()->chip_limits.frequency_margin;
```
afongTT, line 24: Can you limit this value using [CLAMP](https://docs.zephyrproject.org/apidoc/latest/group__sys-util.html#gad6e7d4f6ba5a77f9ee3c04026f1c2b67)?
e.g. `freq_margin_mhz = CLAMP(freq_margin_mhz, -300.0F, 300.0F)`

And make freq_margin_mhz a file global `static float`, and initialize it during InitVFCurve?

lib/tenstorrent/bh_arc/vf_curve.c
```diff
@@ -20,7 +21,12 @@ void InitVFCurve(void)
...
+	float freq_with_margin_mhz = freq_mhz + freq_margin_mhz;
 
-	return voltage_mv + 50.0F; /* Add 50 mV of margin */
+	float voltage_margin_mv = get_fw_table()->chip_limits.voltage_margin;
```
afongTT, line 27: Also limit voltage margin, maybe +/- 150 mV. Same idea as above.

afongTT (changes requested):
Could you update the fw_table.proto so that voltage and frequency margin are [int32](https://protobuf.dev/programming-guides/proto3/#scalar)?

Generally we shouldn't update the types of entries in the .proto files, but converting from uint32 to int32 should be safe-ish according to the [Protobuf docs](https://protobuf.dev/programming-guides/proto3/#updating).

Copy the updated fw_table.proto to t6py as well.

alexanderlayTT:
I have addressed the requested changes in the latest commits - PTAL

afongTT (approved):
Looks good, thanks!

afongTT:
Just want to wait for CI to pass before merging.

PR #58, opened 2025-03-28
lib: tenstorrent: add I2C timeouts on all operations, and fix DVFS library issue with VUART

danieldegrasse:
This PR fixes two issues blocking CI:
- Add bus timeouts to *all* I2C operations that can hang, and recover the I2C bus when they do. This is admittedly a bit of voodoo- all I can say is without this change I saw the CMFW hanging during I2C transactions during init, and with it I don't. Solves issues like those seen here: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14138868875/job/39616640036
- Increase the scheduling deadline for DVFS polling to 10ms. This was done because it appears that DVFS polling itself takes over 1 ms, so rescheduling it this frequently was causing VUART to fail.

### DIFF
--- a/lib/tenstorrent/bh_arc/dvfs.c
+++ b/lib/tenstorrent/bh_arc/dvfs.c
@@ -51,5 +51,5 @@ void InitDVFS(void)
 
 void StartDVFSTimer(void)
 {
-	k_timer_start(&dvfs_timer, K_MSEC(1), K_MSEC(1));
+	k_timer_start(&dvfs_timer, K_MSEC(10), K_MSEC(10));
 }
--- a/lib/tenstorrent/bh_arc/dw_apb_i2c.c
+++ b/lib/tenstorrent/bh_arc/dw_apb_i2c.c
@@ -255,9 +255,19 @@ static void I2CRecoverBus(uint32_t id)
 static void WaitTxFifoEmpty(uint32_t id)
 {
 	uint32_t ic_status = 0;
+	uint64_t ts = k_uptime_get();
 
 	do {
 		ic_status = ReadReg(GetI2CRegAddr(id, GET_I2C_OFFSET(IC_STATUS)));
+
+		if (IS_ENABLED(CONFIG_TT_BH_ARC_I2C_TIMEOUT)) {
+			if ((k_uptime_get() - ts) > COND_CODE_1(CONFIG_TT_BH_ARC_I2C_TIMEOUT,
+							(CONFIG_TT_BH_ARC_I2C_TIMEOUT_DURATION),
+							(0))) {
+				I2CRecoverBus(id);
+				break;
+			}
+		}
 	} while ((ic_status & DW_APB_I2C_IC_STATUS_TFE_MASK) == 0);
 }
 
@@ -282,9 +292,18 @@ static void WaitTxFifoNotFull(uint32_t id)
 static void WaitMasterIdle(uint32_t id)
 {
 	uint32_t ic_status = 0;
+	uint64_t ts = k_uptime_get();
 
 	do {
 		ic_status = ReadReg(GetI2CRegAddr(id, GET_I2C_OFFSET(IC_STATUS)));
+		if (IS_ENABLED(CONFIG_TT_BH_ARC_I2C_TIMEOUT)) {
+			if ((k_uptime_get() - ts) > COND_CODE_1(CONFIG_TT_BH_ARC_I2C_TIMEOUT,
+							(CONFIG_TT_BH_ARC_I2C_TIMEOUT_DURATION),
+							(0))) {
+				I2CRecoverBus(id);
+				break;
+			}
+		}
 	} while (ic_status & DW_APB_I2C_IC_STATUS_MST_ACTIVITY_MASK);
 }
 
@@ -315,6 +334,7 @@ static uint32_t WaitAllTxDone(uint32_t id)
 {
 	uint32_t master_active = 0;
 	uint32_t tx_fifo_not_empty = 0;
+	uint64_t ts = k_uptime_get();
 
 	do {
 		uint32_t ic_error = CheckTxAbrt(id);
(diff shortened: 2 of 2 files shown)


danieldegrasse:
This should pass on the P100. Still looking at the P100A runner- it seems to be unable to flash right now

cfriedt:
@danieldegrasse - can you rebase? The latest main will ensure that jobs for p100 are not cancelled by GitHub CI.

cfriedt:
I'm definitely seeing ARC i2c hanging on p100.
```shell
#0  I2CReadRxFifo (id=id@entry=1, p_read_buf=p_read_buf@entry=0x1002b19c <z_main_stack+1856> "\b\004\003\200\342y")
    at /home/ubuntu/cfriedt/tt-zephyr-platforms-work/tt-zephyr-platforms.git/lib/tenstorrent/bh_arc/dw_apb_i2c.c:345
#1  0x10008288 in I2CTransaction (read_len=<optimized out>, read_data=<optimized out>, write_len=<optimized out>, 
    write_data=<optimized out>, id=<optimized out>)
    at /home/ubuntu/cfriedt/tt-zephyr-platforms-work/tt-zephyr-platforms.git/lib/tenstorrent/bh_arc/dw_apb_i2c.c:476
#2  I2CTransaction (id=1, write_data=0x1002b188 <z_main_stack+1836> "\001", write_len=1, 
    read_data=0x1002b19c <z_main_stack+1856> "\b\004\003\200\342y", read_len=1)
    at /home/ubuntu/cfriedt/tt-zephyr-platforms-work/tt-zephyr-platforms.git/lib/tenstorrent/bh_arc/dw_apb_i2c.c:441
#3  0x100082fe in I2CReadBytes (id=id@entry=1, command=<optimized out>, command@entry=1, command_byte_size=command_byte_size@entry=1, 
    p_read_buf=p_read_buf@entry=0x1002b19c <z_main_stack+1856> "\b\004\003\200\342y", data_byte_size=data_byte_size@entry=1, 
    flip_bytes=flip_bytes@entry=0 '\000')
    at /home/ubuntu/cfriedt/tt-zephyr-platforms-work/tt-zephyr-platforms.git/lib/tenstorrent/bh_arc/dw_apb_i2c.c:505
#4  0x1000694a in SwitchVoutControl (source=source@entry=AVSVoutCommand)
    at /home/ubuntu/cfriedt/tt-zephyr-platforms-work/tt-zephyr-platforms.git/lib/tenstorrent/bh_arc/regulator.c:214
#5  0x100079e2 in InitHW () at /home/ubuntu/cfriedt/tt-zephyr-platforms-work/tt-zephyr-platforms.git/lib/tenstorrent/bh_arc/init.c:462
#6  0x1000add0 in z_sys_init_run_level (level=level@entry=INIT_LEVEL_APPLICATION)
    at /home/ubuntu/cfriedt/tt-zephyr-platforms-work/zephyr/kernel/init.c:371
#7  0x1000adfa in bg_thread_main (unused1=unused1@entry=0x0, unused2=unused2@entry=0x0, unused3=unused3@entry=0x0)
    at /home/ubuntu/cfriedt/tt-zephyr-platforms-work/zephyr/kernel/init.c:536
#8  0x10000ffa in z_thread_entry (entry=0x1000add4 <bg_thread_main>, p1=0x0, p2=0x0, p3=0x0)
    at /home/ubuntu/cfriedt/tt-zephyr-platforms-work/zephyr/lib/os/thread_entry.c:48
#9  0x100022f4 in z_arch_get_next_switch_handle (old_thread=<optimized out>)
    at /home/ubuntu/cfriedt/tt-zephyr-platforms-work/zephyr/arch/arc/core/thread.c:215
```

cfriedt:
Definitely an improvement. Thanks @danieldegrasse! 

Just for clarification, the vuart wasn't causing any issues, as #60 disabled the vuart test entirely and CI was still failing everywhere. Also, vuart is not yet enabled by default (although we really should enabled it by default soon).

The I2C hang was what I was seeing when manually debugging on a `p100` system.

cfriedt (approved):

afongTT:
I think this issue came up because previously we were only issuing i2c writes as the first transactions on boot. The change with I2CRMWV caused an I2C read to happen earlier, so the old workaround detecting I2C timeout only on TX was no longer sufficient.

Of course, this is just the workaround for the i2c hanging in the first place. From what we've seen in the past, that happens when our chip gets reset while in the middle of an i2c transaction. We've avoided this issue in the past by sending an I2C idle message (msg 0xA3) to SMC before all resets. e.g. on a system reboot, ttkmd would send this message. I'm not sure if we currently have cases in our CI where we'd reset BH without sending the 0xA3 message, but it's likely that we do. (Note `tt-smi -r` should already handle this).

danieldegrasse:
> Of course, this is just the workaround for the i2c hanging in the first place. From what we've seen in the past, that happens when our chip gets reset while in the middle of an i2c transaction. We've avoided this issue in the past by sending an I2C idle message (msg 0xA3) to SMC before all resets. e.g. on a system reboot, ttkmd would send this message. I'm not sure if we currently have cases in our CI where we'd reset BH without sending the 0xA3 message, but it's likely that we do. (Note tt-smi -r should already handle this).

We definitely do have these cases in CI- any time we flash via JTAG :). Good to know though- do you think this is just a fact of life we will have to deal with?

afongTT:
> > Of course, this is just the workaround for the i2c hanging in the first place. From what we've seen in the past, that happens when our chip gets reset while in the middle of an i2c transaction. We've avoided this issue in the past by sending an I2C idle message (msg 0xA3) to SMC before all resets. e.g. on a system reboot, ttkmd would send this message. I'm not sure if we currently have cases in our CI where we'd reset BH without sending the 0xA3 message, but it's likely that we do. (Note tt-smi -r should already handle this).
> 
> We definitely do have these cases in CI- any time we flash via JTAG :). Good to know though- do you think this is just a fact of life we will have to deal with?

I guess we could try always starting FW with an i2c bitbang clear?

danieldegrasse:
> > > Of course, this is just the workaround for the i2c hanging in the first place. From what we've seen in the past, that happens when our chip gets reset while in the middle of an i2c transaction. We've avoided this issue in the past by sending an I2C idle message (msg 0xA3) to SMC before all resets. e.g. on a system reboot, ttkmd would send this message. I'm not sure if we currently have cases in our CI where we'd reset BH without sending the 0xA3 message, but it's likely that we do. (Note tt-smi -r should already handle this).
> > 
> > 
> > We definitely do have these cases in CI- any time we flash via JTAG :). Good to know though- do you think this is just a fact of life we will have to deal with?
> 
> I guess we could try always starting FW with an i2c bitbang clear?

Yeah true- that isn't much different than what we have with this PR though, right? I guess this won't be something we run into during normal operation, just when the chip crashes or is reset in CI.

afongTT:
> > I guess we could try always starting FW with an i2c bitbang clear?
> 
> Yeah true- that isn't much different than what we have with this PR though, right? I guess this won't be something we run into during normal operation, just when the chip crashes or is reset in CI.

One difference is that we wouldn't lose the first i2c transaction (and thus hit HW init failure). So we can still pass when we hit this situation in CI.

PR #60, opened 2025-03-29
tests: app: smc: disable app.vuart sample for now

cfriedt:
Something else in the system is broken badly. This test seems to be the most obvious one affected, so disable it for now.


cfriedt:
Closing, as it doesn't improve the CI situation.

PR #61, opened 2025-03-30
bh_arc: fw_table: enable loading from spi with kconfig

cfriedt:
We do not flash fw tables during the normal CI process.

This has the unfortunate consequence of introducing regressions if there is inadequate test coverage.

If we use a Kconfig option to enable or disable loading the firmware table from SPI, then we can be sure there is no hidden state from one build to the next.

This is a stop-gap solution until we have a complete unbrickable firmware update process, and a process by which we tightly couple the smc firmware image with firmware tables.


cfriedt:
This is effectively what the recovery fw is for, so closing.

PR #54, opened 2025-03-28
bh_arc: pvt: reduce TS polling time by reducing resolution

afongTT:
This reduces the amount of time spent polling in GetAvgChipTemp() which should help reduce FW busy-ness. This temporary fix allows logging to still be functional when aiclk_ppm/dvfs is enabled.

PR #63, opened 2025-03-31
dvfs: restore 1 ms period

afongTT:
The 1 ms period should no longer starve CPU time, due to the reduction in TS polling time in #54.


afongTT:
Draft until I add an E2E (?) test to make sure the application does go idle. Trying to figure out how to do that in an automated way.

cfriedt (approved):

PR #66, opened 2025-04-01
ci: update hardware-long to use prepare-zephyr reusable workflow

cfriedt:
Update the hardware-long.yaml workflow, which runs Metal tests, to use the `prepare-zephyr.yaml` reusable workflow.

This allows us to consolidate some of the boilerplate Zephyr setup to one location, and also fixes an issue with west init by removing existing data and using the conventional `zephyrproject-rtos/action-zephyr-setup@v1` GitHub action.

> [!NOTE]  
> The `cleanup` phase will cause the error below in the `Post Run` phase, so it was removed.

```shell
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/opt/tenstorrent/github/_work/tt-zephyr-platforms/tt-zephyr-platforms/tt-zephyr-platforms/.github/workflows/prepare-zephyr'. Did you forget to run actions/checkout before running your local action?
```

> [!NOTE]  
> Also needed to downgrade required cmake version in metal from v3.24..v3.30 to v3.20 to avoid the error below

```shell
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
-- Configuring incomplete, errors occurred!
  CMake 3.24 or higher is required.  You are running version 3.22.1
```

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -1,9 +1,23 @@
 name: Hardware Long Tests
 
 # Run long tests once nightly, at 00:00
+# on:
+#   schedule:
+#     - cron: "0 0 * * *"
+
 on:
-  schedule:
-    - cron: "0 0 * * *"
+  push:
+    branches:
+      - main
+      - v*-branch
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+    branches:
+      - main
+      - v*-branch
 
 jobs:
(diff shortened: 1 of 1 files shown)


cfriedt:
Bumped Metal revision to v0.56.0

danieldegrasse:
The long tests are unfortuntely going to fail until we get harvesting support in metal... the P100a has harvested tensix cores and won't work

PR #67, opened 2025-04-01
Added remaining p150 and p300 board revs

TTDRosen:
Adding p150{a,c} and p300{a,b,c} to the sysbuild


.github/workflows/build-fw.yml
```diff
@@ -43,6 +43,11 @@ jobs:
             p100) BOARD=tt_blackhole@p100/tt_blackhole/smc;;
             p100a) BOARD=tt_blackhole@p100a/tt_blackhole/smc;;
             p150a) BOARD=tt_blackhole@p150a/tt_blackhole/smc;;
+            p150b) BOARD=tt_blackhole@p150b/tt_blackhole/smc;;
+            p150c) BOARD=tt_blackhole@p150c/tt_blackhole/smc;;
+            p300a) BOARD=tt_blackhole@p300a/tt_blackhole/smc;;
+            p300b) BOARD=tt_blackhole@p300b/tt_blackhole/smc;;
+            p300c) BOARD=tt_blackhole@p300c/tt_blackhole/smc;;
```
afongTT, line 50: I think you need to add these other boards to the `strategy.matrix.board` list so that they build too.

TTDRosen: Fixes, now I just have to figure out why twister doesn't like the test identifiers

boards/tenstorrent/tt_blackhole/revision.cmake
```diff
@@ -1,5 +1,5 @@
 set(BOARD_REVISIONS
-    "p100" "p100a" "p150" "p150a" "p150b" "p150c" "p300" "p300a" "p300b" "p300c")
-if(NOT DEFINED CONFIG_TT_SMC_RECOVERY AND NOT BOARD_REVISION IN_LIST BOARD_REVISIONS)
+    "p100" "p100a" "p150" "p150a" "p150b" "p150c" "p300" "p300a" "p300b" "p300c" "recovery")
```
afongTT, line 2: What happens if I try to build the bmc app for the recovery board?

e.g.
`west build -b tt_blackhole@recovery/tt_blackhole/bmc app/bmc`

.github/workflows/build-fw.yml
```diff
@@ -108,10 +118,9 @@ jobs:
           west build \
             -d build-recovery \
             -p auto \
-            -b tt_blackhole/tt_blackhole/smc \
+            -b tt_blackhole@recovery/tt_blackhole/smc \
```
cfriedt, line 121: recovery isn't a board revision

.github/workflows/build-fw.yml
```diff
@@ -108,10 +118,9 @@ jobs:
...
             tt-zephyr-platforms/app/smc \
             -- \
-              -DEXTRA_CONF_FILE=recovery.conf \
```
cfriedt, line 114: this needs to go back

app/smc/sysbuild.cmake
```diff
@@ -85,3 +153,4 @@ add_custom_command(OUTPUT ${OUTPUT_FWBUNDLE}
 # firmware bundles whenever the SMC, BMC, or recovery binaries are
 # updated
 add_custom_target(fwbundle ALL DEPENDS ${OUTPUT_FWBUNDLE})
+endif()
```
cfriedt, line 156: needs a newline

boards/tenstorrent/tt_blackhole/revision.cmake
cfriedt: Whaat? Why would we have a recovery board? It's just adding a Kconfig option to existing boards. The recovery binary needs to exist for all boards still.

That's going a little out of scope for this PR, I think. Probably better to leave the recovery stuff as it is, since it would affect the release workflow, etc.

cfriedt (changes requested):
Making "recovery" a board revision seems kind of out of scope for this PR, and it's not really accurate as a board revision. Once we have unified BMFW, it will potentially remove the need to have board revisions, and at that point, a Kconfig for recovery would still be a better option (Kconfig to configure SW)

boards/tenstorrent/tt_blackhole/revision.cmake
TTDRosen: @afongTT it builds the base bmc fw (no revision). But in my mind it isn't any different from building smc with a bmc board. i.e. `west build -b tt_blackhole@p150/tt_blackhole/bmc app/smc`

@cfriedt I didn't like that the recovery image was generating the bootfs tables. So I decided to make it a revision instead. I also preferred that unlike the EXTRA_CONF_FILE the revision is set earlier in the build.

TTDRosen:
Resolved all chats

I'm still not super happy with the board definitions, but it is working so we are in better shape than before. I would suggest we take another pass at it once we have finished automating the releases and beefed up testing some more.

cfriedt:
Looks better. I'm hoping we can get CI stable again to run through tests properly to get this merged.

cfriedt:
@TTDRosen - this should probably pass CI if you rebase now 👍

TTDRosen:
Tests are finally passing 🎉

cfriedt (approved):

cfriedt:
LGTM - might be worth it to get some other reviews as well.

boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc_p100.yaml
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 identifier: tt_blackhole@p100/tt_blackhole/smc
-name: Tenstorrent Blackhole P100 board (SMC)
+name: Tenstorrent Blackhole P100 (SMC)
```
afongTT, line 5: Why `(Scrappy)` in BMC board name but not SMC board name?

TTDRosen:
I'll give other reviewers until 5 EST to provide additional comment and if no one says no by then I'll merge

boards/tenstorrent/tt_blackhole/board.yml
```diff
@@ -17,3 +17,9 @@ boards:
     - name: "p300c"
   socs:
   - name: tt_blackhole
+
+- name: p300
```
afongTT, line 21: Is the goal to move everything to a board instead of revision?

TTDRosen: Where it makes sense, the reason that I moved the p300 is that the stm32 is different between the p300 and p1x0 boards. So I decided it didn't make sense to make it a revision

boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc_p100.yaml
TTDRosen: 🤷 do you have an opinion either way

TTDRosen: Made smc match bmc

afongTT (approved):

PR #68, opened 2025-04-02
readme: hard-code branch to main for status badges

cfriedt:
Hard-code branch to main for status badges.

Previously, the "default" was used which is to show the most recent status from any branch for a given action, which is not what we're interested in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,11 @@ west blobs fetch
 # Apply local patches
 west patch apply
 
-# Go to the main Zephyr directory
-cd zephyr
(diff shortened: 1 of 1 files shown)


cfriedt:
I might just merge this now, and then continue working on improving CI. Is that ok by you @alewycky-tenstorrent ?

alewycky-tenstorrent:
> I might just merge this now, and then continue working on improving CI. Is that ok by you @alewycky-tenstorrent ?

yes

PR #69, opened 2025-04-02
boards: pad bmfw to nearest 2048 byte boundary

danieldegrasse:
Due to a bug in the update code for BMFW 0.3.1, the erase size is determined by the size of the image rounded to the nearest write size boundary. On the STM32 based BMC, the write size != the erase size, so erases will fail based on this rounding (and updates will not work)

To work around this, pad the BMFW image to a 2048 byte boundary in the tt_boot_fs. This will enable updates to work, at the cost of a bit more flash space used

### DIFF
--- a/boards/tenstorrent/tt_blackhole/bootfs/p150a-bootfs.yaml
+++ b/boards/tenstorrent/tt_blackhole/bootfs/p150a-bootfs.yaml
@@ -41,6 +41,7 @@ images:
 
   # Board Mgmt FW
   - name: bmfw
+    padto: 0xb000
     binary: $BUILD_DIR/bmc/zephyr/zephyr.signed.bin
 
   # Meta data about the flashing process
--- a/scripts/tt_boot_fs.py
+++ b/scripts/tt_boot_fs.py
@@ -290,6 +290,7 @@ class BootImage:
                 raise ValueError(
                     f"{tag} padto value {padto} is < the binary size {len(binary)}"
                 )
+            binary += bytes(padto - len(binary))
         # We always need to pad binaries to 4 byte offsets for checksum verification
         binary += bytes((len(binary) % 8))
 
--- a/boards/tenstorrent/tt_blackhole/bootfs/p100-bootfs.yaml
+++ b/boards/tenstorrent/tt_blackhole/bootfs/p100-bootfs.yaml
@@ -41,6 +41,7 @@ images:
 
   # Board Mgmt FW
   - name: bmfw
+    padto: 0xb000
     binary: $BUILD_DIR/bmc/zephyr/zephyr.signed.bin
 
   # Meta data about the flashing process
--- a/boards/tenstorrent/tt_blackhole/bootfs/p100a-bootfs.yaml
+++ b/boards/tenstorrent/tt_blackhole/bootfs/p100a-bootfs.yaml
@@ -41,6 +41,7 @@ images:
 
   # Board Mgmt FW
   - name: bmfw
+    padto: 0xb000
     binary: $BUILD_DIR/bmc/zephyr/zephyr.signed.bin
 
   # Meta data about the flashing process


cfriedt (commented):
I don't think we should need to add a `padto` directive to any yaml files. Individual tt boot fs files should be rounded to an erase size by default, so I think the fix would be in `tt_boot_fs.py` exclusively.

danieldegrasse:
They are currently rounded to 8 byte offsets, right? Should we just increase the rounding to 2048?

cfriedt:
> They are currently rounded to 8 byte offsets, right? Should we just increase the rounding to 2048?

Not necessarily 2048. I think it's the `alignment` specified from the yaml file that we want to use (which is passed in as `alignment` to that function.

```yaml
alignment:
  flash_device_size: 0x1000000
  flash_block_size: 4096
```

The part where I believe it's already being rounded up is here:
https://github.com/tenstorrent/tt-zephyr-platforms/blob/c4de5858ab2da5b3f95f957cc6eddca6f5f16532/scripts/tt_boot_fs.py#L361

https://github.com/tenstorrent/tt-zephyr-platforms/blob/c4de5858ab2da5b3f95f957cc6eddca6f5f16532/scripts/tt_boot_fs.py#L666

It's very possible that it's inconsistent where you saw the `% 8`.

scripts/tt_boot_fs.py
```diff
@@ -290,6 +290,7 @@ def loads(tag: str, data: dict[str, Any], alignment: FileAlignment, env: dict):
                 raise ValueError(
                     f"{tag} padto value {padto} is < the binary size {len(binary)}"
                 )
+            binary += bytes(padto - len(binary))
```
cfriedt, line 293: I think we may also want to check if padto is a multiple of the `alignment` argument, and then default `padto` using the `_align_up()` function.

boards/tenstorrent/tt_blackhole/bootfs/p150a-bootfs.yaml
```diff
@@ -41,6 +41,7 @@ images:
 
   # Board Mgmt FW
   - name: bmfw
+    padto: 0xb000
```
cfriedt, line 44: these we probably don't need. It should be the default to align up to a block erase boundary

danieldegrasse:
Ok, this makes sense. I can update the PR to use the alignment value to determine the size and offset alignment, and that should fix this issue as well

cfriedt:
CI is sucking hard right now. TBH, I think it's the mcuboot test. It's very poorly conditioned. The test should be modified so that it erases the bootloader partition before writing the app, and then writes the bootloader, and then the app.

cfriedt:
@danieldegrasse - can you copy / paste some "testing done" in the PR description since CI is being flakey?

It would be ideal to update `tests/lib/tenstorrent/tt_boot_fs` with a testcase.

Maybe it makes sense to drop `tests/lib/tenstorrent/tt_boot_fs/pytest/p100.yml` in favour of `boards/tenstorrent/tt_blackhole/bootfs/p100-bootfs.yaml`. That would maybe require generating at least some dummy `.bin`s to test with. It's just `pytest` so shouldn't be too difficult.

afongTT:
I strongly prefer we don't pad every binary up to the alignment size (4KB). This creates FW loading overhead since every "file" loaded from the file system needs to load all these extra bytes.

Can we just pad the files that need padding, e.g. BMC app in this case?

cfriedt:
> I strongly prefer we don't pad every binary up to the alignment size (4KB). This creates FW loading overhead since every "file" loaded from the file system needs to load all these extra bytes.

Yeah, I also don't think we should add a `padto` entry for every image.

> Can we just pad the files that need padding, e.g. BMC app in this case?

I think that should be ok as a temporary workaround. Long term, we should probably move slot1 to spi flash entirely and use a fixed partition. That requires a spi flash "driver" for openocd / stm32 though, so it might not happen overnight.

danieldegrasse:
> I strongly prefer we don't pad every binary up to the alignment size (4KB). This creates FW loading overhead since every "file" loaded from the file system needs to load all these extra bytes.
> 
> Can we just pad the files that need padding, e.g. BMC app in this case?

Fine with me- so is the resolution here to keep using `padto`?

cfriedt:
> > I strongly prefer we don't pad every binary up to the alignment size (4KB). This creates FW loading overhead since every "file" loaded from the file system needs to load all these extra bytes.
> > Can we just pad the files that need padding, e.g. BMC app in this case?
> 
> Fine with me- so is the resolution here to keep using `padto`?

Yeah, I guess.

It still seems counterintuitive to me somehow. MCUboot has some "optimized" path for handling chunks full of pad bytes anyway, though and I'm pretty sure other companies like other hw vendors also pad their images with `0xff` in zephyr too.

It seems like it shouldn't be completely necessary though... where does this 2048 byte value actually come from? The application should know to pad to a certain size when writing pages (and it should be 4k), so there really shouldn't be a need to pad the image...

cfriedt (approved):

PR #70, opened 2025-04-02
Add GDDR HW Memory test

abuckTT:
- CI: Adds check in e2e test that hw_init status is "done" at the end
- GDDR: Adds triggering HW memory test core via MRISC FW at end of init. Will only run if MRISC FW supports this. Version >= 2.7 (not yet released).

### DIFF
--- a/lib/tenstorrent/bh_arc/gddr.c
+++ b/lib/tenstorrent/bh_arc/gddr.c
@@ -14,6 +14,7 @@
 #include "harvesting.h"
 
 #include <zephyr/logging/log.h>
+#include <zephyr/kernel.h>
 
 /* This is the noc2axi instance we want to run the MRISC FW on */
 #define MRISC_FW_NOC2AXI_PORT 0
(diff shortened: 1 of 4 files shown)


lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -145,3 +155,55 @@ uint32_t GetDramMask(void)
...
+	k_timepoint_t timeout = sys_timepoint_calc(K_MSEC(MRISC_MEMTEST_TIMEOUT));
+
+	while (MriscRegRead32(gddr_inst, MRISC_MSG_REGISTER) != 0) {
+		/* Wait for the message to be processed */
+		if (sys_timepoint_expired(timeout)) {
+			LOG_ERR("Timeout after %d ms waiting for GDDR instance %d to run "
+				"memtest.\n",
+				MRISC_MEMTEST_TIMEOUT, gddr_inst);
+			ret = -ETIMEDOUT;
+			break;
```
danieldegrasse, line 214: a `goto` would make more sense here- otherwise someone could assign to `ret` in the future in the code before and cause an issue

danieldegrasse (commented):
Can we split this into two commits?
- one commit adding the gddr memory test
- one commit adding the boot status check

Generally when a feature is introduced in a PR, we want to avoid subsequent commits that make changes to that feature

lib/tenstorrent/bh_arc/gddr.c
abuckTT: Updated the code to just return directly on failure and remove "ret" altogether.

danieldegrasse (approved):

cfriedt (approved):
Looks good to me.

Maybe this could be modified in the future to not require host interaction (i.e. use ZTest instead of or in addition to Pytest).

Another really good addition later would be to add a shell command, so that e.g. different memory ranges, test patterns, timings, etc, could be specified at runtime.

I.e. create an API for a generic memory test function, which could be wrapped by ZTest and the shell.

What do you think, @abuckTT ?

abuckTT:
@cfriedt definitely agree on creating a shell command that can be used to test the memory at runtime. I've created internal JIRA to track. I think end goal is to have this be part of a suite of tests that can be used to check HW in mission mode to determine whether the HW is good or failing.

PR #71, opened 2025-04-02
readme: add instructions for using tt-console

cfriedt:
Add instructions for building app/smc with the vuart overlay, and using tt-console to observe console output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,57 @@ source zephyr/zephyr-env.sh
 cd $MODULE.git
 ```
 
+### Build & Flash
+
+**Build and flash targets with `west`**
+
+> [!NOTE]
+> Below, we build with [application overlay](https://docs.zephyrproject.org/latest/develop/application/index.html#important-build-system-variables)
+> files to support `tt-console` from firmware.
+
+```shell
+# Build `tt-console`
+gcc -Iinclude -std=gnu11 -o tt-console scripts/tt-console/console.c
+
+# Build and flash firmware
+west build --sysbuild -p -b tt_blackhole@p100/tt_blackhole/smc app/smc
+west flash -r tt_flash --force
+
+# Reset the board and rescan the PCIe bus
+tt-smi -r
+./scripts/rescan-pcie.sh
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,6 +41,55 @@ source zephyr/zephyr-env.sh
...
+> files to support `tt-console` from firmware.
+
+```shell
+# Build `tt-console`
+gcc -Iinclude -std=gnu11 -o tt-console scripts/tt-console/console.c
+
+# Build and flash SMC firmware
+west build --sysbuild -p -b tt_blackhole@p100/tt_blackhole/smc app/smc -- -DEXTRA_DTC_OVERLAY_FILE="vuart.overlay" -DEXTRA_CONF_FILE="vuart.conf"
+./scripts/bmc-reset.py
+west flash
```
cfriedt, line 59: We might want to specify to use the `tt_flash` west runner here. What do you think @danieldegrasse ?

danieldegrasse: Yes, I would argue for `west flash -r tt_flash --force` here

danieldegrasse (commented):
Looks good to me. We could consider switching vuart on by default for the SMC app- I don't think that would be harmful, we just can't turn it on systemwide yet. The E2E tests for the SMC app could have the ` --device-serial-pty` argument added to override the PTY used for smoke tests, and use the PCIe VUART instead

cfriedt:
> Looks good to me. We could consider switching vuart on by default for the SMC app- I don't think that would be harmful, we just can't turn it on systemwide yet. The E2E tests for the SMC app could have the ` --device-serial-pty` argument added to override the PTY used for smoke tests, and use the PCIe VUART instead

Yeah, I was thinking the same - I already made a "hw-map-vuart.yml" file on the p100 CI machine. I could potentially give it a whirl here and update GH workflows in this PR if it Just Works (TM). The `tt-console` app is already statically built and lives in `/opt/tenstorrent/bin` for p100 and p100a machines.

danieldegrasse (approved):

PR #75, opened 2025-04-03
hotfix: revert commits causing CI failures

danieldegrasse:
From my best-effort bisect, these commits appear to be causing the CI hangs on the P100 platform. The CI hang seen can be reproduced as follows:
- build fwbundle for the P100 card: `west build -p always -b tt_blackhole@p100/tt_blackhole/smc tt-zephyr-platforms/app/smc/ --sysbuild`
- flash fwbundle using `west flash -r tt_flash --force`
- Make sure BMFW is updated by running `west flash -d build/bmc`
- reset the SMC repeatedly by running `west rtt -d build/bmc`. After 10-20 invocations, the message "BMFW VERSION 0.3.2" should fail to appear

If attaching with a debugger, you will see that the BMFW is stuck waiting for PGOOD to go high on the card. This can be solved by toggling A/C power to the system, but the issue will reappear within CI eventually.

### DIFF
--- a/lib/tenstorrent/bh_arc/regulator.c
+++ b/lib/tenstorrent/bh_arc/regulator.c
@@ -223,45 +223,26 @@ void SwitchVoutControl(VoltageCmdSource source)
 
 void RegulatorInit(PcbType board_type)
 {
-	if (board_type == PcbTypeP150) {
-		/* VCORE */
-		I2CInit(I2CMst, P0V8_VCORE_ADDR, I2CFastMode, PMBUS_MST_ID);
+	/* VCORE */
+	I2CInit(I2CMst, P0V8_VCORE_ADDR, I2CFastMode, PMBUS_MST_ID);
 
-		static const uint8_t vcore_b0[] = {0x15, 0x09, 0x3c, 0x08, 0x0a, 0x02, 0x0f, 0x00,
-						   0x11, 0x00, 0x00, 0x00, 0x00, 0x41, 0x03, 0x00,
-						   0x00, 0x0f, 0x0d, 0x0a, 0x00, 0x00};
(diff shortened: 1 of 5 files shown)


cfriedt:
Test methodology seems sane to me. We should eventually have a repeated reset test in CI as well and that should be able to run on-diff assuming it executes quickly enough, but that can come later.

If we can fix this forward, that would be preferable to reverting, but I just really want CI to be green again 😅

danieldegrasse:
> If we can fix this forward, that would be preferable to reverting, but I just really want CI to be green again 😅

Same. I'd like the following sequence:
- if this PR is reliable, merge it
- follow up PR with "hammer tests" to check for the condition that broke CI
- follow up PR reimplementing reverted features

lib/tenstorrent/bh_arc/regulator.c
```diff
@@ -224,172 +221,28 @@ void SwitchVoutControl(VoltageCmdSource source)
...
-		uint8_t cmd;
-		const uint8_t *data;
-		const uint8_t *mask;
-		uint32_t size;
-	} RegulatorData;
-
-	uint32_t aggregate_i2c_errors = 0;
-	uint32_t i2c_error = 0;
-
-	if (board_type == PcbTypeP150) {
```
afongTT, line 244: Maybe the problem was that I only applied these settings to PcbTypeP150 and not PcbTypeP100.

I can upload a PR that also applies them to P100.

cfriedt: @afongTT - go for it. That could be the problem 🤷‍♂️

danieldegrasse:
Superseded by #77

PR #80, opened 2025-04-03
ci: flash a stable copy of BMFW and CMFW prior to HW CI runs

danieldegrasse:
Flash a stable copy of BMFW and CMFW prior to HW smoke and e2e tests, so any bad state from a prior PR run is fixed.

### DIFF
--- a/.github/workflows/hardware-smoke.yml
+++ b/.github/workflows/hardware-smoke.yml
@@ -33,7 +33,7 @@ jobs:
         - /dev/hugepages-1G:/dev/hugepages-1G
         - /opt/tenstorrent/fw/stable:/opt/tenstorrent/fw/stable
         - /opt/tenstorrent/twister:/opt/tenstorrent/twister
-        - /opt/tenstorrent/bin/openocd-rtt:/opt/tenstorrent/bin/openocd-rtt
+        - /opt/tenstorrent/bin:/opt/tenstorrent/bin
       options: '--device /dev/tenstorrent --device /dev/bus/usb --privileged'
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/hardware-smoke.yml
```diff
@@ -44,6 +44,11 @@ jobs:
         with:
           app-path: tt-zephyr-platforms
 
+      - uses: ./tt-zephyr-platforms/.github/workflows/prepare-hardware
+        with:
+          board: ${{ matrix.config.board }}
+          version: '80.15.0.0'
```
cfriedt, line 50: Might be worth it to bump this to 80.18.0.0, since that includes the regulator fix.

.github/workflows/hardware-smoke.yml
```diff
@@ -158,6 +163,11 @@ jobs:
...
```
cfriedt, line 169: Same here - 80.18.0.0 would probably be a bit more stable, but maybe I'm wrong.

danieldegrasse: It probably is, this PR was just created a while ago. Maybe instead of storing the FW bundles locally, we do something like the following:
- flash an old (but stable) copy of BMFW via the debug connection
- use `tt-flash` to flash the latest stable FW bundle from the tt-firmware repo

Does that seem reasonable?

cfriedt: Yes, I think that would make sense.

In the mean time, I'll approve. We can merge this if you prefer, and then you could work on pulling stable firmware from the repository in a separate PR, or you can update this one, depending on what is easier for you.

cfriedt (approved):

PR #83, opened 2025-04-04
Enable settings straps from device tree settings

TTDRosen:
This PR enables writing the the pca gpio expanders on the p300 to set bh strapping.

The device tree for the p300 was already setup to enable setting the straps so this now will set the straps based on those settings.


cfriedt (approved):
LGTM - CI is green

cfriedt (approved):
LGTM - CI is green

PR #84, opened 2025-04-04
scripts: tt_boot_fs: add support for output in hex format

danieldegrasse:
Add support for outputting the tt_boot_fs in hex format, for cases where the image will be flashed directly (IE via JTAG)

### DIFF
--- a/scripts/tt_boot_fs.py
+++ b/scripts/tt_boot_fs.py
@@ -728,11 +728,14 @@ def cksum(data: bytes):
     return calculated_checksum
 
 
-def mkfs(path: Path, env={"$ROOT": str(ROOT)}) -> bytes:
+def mkfs(path: Path, env={"$ROOT": str(ROOT)}, hex=False) -> bytes:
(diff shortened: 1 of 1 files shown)


scripts/tt_boot_fs.py
```diff
@@ -728,11 +728,14 @@ def cksum(data: bytes):
     return calculated_checksum
 
 
-def mkfs(path: Path, env={"$ROOT": str(ROOT)}) -> bytes:
+def mkfs(path: Path, env={"$ROOT": str(ROOT)}, hex=False) -> bytes:
```
cfriedt, line 731: Would be good to keep units for consistence. I think this might work? 🤷🏼‍♂️

```suggestion
def mkfs(path: Path, env={"$ROOT": str(ROOT)}, hex: Bool = False) -> bytes:
```

cfriedt (commented):
Looks good - minor nit only

cfriedt (approved):

PR #85, opened 2025-04-04
boards: tenstorrent: add definition for SPI DW driver in openocd on ARC

danieldegrasse:
Add definition for SPI DW driver within openocd on ARC. This driver will enable programming the SPI DW flash via the ARC JTAG interface, provided a patched version of openocd is used with the driver included

### DIFF
--- a/boards/tenstorrent/tt_blackhole/support/tt_blackhole_smc.cfg
+++ b/boards/tenstorrent/tt_blackhole/support/tt_blackhole_smc.cfg
@@ -108,6 +108,12 @@ arc-em.cpu4 configure -gdb-port disabled
 # execute at boot.
 arc-em.cpu1 configure -event reset-end "bh_reset_end"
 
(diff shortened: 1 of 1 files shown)


danieldegrasse:
This PR requires a build of openocd based on https://github.com/danieldegrasse/openocd/pull/1 in order to function.

I've attached a build that *should* run on Ubuntu 22.04- if you'd like to test this driver, run something like the following:

```
$ west attach -d build/smc --openocd <custom_openocd>
(gdb) mo flash read_bank 0 spi_flash.bin 0x0 382280 # Dump first 382280 bytes of flash to spi_flash.bin
(gdb) mo flash write_image erase spi_flash.bin # Program spi_flash.bin to flash, be aware this takes a long time and can't be cancelled
```

cfriedt:
We could merge this AND merge a similar stm32 one later.

Long term, I think stm32 would maybe be a slightly better failsafe.

PR #90, opened 2025-04-07
version: report bundle version in telemetry

afongTT:
Change the format of the bundle version to use: major.minor.patch.tweak from the VERSION file. Compile this version number into the fw_table and the manifest.json of the update.fwbundle. Also update manual-release.sh to name the file according to the new scheme.

Add an e2e test to check that the bundle version matches the VERSION file in the repo.

Output of `manual-release.sh`, `tt-flash`, and `tt-smi`:
```
[2254] afong:yyz-syseng-16 /localdev/afong/work/zephyrproject/tt-zephyr-platforms.git > ./scripts/manual-release.sh 
Created TEMP_DIR
Building release 80.16.111.111-rc1 / pack 80.16.111.111
Building tt_blackhole@p100/tt_blackhole/smc
Building tt_blackhole@p100a/tt_blackhole/smc
Building tt_blackhole@p150a/tt_blackhole/smc
Building tt_blackhole@p150b/tt_blackhole/smc
Building tt_blackhole@p150c/tt_blackhole/smc
Building tt_blackhole@p300a/tt_blackhole/smc
Building tt_blackhole@p300b/tt_blackhole/smc
Building tt_blackhole@p300c/tt_blackhole/smc
Creating fw_pack-80.16.111.111.fwbundle
Wrote fwbundle to fw_pack-80.16.111.111.fwbundle
Removed TEMP_DIR

[2254] afong:yyz-syseng-16 /localdev/afong/work/zephyrproject/tt-zephyr-platforms.git > tar -Oxf fw_pack-80.16.111.111.fwbundle ./manifest.json
{"version": "1.0.0", "bundle_version": {"fwId": 80, "releaseId": 16, "patch": 111, "debug": 111}}

[2264] afong:yyz-syseng-16 /localdev/afong/work/zephyrproject/tt-zephyr-platforms.git > tt-flash --force --fw-tar fw_pack-80.16.111.111.fwbundle 
Stage: SETUP
        Searching for default sys-config path
        Checking /etc/tenstorrent/config.json: not found
        Checking ~/.config/tenstorrent/config.json: not found

        Could not find config in default search locations, if you need it, either pass it in explicitly or generate one
        Warning: continuing without sys-config, galaxy systems will not be reset
Stage: DETECT
Stage: FLASH
        Sub Stage: VERIFY
                Verifying fw-package can be flashed: complete
                Verifying Blackhole[0] can be flashed
        Stage: FLASH
                Sub Stage FLASH Step 1: Blackhole[0]
                        Looks like you are running a very old set of fw, assuming that it needs an update
                        Now flashing tt-flash version: (80, 16, 111, 111)
                        Forced ROM update requested. ROM will now be updated.
                Sub Stage FLASH Step 2: Blackhole[0] {P100A-1}
                        Writing new firmware... SUCCESS
                        Firmware verification... SUCCESS
Stage: RESET
 Starting PCI link reset on BH devices at PCI indices: 0 
Waiting for up to 60 seconds for asic to come back after reset
 Config space reset completed for device 0 
 Finishing PCI link reset on BH devices at PCI indices: 0 
FLASH SUCCESS

[2266] afong:yyz-syseng-16 /localdev/afong/work/zephyrproject/tt-zephyr-platforms.git > tt-smi -s | grep fw_bundle_version
                "fw_bundle_version": "80.16.111.111",

```

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -6,6 +6,8 @@
 import logging
 import os
 import time
+import pathlib
+import re
 
 import pyluwen
 import pytest
@@ -13,6 +15,8 @@ import pytest
 from twister_harness import DeviceAdapter
 
 logger = logging.getLogger(__name__)
+
+SCRIPT_DIR = pathlib.Path(os.path.dirname(os.path.abspath(__file__)))
 
 # Constant memory addresses we can read from SMC
 ARC_STATUS = 0x80030060
@@ -134,3 +138,53 @@ def test_boot_status(arc_chip):
     status = arc_chip.axi_read32(BOOT_STATUS)
     assert (status >> 1) & 0x3 == 0x2, "SMC HW boot status is not valid"
     logger.info('SMC boot status "%d"', status)
+
+
+def get_int_version_from_file(filename) -> int:
+    with open(filename, "r") as f:
+        version_data = f.readlines()
+    version_dict = {}
+    for line in version_data:
+        if line:
+            # Split the line into key-value pairs
+            key_value = line.split("=")
+            key = key_value[0].strip()
+
+            if len(key_value) == 2:
+                key = key_value[0].strip()
+                try:
+                    value = int(key_value[1].strip(), 0)
+                except ValueError:
+                    # Some values are strings
+                    value = key_value[1].strip()
+                version_dict[key] = value
+            else:
+                version_dict[key] = None
+
+    if version_dict["EXTRAVERSION"]:
+        version_rc = int(re.sub(r"[^\d]", "", version_dict["EXTRAVERSION"]))
+    else:
+        # version_dict["EXTRAVERSION"] is None or an empty string
+        version_rc = 0
+
+    version_int = (
+        version_dict["VERSION_MAJOR"] << 24
+        | version_dict["VERSION_MINOR"] << 16
+        | version_dict["PATCHLEVEL"] << 8
+        | version_rc
+    )
+    return version_int
+
+
+@pytest.mark.flash
+def test_fw_bundle_version(arc_chip):
+    """
+    Checks that the fw bundle version in telemetry matches the repo
+    """
+    telemetry = arc_chip.get_telemetry()
+
+    exp_bundle_version = get_int_version_from_file(SCRIPT_DIR.parents[2] / "VERSION")
+    assert (
+        telemetry.fw_bundle_version == exp_bundle_version
+    ), f"Firmware bundle version mismatch: {telemetry.fw_bundle_version:#010x} != {exp_bundle_version:#010x}"
+    logger.info(f"FW bundle version: {telemetry.fw_bundle_version:#010x}")
--- a/lib/tenstorrent/bh_arc/CMakeLists.txt
+++ b/lib/tenstorrent/bh_arc/CMakeLists.txt
@@ -90,23 +90,28 @@ if(NOT DEFINED CONFIG_TT_SMC_RECOVERY)
 
 	message(STATUS "Generating rom configuration binary files for ${PROD_NAME}")
 
+	include(${ZEPHYR_CURRENT_MODULE_DIR}/cmake/bundle_version.cmake)
+
 	# variable is needed to avoid protobuf version clashing
 	if (PROD_NAME MATCHES "^P300")
 	set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
 		COMMAND PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
 		  ${PYTHON_EXECUTABLE} ${ZEPHYR_CURRENT_MODULE_DIR}/scripts/encode_spirom_bins.py
 		  --board ${PROD_NAME}_L --build-dir ${CMAKE_BINARY_DIR} --output ${CMAKE_BINARY_DIR}/generated_board_cfg
+		  --bundle-version ${BUNDLE_VERSION_STRING}
 	)
 	set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
 		COMMAND PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
 		  ${PYTHON_EXECUTABLE} ${ZEPHYR_CURRENT_MODULE_DIR}/scripts/encode_spirom_bins.py
 		  --board ${PROD_NAME}_R --build-dir ${CMAKE_BINARY_DIR} --output ${CMAKE_BINARY_DIR}/generated_board_cfg
+		  --bundle-version ${BUNDLE_VERSION_STRING}
 	)
 	else()
 	set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
 		COMMAND PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
 		  ${PYTHON_EXECUTABLE} ${ZEPHYR_CURRENT_MODULE_DIR}/scripts/encode_spirom_bins.py
 		  --board ${PROD_NAME} --build-dir ${CMAKE_BINARY_DIR} --output ${CMAKE_BINARY_DIR}/generated_board_cfg
+		  --bundle-version ${BUNDLE_VERSION_STRING}
 	)
 	endif()
 endif()
--- a/scripts/encode_spirom_bins.py
+++ b/scripts/encode_spirom_bins.py
@@ -86,6 +86,16 @@ def convert_proto_txt_to_bin_file(
     print()
 
 
+def decode_bundle_version(bundle_version: str) -> int:
+    try:
+        major, minor, patch, rc = map(int, bundle_version.split("."))
+        return (major << 24) | (minor << 16) | (patch << 8) | rc
+    except ValueError:
+        raise ValueError(
+            "Invalid bundle version format. Expected format: major.minor.patch.rc"
+        )
+
+
 def main():
     parser = argparse.ArgumentParser(
         description="Encode SPIROM bins", allow_abbrev=False
@@ -111,6 +121,13 @@ def main():
         default="build",
         help="Name of build folder with generated protobuf files",
     )
+    parser.add_argument(
+        "--bundle-version",
+        type=str,
+        default="0.0.0.0",
+        required=True,
+        help="Firmware bundle version to encode",
+    )
     args = parser.parse_args()
     build_folder = os.path.join(args.build_dir, "zephyr/python_proto_files")
     if not os.path.exists(build_folder):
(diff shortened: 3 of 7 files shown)


scripts/manual-release.sh
```diff
@@ -40,8 +41,8 @@ else
   EXTRAVERSION_NUMBER=0
 fi
 
-RELEASE="$MAJOR.$MINOR.$PATCH$EXTRAVERSION"
-PRELEASE="$MAJOR.$MINOR.$PATCH.$EXTRAVERSION_NUMBER"
+RELEASE="$MAJOR.$MINOR.$PATCH.$TWEAK$EXTRAVERSION"
```
TTDRosen, line 44: Should there be a `-` separator here?

afongTT: `-` is added in line 38

TTDRosen: AH...

scripts/manual-release.sh
```diff
@@ -40,13 +41,13 @@ else
...
+PRELEASE="$MAJOR.$MINOR.$PATCH.$TWEAK"
```
cfriedt, line 45: I don't think we can use the tweak version, unfortunately, because the least-significant byte of the 32-bit version number is used by the release candidate number (which is 0 for final releases)

scripts/manual-release.sh
```diff
@@ -26,6 +26,7 @@ BOARD_REVS="p100 p100a p150a p150b p150c p300a p300b p300c"
 MAJOR="$(grep "^VERSION_MAJOR" VERSION | awk '{print $3}')"
 MINOR="$(grep "^VERSION_MINOR" VERSION | awk '{print $3}')"
 PATCH="$(grep "^PATCHLEVEL" VERSION | awk '{print $3}')"
+TWEAK="$(grep "^VERSION_TWEAK" VERSION | awk '{print $3}')"
```
cfriedt, line 29: I don't think we can use the tweak version, unfortunately, because the least-significant byte of the 32-bit version number is used by the release candidate number (which is 0 for final releases)

scripts/manual-release.sh
cfriedt, line 45: This needs to be reverted, as it would break the release process which relies on -rc1, -rc2. 

I don't know what we have used or would use the TWEAK version for, but we should probably just use the PATCH version instead.

cfriedt (changes requested):
I don't think we can use the tweak version, unfortunately, because the least-significant byte of the 32-bit version number is used by the release candidate number (which is 0 for final releases)

danieldegrasse (commented):
I like the idea here- @cfriedt what would be the risk of setting the RC version to `99` during development, or something similar? I think we need some way to tell "what I just flashed onto the card is now actually running"

Ideally I'd rather report something truly unique (like the git sha of `tt-zephyr-platforms`) from the firmware, but this would require extending our telemetry tooling

scripts/manual-release.sh
afongTT: I was following the Zephyr convention for how it generates the version defines. Are we ok with doing it differently for our fwbundle?

e.g. if I set this in `app/smc/VERSION`:
```
VERSION_MAJOR = 0
VERSION_MINOR = 9
PATCHLEVEL = 1
VERSION_TWEAK = 111
EXTRAVERSION =
```

Then it generates `APPVERSION = (major << 24) | (minor << 16) | (patch << 8) | tweak` in `build/smc/zephyr/include/generated/zephyr/app_version.h`:

```c
#ifndef _APP_VERSION_H_
#define _APP_VERSION_H_

/* The template values come from cmake/version.cmake
 * BUILD_VERSION related template values will be 'git describe',
 * alternatively user defined BUILD_VERSION.
 */

/* #undef ZEPHYR_VERSION_CODE */
/* #undef ZEPHYR_VERSION */

#define APPVERSION                   0x9016f
#define APP_VERSION_NUMBER           0x901
#define APP_VERSION_MAJOR            0
#define APP_VERSION_MINOR            9
#define APP_PATCHLEVEL               1
#define APP_TWEAK                    111
#define APP_VERSION_STRING           "0.9.1"
#define APP_VERSION_EXTENDED_STRING  "0.9.1+111"
#define APP_VERSION_TWEAK_STRING     "0.9.1+111"

#define APP_BUILD_VERSION 755c7a9c3077


#endif /* _APP_VERSION_H_ */
```

afongTT:
> I don't think we can use the tweak version, unfortunately, because the least-significant byte of the 32-bit version number is used by the release candidate number (which is 0 for final releases)

I was following the Zephyr convention for how it generates the version defines. Are we ok with doing it differently for our fwbundle?

e.g. if I set this in `app/smc/VERSION`:
```
VERSION_MAJOR = 0
VERSION_MINOR = 9
PATCHLEVEL = 1
VERSION_TWEAK = 111
EXTRAVERSION =
```

Then it generates `APPVERSION = (major << 24) | (minor << 16) | (patch << 8) | tweak` in `build/smc/zephyr/include/generated/zephyr/app_version.h`:

```c
#ifndef _APP_VERSION_H_
#define _APP_VERSION_H_

/* The template values come from cmake/version.cmake
 * BUILD_VERSION related template values will be 'git describe',
 * alternatively user defined BUILD_VERSION.
 */

/* #undef ZEPHYR_VERSION_CODE */
/* #undef ZEPHYR_VERSION */

#define APPVERSION                   0x9016f
#define APP_VERSION_NUMBER           0x901
#define APP_VERSION_MAJOR            0
#define APP_VERSION_MINOR            9
#define APP_PATCHLEVEL               1
#define APP_TWEAK                    111
#define APP_VERSION_STRING           "0.9.1"
#define APP_VERSION_EXTENDED_STRING  "0.9.1+111"
#define APP_VERSION_TWEAK_STRING     "0.9.1+111"

#define APP_BUILD_VERSION 755c7a9c3077


#endif /* _APP_VERSION_H_ */
```

lib/tenstorrent/bh_arc/CMakeLists.txt
```diff
@@ -90,23 +90,36 @@ if(NOT DEFINED CONFIG_TT_SMC_RECOVERY)
...
+	# Get the bundle version
+	set(VERSION_FILE ${ZEPHYR_CURRENT_MODULE_DIR}/VERSION)
+	set(VERSION_TYPE BUNDLE)
+	include(${ZEPHYR_BASE}/cmake/modules/version.cmake)
+	string(REGEX REPLACE "[^0-9]+" "" BUNDLE_VERSION_EXTRA_NUM "${BUNDLE_VERSION_EXTRA}")
+	set(BUNDLE_VERSION_STRING "${BUNDLE_VERSION_MAJOR}.${BUNDLE_VERSION_MINOR}.${BUNDLE_PATCHLEVEL}.${BUNDLE_VERSION_EXTRA_NUM}")
+	if("${BUNDLE_VERSION_STRING}" STREQUAL "...")
+		message(FATAL_ERROR "Unable to extract bundle version")
+	endif()
+
```
cfriedt, line 102: Can you move this out to `cmake/bundle-version.cmake`? It should be reusable.

https://github.com/tenstorrent/tt-zephyr-platforms/pull/79/commits/9500ee193c3351e9202546f570eb57c4bd5a70e6

With the change above, it just needs to be included via cmake anywhere the bundle version is needed.

lib/tenstorrent/bh_arc/CMakeLists.txt
```diff
@@ -90,23 +90,36 @@ if(NOT DEFINED CONFIG_TT_SMC_RECOVERY)
...
 	# variable is needed to avoid protobuf version clashing
 	if (PROD_NAME MATCHES "^P300")
 	set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
 		COMMAND PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
 		  ${PYTHON_EXECUTABLE} ${ZEPHYR_CURRENT_MODULE_DIR}/scripts/encode_spirom_bins.py
 		  --board ${PROD_NAME}_L --build-dir ${CMAKE_BINARY_DIR} --output ${CMAKE_BINARY_DIR}/generated_board_cfg
+		  --bundle-version ${BUNDLE_VERSION_STRING}
```
cfriedt, line 109: I opted to pass in the bundle version as an integer, in case you like that as an option.

https://github.com/tenstorrent/tt-zephyr-platforms/pull/79/commits/0bbb41b1c5ca625f4ceff78f420fff443d91a013

scripts/encode_spirom_bins.py
```diff
@@ -86,6 +86,16 @@ def convert_proto_txt_to_bin_file(
...
+def decode_bundle_version(bundle_version: str) -> int:
+    try:
+        major, minor, patch, rc = map(int, bundle_version.split("."))
+        return (major << 24) | (minor << 16) | (patch << 8) | rc
+    except ValueError:
+        raise ValueError(
+            "Invalid bundle version format. Expected format: major.minor.patch.rc"
+        )
+
+
```
cfriedt, line 98: Could drop this part if you're ok passing in the bundle version as an int

lib/tenstorrent/bh_arc/CMakeLists.txt
afongTT: Cherry-picked commit 9500ee1 from your other PR

lib/tenstorrent/bh_arc/CMakeLists.txt
afongTT: Hmm, the BUNDLE_VERSION_NUMBER seems to encode it differently than how I have it here. It seems to only keep `(major << 16) | (minor << 8) | patch` and doesn't keep anything from EXTRAVERSION.

Rather than figure out how to do that in a cmake rule, I opted to parse it in the Python script

app/smc/pytest/e2e_smoke.py
```diff
@@ -134,3 +138,55 @@ def test_boot_status(arc_chip):
...
+
+@pytest.mark.flash
+def test_fw_bundle_version(arc_chip):
+    """
+    Checks that the fw bundle version in telemetry matches the repo
+    """
+    telemetry = arc_chip.get_telemetry()
+
+    exp_bundle_version = get_int_version_from_file(
+        SCRIPT_DIR / ".." / ".." / ".." / "VERSION"
```
cfriedt, line 187:
```suggestion
        SCRIPT_DIR.parent[2] / "VERSION"
```

cfriedt (approved):
Just a nit, but not blocking

app/smc/pytest/e2e_smoke.py
afongTT: Oh that's neat, I didn't know about this. Updated

app/smc/pytest/e2e_smoke.py
```diff
@@ -14,6 +16,8 @@
 
 logger = logging.getLogger(__name__)
 
+SCRIPT_DIR = pathlib.Path(os.path.dirname(os.path.abspath(__file__)))
```
cfriedt, line 19: I think this would work too. Not blocking again.
```suggestion
SCRIPT_DIR = pathlib.Path(__file__).parent()
```

lib/tenstorrent/bh_arc/CMakeLists.txt
cfriedt: Weird - it definitely works in the commit that I linked above and gets printed with each build on that branch.

PR #91, opened 2025-04-07
fw_table: enable fan control for P100A and P150A

palexsonTT:
Configure fan_ctrl_en flag in the fw_table to enable/disable fan curve for each board type. Enable for P100A and P150A, but keep disabled for all other boards.


danieldegrasse (approved):
Tried this on a P150A and the ASIC temp was stable at ~56C, with the fan ramped down significantly

PR #93, opened 2025-04-07
tenstorrent: bh_arc: add gddr throttler

palexsonTT:
Add current maximum GDDR temperature to the telemetry table. Add a throttler based on GDDR temperature. Add GDDR thermal throttler limit to FW table.

### DIFF
--- a/boards/tenstorrent/tt_blackhole/spirom_data_tables/GALAXY/fw_table.txt
+++ b/boards/tenstorrent/tt_blackhole/spirom_data_tables/GALAXY/fw_table.txt
@@ -13,6 +13,7 @@ chip_limits {
   therm_trip_l1_limit: 0
   bus_peak_limit: 0
   frequency_margin: 0
+  gddr_thm_limit: 85
 }
 
 feature_enable {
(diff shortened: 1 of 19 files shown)


boards/tenstorrent/tt_blackhole/spirom_data_tables/GALAXY/fw_table.txt
```diff
@@ -13,6 +13,7 @@ chip_limits {
   therm_trip_l1_limit: 0
   bus_peak_limit: 0
   frequency_margin: 0
+  gddr_thm_limit: 0
```
abuckTT, line 16: Is there a reason Galaxy sets to 0? If I understand the code correctly, that means it will actually clamp to minimum of 50 which seems too low.

palexsonTT: Nope, I just left it blank because the other galaxy throttler limits are left blank I thought they might be filled in later...just changed it to match all the others.

abuckTT (approved):

PR #97, opened 2025-04-08
ci: allow forks to run their own CI

cfriedt:
If a customer buys a Tenstorrent card, they should be able to set up their own CI infrastructure.

Approval for workflow runs is already restricted via settings.

```
Settings -> Actions -> General ->
Approval for running fork pull requests... from contributors
-> Require approval for all external contributors
```

Secrets are per-organization / repo

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -7,7 +7,6 @@ on:
 
 jobs:
   hardware-metal-test:
-    if: github.repository_owner == 'tenstorrent'
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


cfriedt:
> This should be fine given that secrets are per organization. Have you tested this within a fork to see what happens with these CI jobs when no runners are available to execute them?

I've tested it on my fork, and I have 1 runner available for p100, but the p100a test just queues for ever.

danieldegrasse (approved):

danieldegrasse:
Not sure why CI continues to fail here- maybe rebase? I opened https://github.com/tenstorrent/tt-zephyr-platforms/pull/119 to try to check how github creates the branch to run tests from- github docs suggest the PR is merged with main, but CI failures here suggest otherwise

PR #98, opened 2025-04-08
aiclk_ppm: add aiclk sweeping functionality

yangxuTT:
Add a handler for the aisweep_start and aisweep_stop messages to toggle a sweep_en flag. If sweep enabled, calculation for targ aiclk is randomized between set min and max freq.

`sweep_aiclk.py` script available in `syseng` under `yxu/aiclk_sweep` branch.

FYI @tcolacoTT 

https://github.com/user-attachments/assets/03c2c863-76f5-4111-ad6a-1772dd214085


lib/tenstorrent/bh_arc/aiclk_ppm.c
```diff
@@ -64,6 +64,12 @@ void CalculateTargAiclk(void)
 	/* (it will not be above Fmax, since we calculated the max limits last) */
 	aiclk_ppm.targ_freq = MAX(targ_freq, aiclk_ppm.fmin);
 
+	/* Apply random frequency if sweep is enabled */
+	if (aiclk_ppm.sweep_en == 1) {
+		aiclk_ppm.targ_freq = rand() % (aiclk_ppm.sweep_high - aiclk_ppm.sweep_low + 1)
+							  + aiclk_ppm.sweep_low;
+	}
```
cfriedt, line 71: This is generally not great code to have enabled in production. Consider the case where a stack overflow occurs and some rogue memory write flips `sweep_en` to 1. Bad things could happen 😬

cfriedt (changes requested):
I'd really like to see some tests added here for CI.

* a new [ZTest](https://docs.zephyrproject.org/latest/develop/test/ztest.html) under https://github.com/tenstorrent/tt-zephyr-platforms/tree/main/tests/lib/tenstorrent/bh_arc/src
* before each test in the suite, in the `before` function, set aiclk to it's lowest value
* after each test in the suite has executed, in the `after` function, set aiclk back to it's lowest value
* a new library function `int SetAiClkPpm(uint32_t freq)` - it should return `-ERANGE` if an invalid frequency is used and 0 on success.
* another new library function `int GetAiClkPpm(uint32_t *freq)` - it should return `-EINVAL` if `freq` is NULL, and 0 on success.
* one `ZTEST()`
  * tries a few bad target frequencies (0, `UINT32_MAX`, `aiclk_ppm.fmin - 1`, `aiclk_ppm.fmax + 1`) and asserts that `SetAiClkPpm()` fails and `GetAiClkPpm()` succeeds and returns the original target frequency, unchanged
  * iterates over a series of good target frequencies
  * asserts that `SetAiClkPpm()` succeeds and that `GetAiClkPpm()` succeeds and returns the expected target frequency.
  
If there are only specific discrete values that can be used, and the clock driver rounds up or down by some amount, that should be tested as well. Again, just comparing between actual and expected values.

cfriedt:
@yangxuTT - please let me know if you need any help making these changes

yangxuTT:
Hi @cfriedt, yes! I want to clarify a few things and I'll get started on making the requested changes.

For the `sweep_en` flag, is there a safer way for me to implement this to avoid the issues you mentioned? Should I add a hardcoded limit/check after the randomization to ensure the frequency is always in a safe range, regardless of potential memory issues?

Also, for the ZTEST, could you clarify the difference between SetAiclkPpm/GetAiclkPpm library functions and the existing SetAICLK/GetAICLK functions?

Thank you!

cfriedt:
> For the `sweep_en` flag, is there a safer way for me to implement this to avoid the issues you menioned?

Yes. I don't think we should have a `SweepAiclkHandler()` function present in production firmware. It sounds like it's mainly there for testing purposes. So, at the very minimum, it should be surrounded by an `#ifdef ... #endif`. E.g. `#ifdef CONFIG_ZTEST`. Similarly, the fields `sweep_en`, `sweep_low`, `sweep_high` fields in `AiclkPPM` should be `#ifdef`'ed as well.

> Should I add a hardcoded limit/check after the randomization to ensure the frequency is always in a safe range, regardless of potential memory issues?

Why is randomization required. 

> Also, for the ZTEST, could you clarify the difference between SetAiclkPpm/GetAiclkPpm library functions and the existing SetAICLK/GetAICLK functions?

SetAiclkPpm and GetAiclkPpm would be used to set & get the `aiclk_ppm.targ_freq` value.

Additionally, most firmware functions should return an `int`. 0 means success and a negative errno value (from `<errno.h>`) means failure.

cfriedt:
Finished in #398

PR #99, opened 2025-04-08
ci: enable building metal on a per-PR basis with caching

danieldegrasse:
Enable building metal on a per-PR basis using build artifact caching

### DIFF
--- a/.github/workflows/hardware-smoke.yml
+++ b/.github/workflows/hardware-smoke.yml
@@ -254,3 +254,212 @@ jobs:
           python3 ./scripts/bmc_rtt.py -n --openocd /opt/tenstorrent/bin/openocd-rtt
           echo "SMC RTT logs:"
           python3 ./scripts/smc_rtt.py -n --openocd /opt/tenstorrent/bin/openocd-rtt
+
+  build-metal:
+    runs-on: self-hosted
+    container:
+      image: ghcr.io/zephyrproject-rtos/ci-repo-cache:v0.27.4.20241026
+      volumes:
+        - /dev/hugepages-1G:/dev/hugepages-1G
+        - /opt/tenstorrent/:/opt/tenstorrent/
+      options: '--device /dev/tenstorrent --device /dev/bus/usb --privileged'
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          path: tt-zephyr-platforms
+
+      - name: Check Metal Build Cache
+        id: metal-cache-restore
+        uses: actions/cache/restore@v4
+        with:
+          lookup-only: true
+          path: |
+            ci/tt-metal/build_Release/programming_examples
+            ci/tt-metal/build_Release/test
+            ci/tt-metal/build_Release/lib
+            ci/tt-metal/python_env
+            ci/tt-metal/runtime
+          key: metal-build-cache-${{ hashFiles('tt-zephyr-platforms/submanifests/ci.yaml') }}
+
+      - uses: ./tt-zephyr-platforms/.github/workflows/prepare-zephyr
+        if: ${{ steps.metal-cache-restore.outputs.cache-hit != 'true' }}
+        with:
+          app-path: tt-zephyr-platforms
+
+      - name: Fix APT URLs
+        if: ${{ steps.metal-cache-restore.outputs.cache-hit != 'true' }}
+        run: |
+          # Patch APT URLs to use https
+          sed -i -e 's/http/https/g' /etc/apt/sources.list
+          apt update
+
+      - name: Install clang 17
+        if: ${{ steps.metal-cache-restore.outputs.cache-hit != 'true' }}
+        run: |
+          wget https://apt.llvm.org/llvm.sh
+          sed -i "s/http:\/\/apt.llvm.org/https:\/\/apt.llvm.org/" -i llvm.sh
+          chmod +x llvm.sh
+          ./llvm.sh 17 all
+          rm -f llvm.sh
+
+      - name: Clone Metal and Install Dependencies
+        if: ${{ steps.metal-cache-restore.outputs.cache-hit != 'true' }}
+        shell: bash
+        run: |
+          west config manifest.group-filter -- +ci
+          west update
+          cd ci/tt-metal
+          sed -i "s/http:\/\/apt.llvm.org/https:\/\/apt.llvm.org/" -i install_dependencies.sh
+          ./install_dependencies.sh --mode build --docker
+
+      - name: Build Metal Tests
+        if: ${{ steps.metal-cache-restore.outputs.cache-hit != 'true' }}
+        shell: bash
+        run: |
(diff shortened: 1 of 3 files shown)


danieldegrasse:
Blocked by https://github.com/tenstorrent/tt-zephyr-platforms/issues/118, I believe the metal build is working here though :)

danieldegrasse:
Currently the metal test takes ~7 minutes to run if metal isn't rebuilt. We can cut that down if we reuse the combined firmware bundle built with other jobs (probably by about 2 minutes)

danieldegrasse:
> Currently the metal test takes ~7 minutes to run if metal isn't rebuilt. We can cut that down if we reuse the combined firmware bundle built with other jobs (probably by about 2 minutes)
> 
> If we do have to rebuild metal, the test takes ~45 minutes to run. This will only occur each time the CI submanifest is updated though (so even if a PR updating metal is merged, we may only need one metal build in total to update the build cache)

Unfortunately github doesn't have a great way to reuse uploaded build artifacts in future jobs. What I think we should do in a future PR is the following:
- make the build firmware step a reusable action that caches the output of the build
- In future uses of the reusable action, skip the build if the firmware bundle already exists

.github/workflows/hardware-smoke.yml
```diff
@@ -236,3 +236,207 @@ jobs:
...
+    runs-on: ${{ matrix.config.runs-on }}
+    env:
+      "ZEPHYR_SDK_INSTALL_DIR": /opt/toolchains
+    container:
+      image: ghcr.io/zephyrproject-rtos/ci-repo-cache:v0.27.4.20241026
+      volumes:
+        - /dev/hugepages-1G:/dev/hugepages-1G
+        - /opt/tenstorrent/fw/stable:/opt/tenstorrent/fw/stable
+        - /opt/tenstorrent/twister:/opt/tenstorrent/twister
+        - /opt/tenstorrent/bin/openocd-rtt:/opt/tenstorrent/bin/openocd-rtt
```
cfriedt, line 341:
```suggestion
        - /opt/tenstorrent:/opt/tenstorrent
```

.github/workflows/hardware-smoke.yml
```diff
@@ -236,3 +236,207 @@ jobs:
...
+          west flash -r tt_flash --force
+
+      - name: Prepare Metal Tests
+        working-directory: ci/tt-metal
+        shell: bash
+        run: |
+          # Place _ttnn.so in the correct location
+          cp build_Release/lib/_ttnn.so ttnn/ttnn/_ttnn.so
+          # Symlink build dir to build_Release
+          ln -s build_Release build
```
cfriedt, line 427: What about using or appending to `LD_LIBRARY_PATH` in `GITHUB_ENV`?

.github/workflows/hardware-smoke.yml
```diff
@@ -236,3 +236,207 @@ jobs:
...
+        run: |
+          west config manifest.group-filter -- +ci
+          west update
+          cd ci/tt-metal
+          sed -i "s/http:\/\/apt.llvm.org/https:\/\/apt.llvm.org/" -i install_dependencies.sh
+          ./install_dependencies.sh --mode build --docker
+          echo "ARCH_NAME=blackhole" >> "$GITHUB_ENV"
+          echo "TT_METAL_HOME=$PWD" >> "$GITHUB_ENV"
+          echo "PYTHONPATH=$PWD" >> "$GITHUB_ENV"
+          echo "LD_LIBRARY_PATH=$PWD/build_Release/lib" >> "$GITHUB_ENV"
```
cfriedt, line 377: This could be used instead of copying further down.

cfriedt (commented):
Just some minor suggestions. Looks good 👍

.github/workflows/hardware-smoke.yml
danieldegrasse: I'm not sure why, but I needed to symlink the build dir as well. I suspect that some of the metal test scripts expect to find the `build` folder

.github/workflows/hardware-smoke.yml
danieldegrasse: Are you talking about the `_ttnn.so` file? I think that the python imports expect to find it there- not 100% sure though

.github/workflows/hardware-smoke.yml
danieldegrasse: Yeah, I think that would be awesome! I'd prefer it to be built on each PR though- that way we can pull the container from PRs when there is a change (like harvesting) that needs to be tested with a new firmware version prior to both merging

.github/workflows/hardware-smoke.yml
cfriedt: If python has hard coded paths, then I guess a copy is necessary.

cfriedt (dismissed):

danieldegrasse:
> @danieldegrasse - ok to merge from my perspective. I guess future container improvements can be made later?

I think @dimitri-tenstorrent wanted to sync about this, we should have a resolution on this by EOD

.github/workflows/hardware-smoke.yml
```diff
@@ -254,3 +254,212 @@ jobs:
...
+        if: ${{ steps.metal-cache-restore.outputs.cache-hit != 'true' }}
+        uses: actions/cache/save@v4
+        with:
+          path: |
+            ci/tt-metal/build_Release/programming_examples
+            ci/tt-metal/build_Release/test
+            ci/tt-metal/build_Release/lib
+            ci/tt-metal/python_env
+            ci/tt-metal/runtime
+          key: metal-build-cache-${{ hashFiles('tt-zephyr-platforms/submanifests/ci.yaml') }}
```
cfriedt, line 334: I think this could easily be converted into a reusable workflow, like `build-fw.yml`

.github/workflows/hardware-smoke.yml
```diff
@@ -254,3 +254,212 @@ jobs:
...
+    env:
+      "ZEPHYR_SDK_INSTALL_DIR": /opt/toolchains
+    container:
+      image: ghcr.io/zephyrproject-rtos/ci-repo-cache:v0.27.4.20241026
+      volumes:
+        - /dev/hugepages-1G:/dev/hugepages-1G
+        - /opt/tenstorrent/:/opt/tenstorrent/
+      options: '--device /dev/tenstorrent --device /dev/bus/usb --privileged'
+    steps:
+      - uses: actions/checkout@v4
```
cfriedt, line 358: could add a `- uses: build-metal.yml` or something along those lines if the above were a reusable workflow.

danieldegrasse:
Given the discussion today with @dimitri-tenstorrent I think the plan is to not move forwards with this PR, and instead use the metal container with prebuilt tests- I will keep this PR open until we get that enabled in case plans change, but close it out at that point

cfriedt:
Closing

PR #106, opened 2025-04-09
app: bmc + smc: print the tenstorrent git revision along with the boot banner

cfriedt:
Print the git revision in the format

```
*** TT_GIT_VERSION b7d6567 ***
```

when `CONFIG_TT_BOOT_BANNER_GIT_VERSION` is enabled.

### DIFF
--- a/lib/tenstorrent/banner/tt_banner.c
+++ b/lib/tenstorrent/banner/tt_banner.c
@@ -8,6 +8,11 @@
 #include <zephyr/kernel.h>
 #include <version.h>
 
+#ifndef CONFIG_TT_BOOT_BANNER_GIT_VERSION
+/* prevent compiler warning */
+#define TT_GIT_VERSION ""
+#endif
+
 static int tt_boot_banner(void)
 {
 	/* clang-format off */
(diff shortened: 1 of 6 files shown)


lib/tenstorrent/banner/tt_banner.c
```diff
@@ -44,6 +49,10 @@ static int tt_boot_banner(void)
 	printk(logo);
 	printk("*** Booting " CONFIG_BOARD " with Zephyr OS " STRINGIFY(BUILD_VERSION) " ***\n");
 
+	if (IS_ENABLED(CONFIG_TT_BOOT_BANNER_GIT_VERSION)) {
+		printk("*** TT_GIT_VERSION " TT_GIT_VERSION " ***\n");
+	}
+
```
afongTT, line 55: Why do we runtime check these rather than #ifdef compile it out?

afongTT (approved):

lib/tenstorrent/banner/tt_banner.c
cfriedt: `IS_ENABLED()` lets the compiler, static analysis tools, etc check code where as `#ifdef` does not.

PR #111, opened 2025-04-09
fw_bundle: bump version to rc2

danieldegrasse:
Bump the firmware bundle to rc2, with harvesting and noc translation disabled


danieldegrasse:
Closing this out, as we will create a 80.17.0.0 release with the VM fix and I2C hang fix cherry-picked, and use that to unblock metal VM issues

PR #113, opened 2025-04-10
Send board input power to chip instead of board input current

palexsonTT:
Enable input power reading in BMC FW using the INA228. Send input power to chip instead of input current. Add input power to the telemetry table.

Since SMC FW only uses current to calculate board input power for throttling, we can just send power directly instead. If needed, we could continue to send current. Nothing else uses it, but there is a telemetry field for it.

### DIFF
--- a/lib/tenstorrent/bh_arc/cm2bm_msg.c
+++ b/lib/tenstorrent/bh_arc/cm2bm_msg.c
@@ -29,6 +29,7 @@ typedef struct {
 static Cm2BmMsgState cm2bm_msg_state;
 static bool bmfw_ping_valid;
 static int32_t current;
+static uint32_t power;
 K_MSGQ_DEFINE(cm2bm_msg_q, sizeof(Cm2BmMsg), 4, _Alignof(Cm2BmMsg));
 
 int32_t EnqueueCm2BmMsg(const Cm2BmMsg *msg)
@@ -233,10 +234,27 @@ int32_t Bm2CmSendCurrentHandler(const uint8_t *data, uint8_t size)
 	return 0;
 }
 
+int32_t Bm2CmSendPwrHandler(const uint8_t *data, uint8_t size)
+{
+	if (size != 4) {
+		return -1;
+	}
+
+	power = *(uint32_t *)data;
+
+	return 0;
+}
+
+
 /* TODO: Put these somewhere else? */
 int32_t GetInputCurrent(void)
 {
 	return current;
+}
+
+uint32_t GetInputPower(void)
+{
+	return power;
 }
 
 int32_t Bm2CmSendFanRPMHandler(const uint8_t *data, uint8_t size)
--- a/lib/tenstorrent/bh_arc/cm2bm_msg.h
+++ b/lib/tenstorrent/bh_arc/cm2bm_msg.h
@@ -51,7 +51,9 @@ typedef struct bmStaticInfo {
 int32_t Bm2CmSendDataHandler(const uint8_t *data, uint8_t size);
 int32_t Bm2CmPingHandler(const uint8_t *data, uint8_t size);
 int32_t Bm2CmSendCurrentHandler(const uint8_t *data, uint8_t size);
+int32_t Bm2CmSendPwrHandler(const uint8_t *data, uint8_t size);
 int32_t GetInputCurrent(void);
+uint32_t GetInputPower(void);
 int32_t Bm2CmSendFanRPMHandler(const uint8_t *data, uint8_t size);
 
 #endif
--- a/lib/tenstorrent/bh_arc/smbus_target.c
+++ b/lib/tenstorrent/bh_arc/smbus_target.c
@@ -198,6 +198,13 @@ static SmbusConfig smbus_config = {
 			  .handler = {
 					  .rcv_handler = &Bm2CmSetBoardPwrLimit
 			  }},
+		[0x25] = {
+			.valid = 1,
+			.trans_type = kSmbusTransBlockWrite,
+			.expected_blocksize = 4,
+			.handler = {
+					.rcv_handler = &Bm2CmSendPwrHandler
+			}},
 #endif
 		[0xD8] = {
 
--- a/lib/tenstorrent/bh_arc/telemetry.c
+++ b/lib/tenstorrent/bh_arc/telemetry.c
@@ -110,7 +110,7 @@ static void UpdateGddrTelemetry(void)
 			 * [15] - Error GDDR 7
 			 */
 			telemetry[GDDR_STATUS] |= (gddr_telemetry.training_complete << (i * 2)) |
-				(gddr_telemetry.gddr_error << (i * 2 + 1));
+						  (gddr_telemetry.gddr_error << (i * 2 + 1));
 
 			/* DDR_x_y_TEMP:
 			 * [31:24] GDDR y top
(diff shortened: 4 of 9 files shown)


cfriedt:
@palexsonTT - is this fixing a bug or is it otherwise critical for the next release? If so, please identify the jira issue. If not, we'll hold off  for a bit.

palexsonTT:
@cfriedt Nope, not needed necessarily, just a cleanup! It can wait.

cfriedt (approved):
Just a couple of suggestions, but given that there is an expected block size, I don't see any reason not to approve anyway.

cfriedt (dismissed):
@palexsonTT - just saw that there were merge conflicts. Probably just needs a rebase.

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -264,7 +264,9 @@ static void update_telemetry(void)
 	UpdateGddrTelemetry();
 	telemetry[MAX_GDDR_TEMP] = GetMaxGDDRTemp();
 	telemetry[INPUT_CURRENT] =
-		GetInputCurrent();    /* Input current - reported in A in signed int 16.16 format */
+		GetInputCurrent(); /* Input current - reported in A in signed int 16.16 format */
+	telemetry[INPUT_POWER] =
+		GetInputPower();      /* Input power - reported in W in unsigned int 16.16 format */
```
afongTT, line 269: Let's hide `INPUT_CURRENT` telemetry entry and only report `INPUT_POWER`, since input current stays at 0 after this change.

lib/tenstorrent/bh_arc/cm2bm_msg.c
```diff
@@ -233,12 +234,29 @@ int32_t Bm2CmSendCurrentHandler(const uint8_t *data, uint8_t size)
 	return 0;
 }
 
+int32_t Bm2CmSendPwrHandler(const uint8_t *data, uint8_t size)
+{
+	if (size != 4) {
```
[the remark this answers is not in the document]
afongTT: I think leave this check for `size != 4` as extra data validation, since we expect this to be 4-bytes long.

afongTT (changes requested):

afongTT (dismissed):

lib/tenstorrent/bh_chip/bh_chip.c
```diff
@@ -66,16 +66,17 @@ int bh_chip_set_input_current(struct bh_chip *chip, int32_t *current)
 
 	return ret;
 }
-int bh_chip_set_fan_rpm(struct bh_chip *chip, uint16_t rpm)
+
+int bh_chip_set_input_pwr(struct bh_chip *chip, uint32_t *power)
```
alewycky-tenstorrent, line 70: Please pass `power` by value instead of by pointer.

lib/tenstorrent/bh_arc/cm2bm_msg.h
```diff
@@ -51,7 +51,9 @@ typedef struct bmStaticInfo {
 int32_t Bm2CmSendDataHandler(const uint8_t *data, uint8_t size);
 int32_t Bm2CmPingHandler(const uint8_t *data, uint8_t size);
 int32_t Bm2CmSendCurrentHandler(const uint8_t *data, uint8_t size);
+int32_t Bm2CmSendPwrHandler(const uint8_t *data, uint8_t size);
```
alewycky-tenstorrent, line 54: Why abbreviate to "pwr" here and various other places? "Power" is already one letter shorter than "Current".

lib/tenstorrent/bh_arc/cm2bm_msg.h
```diff
@@ -51,7 +51,9 @@ typedef struct bmStaticInfo {
 int32_t Bm2CmSendDataHandler(const uint8_t *data, uint8_t size);
 int32_t Bm2CmPingHandler(const uint8_t *data, uint8_t size);
 int32_t Bm2CmSendCurrentHandler(const uint8_t *data, uint8_t size);
```
alewycky-tenstorrent, line 53: Why keep Bm2CmSendCurrentHandler? Do we allow version skew between BM and CM? If so then it seems that Bm2CmSendCurrentHandler should be updated to `power = 12 * new_current;` so that it still works.

lib/tenstorrent/bh_arc/smbus_target.c
```diff
@@ -198,6 +198,13 @@ static SmbusConfig smbus_config = {
 			  .handler = {
 					  .rcv_handler = &Bm2CmSetBoardPwrLimit
 			  }},
+		[0x25] = {
+			.valid = 1,
+			.trans_type = kSmbusTransBlockWrite,
+			.expected_blocksize = 4,
+			.handler = {
+					.rcv_handler = &Bm2CmSendPwrHandler
```
alewycky-tenstorrent, line 206: BM I2C is bit-banged so it is very slow (~44KHz). Can we send fewer bytes? Fixed size transfers are smaller than block writes. Can we drop the fractional bits or reduce to 12.4 so that it fits in 16 bits and use a word transfer?

lib/tenstorrent/bh_arc/cm2bm_msg.h
kikkiakki: I don't think we want to allow version skew between BM and CM, so it makes sense to remove it.

lib/tenstorrent/bh_arc/smbus_target.c
kikkiakki: I think this makes sense, I don't think we need this measurement to be super high resolution. I added an update to drop the fractional part and do a word transfer instead.

lib/tenstorrent/bh_arc/smbus_target.c
alewycky-tenstorrent, line 1: If possible, please revert the formatting changes to this file.

lib/tenstorrent/bh_arc/cm2bm_msg.c
```diff
@@ -222,21 +222,21 @@ int32_t Bm2CmPingHandler(const uint8_t *data, uint8_t size)
...
-int32_t Bm2CmSendCurrentHandler(const uint8_t *data, uint8_t size)
+int32_t Bm2CmSendPowerHandler(const uint8_t *data, uint8_t size)
 {
-	if (size != 4) {
+	if (size != 2) {
 		return -1;
 	}
 
-	current = *(int32_t *)data;
+	power = UNALIGNED_GET((uint16_t *)data);
```
alewycky-tenstorrent, line 231: I think [sys_get_le16](https://docs.zephyrproject.org/apidoc/latest/sys_2byteorder_8h.html#a5c2c537b960a29bdd29133f902a67f6a) is ideal. Good catch that we're abusing aligned accesses.

lib/tenstorrent/bh_arc/cm2bm_msg.h
afongTT: Ideally we don't have cases where BM and CM are out of sync, but we still have some issues there (e.g. BMFW fails to update). I'd rather we keep the `Bm2CmSendCurrentHandler` in CMFW just in case.

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -189,24 +183,24 @@ void CalculateThrottlers(void)
 	UpdateThrottler(kThrottlerFastTDC, telemetry_internal_data.vcore_current);
 	UpdateThrottler(kThrottlerTDC, telemetry_internal_data.vcore_current);
 	UpdateThrottler(kThrottlerThm, telemetry_internal_data.asic_temperature);
-	UpdateThrottler(kThrottlerBoardPwr, 12 * ConvertTelemetryToFloat(GetInputCurrent()));
+	UpdateThrottler(kThrottlerBoardPower, ConvertTelemetryToFloat(GetInputPower()));
```
afongTT, line 186: `GetInputPower()` is `uint16_t` and not Telemetry fixed point format, so doesn't need `ConvertTelemetryToFloat`.

lib/tenstorrent/bh_arc/cm2dm_msg.c
```diff
@@ -236,7 +238,18 @@ int32_t Dm2CmSendCurrentHandler(const uint8_t *data, uint8_t size)
 		return -1;
 	}
 
-	current = *(int32_t *)data;
+	current = sys_get_le32(data);
+
+	return 0;
+}
```
afongTT, line 244: Sorry, I guess `current` doesn't get used by anything, so you should just get rid of `Dm2CmSendCurrentHandler`. Not worth the complication to handle both mechanisms.

afongTT (changes requested):

afongTT (approved):

afongTT:
After rebase, can you just make sure we can hit 1350 MHz at idle after GO_BUSY?

kikkiakki:
> After rebase, can you just make sure we can hit 1350 MHz at idle after GO_BUSY?

Confirmed!

afongTT (approved):

Issue #101, opened 2025-04-09
boards: p100a + p150a: smc: jtag srst does not work


cfriedt:
On P100 we have a signal routed for SRST on ARC via debug interfaces. However, on P100A and P150A, we do not. Hopefully on future products we do, since omitting it is a massive footgun.

Therefore, there needs to be a bit of a hack to toggle the reset pin *through BMC* when flashing SMC.

cfriedt:
Just outlining a possible fix:

* `Kconfig` option `CONFIG_TT_TEST_RESET` that gates a periodic task that executes in the system workqueue
  * only needs to be enabled for testing, not in production FW
  * when enabled, a region of SRAM is linked in to the build via linker section, at a predefined location, also determined via Kconfig `CONFIG_TT_TEST_RESET_ADDR`
  * periodic task checks to see if a predefined value `CONFIG_TT_TEST_RESET_KEY` is written to `CONFIG_TT_TEST_RESET_ADDR`
  * if so, then it toggles the pin with node label `asic_reset`, and clears `CONFIG_TT_TEST_RESET_ADDR`
* reset function in `tt_blackhole_smc.cfg` (tcl) calls a shell script in the repository, `scripts/smc-reset-via-bmc.sh`
* shell script calls another instance of openocd
  * this instance of openocd cannot listen on the same ports as the original instance, otherwise it will not run
  * with this instance of openocd, we need to attach to the BMC target, then write `CONFIG_TT_TEST_RESET_KEY` to `CONFIG_TT_TEST_RESET_ADDR`, and then detach from the running BMC target

PR #114, opened 2025-04-10
ci: p100a: temporarily disable p100a tests due to jtag reset issues

cfriedt:
Temporarily disable P100A tests due to lack of JTAG SRST signal.

This affects P150A and other platforms as well. P100 is the only platform not affected.

The tests will be re-enabled when issue https://github.com/tenstorrent/tt-zephyr-platforms/issues/101 is fixed. For any bugfixes targeting P100A+, a "Testing Done" section will need to be included with the PR description, as we are in feature freeze / stabilization.

### DIFF
--- a/.github/workflows/hardware-smoke.yml
+++ b/.github/workflows/hardware-smoke.yml
@@ -132,13 +132,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # Temporarily disable P100A tests due to reset issue #101
+        # Note: YAMLLint complains about 'comment not indented like content', so removed instead
         config:
           - board: p100
             runs-on:
               - yyz-zephyr-lab-p100
-          - board: p100a
-            runs-on:
-              - yyz-zephyr-lab-p100a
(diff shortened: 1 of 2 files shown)


.github/workflows/hardware-smoke.yml
```diff
@@ -20,13 +20,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # Temporarily disable P100A tests due to reset issue #101
+        # Note: YAMLLint complains about 'comment not indented like content', so removed instead
         config:
           - board: p100
             runs-on:
               - yyz-zephyr-lab-p100
-          - board: p100a
```
afongTT, line 27: I think this test is still passing. Maybe keep it?

cfriedt: Well, it's only passing because of happenstance (there is only one test executed by twister), but sure 🤷‍♂️

afongTT (approved):

afongTT (changes requested):
Favour #117 instead

PR #117, opened 2025-04-11
boards: tenstorrent: implement workaround to reset BH SOC on PCIE cards

danieldegrasse:
P100 cards have the BH_RST gpio wired to the ARC JTAG interface, enabling the ARC JTAG to reset the SOC directly. However, P100a/b/c, P150 a/b/c, and P300 a/b/c cards don't connect this GPIO.

To workaround this, we need to directly instruct the BMC to toggle the reset pin of the BH SOC. We accomplish this by starting an openocd instance for the BMC within the reset procedure for the BH SOC. This openocd instance directly sets the GPIO register for the reset pin, then clears it. This allows us to assert the BH_RST pin on cards that do not wire it to the ARC_JTAG interface.

Note that in order to export the proper variables for the openocd command, the build process now creates a temporary file "bmc-reset-vars.cfg" with the board-specific data for the reset procedure, as well as the path to the openocd instance to run. This file is added to gitignore as we don't want to track it.


cfriedt:
Not really going to argue for a better solution if this is good enough.

PR #119, opened 2025-04-11
ci: test to see what SHA github uses for CI runs

danieldegrasse:
See what SHA github uses for CI runs, to make sure it is merged with latest main

PR #120, opened 2025-04-11
ci: workflows: checkout github.ref instead of default SHA

danieldegrasse:
Checkout ${{ github.ref }} instead of the default SHA for the github checkout event. This resolves an issue where restarting CI on a failed job would not merge the PR branch with the latest main, so CI fixes from external PRs would not be picked up.

### DIFF
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -37,6 +37,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           path: tt-zephyr-platforms
+          ref: ${{ github.ref }}
       - uses: ./tt-zephyr-platforms/.github/workflows/prepare-zephyr
         with:
           app-path: tt-zephyr-platforms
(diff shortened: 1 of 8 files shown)


danieldegrasse:
For more details on this, take a look at https://github.com/tenstorrent/tt-zephyr-platforms/pull/119. Essentially the summary is that restarting CI tests doesn't currently update the git SHA used to be a merge with the latest main- with this change, it seems that it does (so we can restart CI runs manually and expect them to pick up fixes)

cfriedt:
Is this ok to wait until after release? I kind of also wonder if it's possible to just rebase to get rid of these issues.

danieldegrasse:
> Is this ok to wait until after release? I kind of also wonder if it's possible to just rebase to get rid of these issues.

Yes and yes- a rebase fixes these issues, I just don't enjoy playing "rebase everyone's PR" after a CI failure in main

cfriedt:
> I just don't enjoy playing "rebase everyone's PR" after a CI failure in main

Yep

danieldegrasse:
@cfriedt what do you want to do with this PR? I reported this behavior as an issue in the upstream github actions repo, but I'm not sure if we want to wait for them to triage that or add this workaround now

cfriedt (approved):

cfriedt:
@danieldegrasse - it improves stability. Would be great if the action were fixed but we can always  switch back later when it is. It's a door that goes both ways.

PR #122, opened 2025-04-11
Remove tt-metal tests that are not part of v55 of metal to which hardware long tests are currently pinned.

dimitri-tenstorrent:
Remove tt-metal tests that are not part of v55 of metal to which hardware long tests are currently pinned.


cfriedt:
@dimitri-tenstorrent - looks like there is just a minor compliance check getting thrown. You will need to amend your commit message and resubmit.

Looks fine from my perspective though.

It might be worth it to comment somewhere that those should be added whenever we are able to test the next metal version.

cfriedt (approved):

PR #123, opened 2025-04-11
lib: tenstorrent: banner: use APPLICATION_SOURCE_DIR for git describe

danieldegrasse:
Use APPLICATION_SOURCE_DIR for the `git describe` step- otherwise, git describe will fail when building the application from outside of the tt-zephyr-platforms repo


danieldegrasse:
Easy enough way to reproduce the issue here:
```
$ west build -p always -b tt_blackhole@p100/tt_blackhole/smc --sysbuild tt-zephyr-platforms/app/smc/
$ strings build/bmc/zephyr/zephyr.signed.bin | grep GIT
```

As an aside, this is my new favorite FW feature

cfriedt (approved):

PR #130, opened 2025-04-15
pcie: improve mantainability

nxuTT:
Added a few comments
- function declarations that should not be changed as pre-compiled library uses them
- highlights functions implemented in pre-compiled library
https://tenstorrent.atlassian.net/browse/SYS-1381

### DIFF
--- a/lib/tenstorrent/bh_arc/pciesd.h
+++ b/lib/tenstorrent/bh_arc/pciesd.h
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2024 Tenstorrent AI ULC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdint.h>
+#include <stdlib.h>
+
+#include "arc_dma.h"
+#include "pcie.h"
+
+/* Verify prototype of ArcDmaTransfer, because it's used by libpciesd.a. */
+__unused static bool (*verify_ArcDmaTransfer)(const void *, void *, uint32_t) = ArcDmaTransfer;
+
+/* The functions below are implemented in tt_blackhole_libpciesd.a */
+PCIeInitStatus SerdesInit(uint8_t pcie_inst, PCIeDeviceType device_type,
+			  uint8_t num_serdes_instance);
+void ExitLoopback(void);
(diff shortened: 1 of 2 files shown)


alewycky-tenstorrent:
Ideas: 
- move the externally-implemented functions into a header for the library, "libpciesd.h"
- That header should have a comment saying that it represents the API for that library.
- The comment helps but doesn't enforce ArcDma's interface. You could enforce it by assigning ArcDmaTransfer to a function pointer in a way that the compile will fail if the prototype changes. This is unused so maybe you have to convince the compiler not to warn about it.
```
/* Verify prototype of ArcDmaTransfer, because it's used by libpciesd.a. */
static (*bool verify_ArcDmaTransfer)(const void *, void *, uint32_t) = ArcDmaTransfer;
```

lib/tenstorrent/bh_arc/pciesd.h
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) 2024 Tenstorrent AI ULC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdlib.h>
+#include "arc_dma.h"
+
+/* Verify prototype of ArcDmaTransfer, because it's used by libpciesd.a. */
+static bool (*verify_ArcDmaTransfer)(const void *, void *, uint32_t)
+	__attribute__((unused)) = ArcDmaTransfer;
```
alewycky-tenstorrent, line 11: Use zephyr's `__unused`

lib/tenstorrent/bh_arc/pciesd.h
```diff
@@ -0,0 +1,19 @@
...
+
+/* The functions below are implemented in tt_blackhole_libpciesd.a */
+PCIeInitStatus SerdesInit(uint8_t pcie_inst, PCIeDeviceType device_type,
```
alewycky-tenstorrent, line 14: include header file that defines PCIeInitStatus and PCIeDeviceType

I think you need stdint.h too

alewycky-tenstorrent (approved):

PR #131, opened 2025-04-16
lib: tenstorrent: banner: add print of Zephyr SDK revision

danieldegrasse:
Add print of Zephyr SDK revision during boot. This will enable truly reproducible builds, since we will be able to use the same toolchain and SHA.


danieldegrasse:
@cfriedt if you think it is reasonable I'd like to put this in pre-release, BMFW 0.3.1 has given me a special kind of paranoia about reproducible builds

cfriedt (approved):

PR #132, opened 2025-04-16
drivers: flash: move to mspi based flash nor driver

danieldegrasse:
Move to the MSPI based flash nor driver present in upstream Zephyr. This will allow us to long term remove all downstream flash drivers for blackhole, and rely exclusively on upstream support.

~Draft PR for now, as I still need to implement support for resetting the SPI flash at boot time to properly support BH galaxy systems (which reconfigure the SPI flash). The driver also appears significantly slower, which needs to be debugged.~


danieldegrasse:
This driver should properly support BH galaxy- I have verified that the driver initializes and works correctly on a scrappy board when the bootrom GPIO straps are set to switch the SPI EEPROM into octal DDR mode

nxuTT:
> This driver should properly support BH galaxy- I have verified that the driver initializes and works correctly on a scrappy board when the bootrom GPIO straps are set to switch the SPI EEPROM into octal DDR mode

That is good to hear! Does it also work for OCTAL DDR and DQS? What about QUAD DDR?

danieldegrasse:
> > This driver should properly support BH galaxy- I have verified that the driver initializes and works correctly on a scrappy board when the bootrom GPIO straps are set to switch the SPI EEPROM into octal DDR mode
> 
> That is good to hear! Does it also work for OCTAL DDR and DQS? What about QUAD DDR?

It does work with DQS. How would I verify QUAD DDR though? Can we set the GPIO straps on the P150a/P100a boards? The MT35 flash on scrappy won't run in quad mode.

nxuTT:
> > > This driver should properly support BH galaxy- I have verified that the driver initializes and works correctly on a scrappy board when the bootrom GPIO straps are set to switch the SPI EEPROM into octal DDR mode
> > 
> > 
> > That is good to hear! Does it also work for OCTAL DDR and DQS? What about QUAD DDR?
> 
> It does work with DQS. How would I verify QUAD DDR though? Can we set the GPIO straps on the P150a/P100a boards? The MT35 flash on scrappy won't run in quad mode.

It should work on p150A and p100A, they use MT25. Those are Quanta boards. I think only Scrappy boards use MT35.

drivers/flash/flash_mspi_nor_devs.c
```diff
@@ -0,0 +1,332 @@
+/*
+ * Copyright (c) 2025 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef __FLASH_MSPI_NOR_DEVS_H__
+#define __FLASH_MSPI_NOR_DEVS_H__
```
cfriedt, line 8: This is kind of unusual - is this file being included into another .c file? Since all of the objects are `const` rather than `static const` it could just be compiled.

danieldegrasse: No, this file started as a header file- I'll remove the include guards

danieldegrasse:
> It should work on p150A and p100A, they use MT25. Those are Quanta boards. I think only Scrappy boards use MT35. If you see PCIE5.0 cables on the side, then they are Quanta boards. Scrappy uses 3 CPU connectors on the side.

PR has been updated, I validated on a P100a using the MT25 flash in DDR mode (set using the GPIO strap pins)

So to summarize, this driver has been checked with the following modes:
- Octal DDR on MX35, on P100
- Octal standard on MX35, on P100
- 1-8-8 mode on MX35, on P100
- Quad DDR on MT25, on P100a
- 1-4-4 mode of MT25, on P100a

nxuTT:
> > It should work on p150A and p100A, they use MT25. Those are Quanta boards. I think only Scrappy boards use MT35. If you see PCIE5.0 cables on the side, then they are Quanta boards. Scrappy uses 3 CPU connectors on the side.
> 
> PR has been updated, I validated on a P100a using the MT25 flash in DDR mode (set using the GPIO strap pins)
> 
> So to summarize, this driver has been checked with the following modes:
> 
> * Octal DDR on MX35, on P100
> * Octal standard on MX35, on P100
> * 1-8-8 mode on MX35, on P100
> * Quad DDR on MT25, on P100a
> * 1-4-4 mode of MT25, on P100a

Can I try it on BH Galaxy?

cfriedt (approved):

cfriedt:
@danieldegrasse - if you can rebase this on the changes that add `p150a` to CI, and it passes, I think it would be smart to get it in, if possible.

cfriedt:
> Can I try it on BH Galaxy?

@nxuTT  - Sorry - just saw this. Would be great to get it checked on BH Galaxy as well.

danieldegrasse:
> > Can I try it on BH Galaxy?
> 
> @nxuTT - Sorry - just saw this. Would be great to get it checked on BH Galaxy as well.

Agree- I think for that we will need to move the CMFW app over to using this SPI driver. I'm going to create the PR to do that, and if that PR passes CI as well (and runs on Galaxy) we should merge both IMO

danieldegrasse:
Closing since #206 includes these changes, and enables the driver in the CMFW

PR #133, opened 2025-04-17
i2c: adjust mask for iout cal

nxuTT:
Should overwrite the entire 11 bits of mantissa field for IOUT_CAL_GAIN and IOUT_CAL_OFFSET, exponent fields are not overwritable.

https://tenstorrent.atlassian.net/browse/SYS-1430


alewycky-tenstorrent (approved):

cfriedt (changes requested):
Can we make this PR pass CI?

cfriedt (approved):

cfriedt:
Restarted the system with a cold boot - tests pass again.

This highlights possibly another issue since we can reset the ASIC, but the board can still fail to enumerate due to lingering state from off-chip peripherals.

cfriedt:
follow-up ticket:
https://tenstorrent.atlassian.net/browse/SYS-880

cfriedt:
Looks like this might have been a regression from 667487f6acc482d1eeabb99280ef7dd657fa91a2

PR #135, opened 2025-04-21
tests: drivers: flash: add a flash performance test

danieldegrasse:
Add a flash performance test. This test validates the performance of the `bh_arc` EEPROM driver as well and the Zephyr native flash driver, and verifies that performance has not regressed.


danieldegrasse:
A note from the results of this test locally- it looks like the native flash driver significantly improves read performance, but we should keep the functionality provided by `SpiSmartWrite`, as that significantly improves program times

cfriedt (approved):

PR #136, opened 2025-04-21
Use Metal Container for running Hardware Long tests & Add required Metal tests

dimitri-tenstorrent:
To update the metal version to the latest version, we need to change how we install dependencies in the workflows.

- Broke down the workflow into two jobs
- Got rid of manual installation of metal and started using a metal container 
- Upgraded metal to ~v0.56.0~ ~almeet's branch `abhullar/bh-virtual` ~ now on Metal main.

TODO:
- [x] Figure out the unit-tests errors
- [x] Reset of P100a does not seem to be working

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  hardware-metal-test:
+  hardware-zephyr-test:
     strategy:
       fail-fast: false
       matrix:
@@ -137,25 +137,31 @@ jobs:
             --hardware-map /opt/tenstorrent/twister/hw-map.yml --west-flash \
             -T ../tt-zephyr-platforms/app -ll DEBUG
 
-
-      - name: Fix APT URLs
-        run: |
-          # Patch APT URLs to use https
-          sed -i -e 's/http/https/g' /etc/apt/sources.list
-          apt update
-
-      - name: Install clang 17
-        run: |
-          wget https://apt.llvm.org/llvm.sh
-          sed -i "s/http:\/\/apt.llvm.org/https:\/\/apt.llvm.org/" -i llvm.sh
-          chmod +x llvm.sh
-          ./llvm.sh 17 all
-          rm -f llvm.sh
-
+  hardware-metal-test:
+    needs:
+      - hardware-zephyr-test
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - board: p100
+            runs-on:
+              - yyz-zephyr-lab-p100
+          - board: p100a
+            runs-on:
+              - yyz-zephyr-lab-p100a
+    runs-on: ${{ matrix.config.runs-on }}
+    container:
+      image: ghcr.io/tenstorrent/tt-metal/tt-metalium/ubuntu-22.04-dev-amd64
+      volumes:
+        - /dev/hugepages-1G:/dev/hugepages-1G
+        - /dev/hugepages:/dev/hugepages
+      options: '--device /dev/tenstorrent --device /dev/bus/usb --privileged'
+    steps:
       - name: Install Git LFS
         run: |
+          apt update
           apt install -y git-lfs
-
       - name: Checkout Metal
         uses: actions/checkout@v4
         with:
@@ -163,19 +169,7 @@ jobs:
           path: tt-metal
           lfs: true
           submodules: true
-          ref: v0.55.0
-
-      - name: Install Metal Dependencies
-        working-directory: tt-metal
-        continue-on-error: true
-        shell: bash
-        run: |
-          sed -i 's/apt-get install/apt-get install -y/' install_dependencies.sh
-          sed -i 's/python3.8-venv/python3.10-venv/' install_dependencies.sh
-          apt install -y python3.10-dev python3.10-venv libpython3.10-dev python3.10
-          # This script will fail to install the tt-hugepages service, just ignore it
-          ./install_dependencies.sh
-
+          ref: v0.56.0
       - name: Run Metal Tests
         working-directory: tt-metal
         shell: bash
@@ -192,13 +186,9 @@ jobs:
           # Run metal tests
           tests/scripts/run_cpp_unit_tests.sh
           TT_METAL_SLOW_DISPATCH_MODE=1 tests/scripts/run_cpp_unit_tests.sh
+
       - name: cleanup
         if: ${{ always() }}
         run: |
-          # Clean up patched Zephyr repo
-          west patch clean
           # Clean out metal
           rm -rf tt-metal
-          # Cleanup the checked out repo, we can leave everything else
-          rm -rf tt-zephyr-platforms
-          rm -rf .west
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -187,7 +187,7 @@ jobs:
           path: tt-metal
           lfs: true
           submodules: true
-          ref: abhullar/bh-virtual
+          ref: main
           fetch-depth: 500
           fetch-tags: true
       - name: Build Metal UMD Tests
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -159,7 +159,7 @@ jobs:
               - yyz-zephyr-lab-p100a
     runs-on: ${{ matrix.config.runs-on }}
     container:
-      image: ghcr.io/tenstorrent/tt-metal/tt-metalium/ubuntu-22.04-dev-amd64:latest
+      image: ghcr.io/tenstorrent/tt-metal/tt-metalium/ubuntu-22.04-dev-amd64:bd47135f35f7b26851eb09066076fb3e6074c54f # Change this to a tag when metal has release versions for this image.
       volumes:
         - /dev/hugepages-1G:/dev/hugepages-1G
         - /dev/hugepages:/dev/hugepages
@@ -188,7 +188,7 @@ jobs:
           path: tt-metal
           lfs: true
           submodules: true
-          ref: main
+          ref: 12652be514e62224b5ccb160e5248fc83dfd96c1 # Change this when metal revision is available.
           fetch-depth: 500
           fetch-tags: true
       - name: Build Metal UMD Tests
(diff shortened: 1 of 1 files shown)


cfriedt:
@dimitri-tenstorrent - do tests pass when manually run using this change?

In many cases, especially because FW lacks continuous testing in many areas, it might be good to include a "Testing Done" snippet of console output with PRs.

That way, it should be easier to reproduce successful runs.

We did discover an issue last week where if voltage regulators are incorrectly programmed, then subsequent tests will fail in CI and it requires a power-cycle to resolve.

Metal _shouldn't_ be causing that kind of issue, but we also unfortunately have never had tests added for firmware in that area (and many other areas, something I will require after the next release).

cfriedt:
Please squash all commits into one, fix up the compliance failure (I think just the commit message), and push again.

danieldegrasse:
@dimitri-tenstorrent I think this change makes sense, from what I can see this basically saves us from installing build dependencies? I agree with @cfriedt- these changes should be squashed to 1 (or perhaps a few) commits. Maybe one commit to uprev metal to v0.56, and one commit to split the test into two jobs and use the container?

I don't think we can expect the P100a to pass metal tests right now- some of the GDDR on the card is harvested and does not function correctly (even if soft harvesting is disabled in the FW). We may just want to skip P100a outright for now.

Another thing I'd like your opinion on is this PR: https://github.com/tenstorrent/tt-zephyr-platforms/pull/99 I want to get us to the point where we can build and test a new revision of metal within the CI for this repo, then reuse that revision as a cached CI artifact in other CI jobs. Does that could reasonable?

danieldegrasse:
@dimitri-tenstorrent can you add a link to a manual dispatch of this workflow so we can see the results with these changes to the CI job?

.github/workflows/hardware-long.yml
```diff
@@ -133,45 +133,39 @@ jobs:
...
-        continue-on-error: true
-        shell: bash
-        run: |
-          sed -i 's/apt-get install/apt-get install -y/' install_dependencies.sh
-          sed -i 's/python3.8-venv/python3.10-venv/' install_dependencies.sh
-          apt install -y python3.10-dev python3.10-venv libpython3.10-dev python3.10
-          # This script will fail to install the tt-hugepages service, just ignore it
-          ./install_dependencies.sh
-
+          ref: v0.56.0
```
cfriedt, line 168: @dimitri-tenstorrent - in order to test @abhullar-tt's Metal PR [18131](https://github.com/tenstorrent/tt-metal/pull/18131) here, you would need to change this ref to:

```
pull/18131/head
```

cfriedt: That will allow us to verify that Metal works with RC2 of `tt-zephyr-platforms` prior to cutting a release of `tt-zephyr-platforms`.

This is the mechanism we need to use going forward to coordinate releases and ensure that there are no regressions.

.github/workflows/hardware-long.yml
```diff
@@ -11,17 +11,17 @@ concurrency:
...
         config:
           - board: p100
             runs-on:
               - yyz-zephyr-lab-p100
-          - board: p100a
-            runs-on:
-              - yyz-zephyr-lab-p100a
+          # - board: p100a
+          #   runs-on:
+          #     - yyz-zephyr-lab-p100a
```
cfriedt: Just remove this - there is 4-year-old a bug in yamllint for it.

https://github.com/adrienverge/yamllint/issues/375

.github/workflows/hardware-long.yml
```diff
@@ -136,69 +137,97 @@ jobs:
...
-          chmod +x llvm.sh
-          ./llvm.sh 17 all
-          rm -f llvm.sh
-
+          apt install -y cargo
+          pip install git+https://github.com/tenstorrent/tt-smi@v3.0.5
+          tt-smi -r 0
+  hardware-metal-test:
+    needs:
+      - hardware-zephyr-test
```
cfriedt, line 149: indentation looks wrong here.

.github/workflows/hardware-long.yml
```diff
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  hardware-metal-test:
+  hardware-zephyr-test:
```
cfriedt, line 14: Do we need to rename this job?

dimitri-tenstorrent: Since the workflow was split into two jobs, this job is no longer doing metal tests. What do you think is a better name for this job?

.github/workflows/hardware-long.yml
```diff
@@ -136,69 +137,97 @@ jobs:
...
           apt install -y git-lfs
-
+      - name: Checkout Zephyr Platforms (only for rescan-pcie.sh)
+        uses: actions/checkout@v4
+        with:
+          path: tt-zephyr-platforms
+          ref: ${{ github.ref }}
+      - name: Run the rescan-pcie.sh script
+        run: |
+          ./tt-zephyr-platforms/scripts/rescan-pcie.sh
```
cfriedt, line 183: It's possible that flashing via `tt-flash` will flash + rescan pcie here. I vaguely recall that someone said something about `tt-smi -r` and `tt-flash` doing that.

In any case, it seems unnecessary to check out tt-zephyr-platforms for 1 script.

Maybe just this?
```suggestion
      - name: Rescan PCIe
        run: |
          curl -o /tmp/rescan-pcie.sh \
          'https://github.com/tenstorrent/tt-zephyr-platforms/blob/cc07066/scripts/rescan-pcie.sh'
           /tmp/rescan-pcie.sh
```

.github/workflows/hardware-long.yml
```diff
@@ -136,69 +137,97 @@ jobs:
...
           repository: tenstorrent/tt-metal
           path: tt-metal
           lfs: true
           submodules: true
-          ref: v0.55.0
-
-      - name: Install Metal Dependencies
+          ref: main
+          fetch-depth: 500
+          fetch-tags: true
```
cfriedt, line 193: I think we need to pin to a specific revision

.github/workflows/hardware-long.yml
```diff
@@ -136,69 +137,97 @@ jobs:
...
+        config:
+          - board: p100
+            runs-on:
+              - yyz-zephyr-lab-p100
+          - board: p100a
+            runs-on:
+              - yyz-zephyr-lab-p100a
+    runs-on: ${{ matrix.config.runs-on }}
+    container:
+      image: ghcr.io/tenstorrent/tt-metal/tt-metalium/ubuntu-22.04-dev-amd64:latest
```
cfriedt, line 162: This should be pinned to a specific revision

.github/workflows/hardware-long.yml
dimitri-tenstorrent: There is no current specific version and it was just merged in. I will change as soon as there is a specific version. I will change to the commit sha for now.

.github/workflows/hardware-long.yml
dimitri-tenstorrent: @cfriedt Could you identify the unnecessary steps?
The idea that we would build the current commit firmware and flash it onto the card and then run metal tests.

.github/workflows/hardware-long.yml
dimitri-tenstorrent: tt-flash and tt-smi -r did not do that in my testing on the lab machine. I am happy to be wrong, but for now only rescanning works.

.github/workflows/hardware-long.yml
dimitri-tenstorrent: Set to the commit reference where @abhullar-tt merged her PR.

.github/workflows/hardware-long.yml
dimitri-tenstorrent: Set to a recent docker image sha and will change once metal releases tagged versions of this image.

.github/workflows/hardware-long.yml
dimitri-tenstorrent: To clarify, I do not fully understand the steps that include testing, building board metadata, etc. Please highlight the ones that are not necessary :)

.github/workflows/hardware-long.yml
cfriedt: > I like this! Maybe I would just use `git_ref` to get the current version of the script :)

There is a git ref in the link:

> https://github.com/tenstorrent/tt-zephyr-platforms/blob/cc07066/scripts/rescan-pcie.sh

.github/workflows/hardware-long.yml
dimitri-tenstorrent: Could we address this as part of @kkabilarTT's PR? I do not want to redo how we flash in this workflow. I am happy to dive in after but this is a very important unit of work.

.github/workflows/hardware-long.yml
dimitri-tenstorrent: Addressed.

dimitri-tenstorrent:
> @dimitri-tenstorrent can you add a link to a manual dispatch of this workflow so we can see the results with these changes to the CI job?

https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14753944860

.github/workflows/hardware-long.yml
cfriedt: Ah sorry - it's not the indentation, it's that another job was defined.

All steps should all be part of the same job.

.github/workflows/hardware-long.yml
```diff
@@ -136,69 +137,93 @@ jobs:
...
-          sed -i "s/http:\/\/apt.llvm.org/https:\/\/apt.llvm.org/" -i llvm.sh
-          chmod +x llvm.sh
-          ./llvm.sh 17 all
-          rm -f llvm.sh
-
+          apt install -y cargo
+          pip install git+https://github.com/tenstorrent/tt-smi@v3.0.5
+          tt-smi -r 0
+  hardware-metal-test:
+    needs:
```
cfriedt, line 148: This needs to be part of the previously defined job if all steps are to be executed in sequence. Otherwise, steps from other queued jobs will be executed between these two.

danieldegrasse (dismissed):

.github/workflows/hardware-long.yml
dimitri-tenstorrent: @cfriedt I had to follow the advice above. We probably compromised the ability to test individual commits but we will re-add this with @kkabilarTT 's PR.

PR #139, opened 2025-04-21
bh_arc: fix I2C recovery sequence

danieldegrasse:
At the final step of the I2C recovery sequence, we need to release the SDA line to allow it to be pulled high. Fix the GPIO TRIEN bitmask we write so that this will actually occur.


cfriedt (approved):
LGTM - @afongTT - I think this should go into the release I'll be doing tonight. Would you agree?

PR #148, opened 2025-04-23
Revert "ci: flash a stable copy of BMFW and CMFW prior to HW CI runs"

danieldegrasse:
This reverts commit 34a1b20cf563390a2c7420950c173e1e8c8b446c. Reverting due to CI failures seen with tt-flash in some runs.


cfriedt:
@danieldegrasse  - will the failing test pass if we re-run CI in this PR? I want to avoid the p100a runner getting stuck in a bad state if we can.

danieldegrasse:
Closing this out, as it doesn't seem to be needed.

PR #151, opened 2025-04-24
scripts: bmc-reset: add support for waiting for SMC boot, autodetect paths

danieldegrasse:
Add support in the bmc-reset script for waiting for the SMC boot, and autodetect paths like the Zephyr SDK where possible.

~Use this script within CI, so that we can have more reliable BMC flashing~

### DIFF
--- a/scripts/bmc-reset.py
+++ b/scripts/bmc-reset.py
@@ -22,6 +22,10 @@ from pathlib import Path
 
 logger = logging.getLogger(Path(__file__).stem)
 
+DEFAULT_BMC_CFG = (
+    Path(__file__).parents[1]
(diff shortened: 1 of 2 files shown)


cfriedt (commented):
I don't think that we should make this change. This script was supposed to work without any context - repo, SDK, etc. I symlinked the BMC config file on one CI machine and copied the generated vars.

Probably better to focus on pyocd and a gpio toggle thing and then eventually integrate that into our tt_flash west runner.

danieldegrasse:
Updated PR to no longer use script in CI

cfriedt (approved):

PR #152, opened 2025-04-24
Fix perst handler to avoid deadlock

nxuTT:
- remove reset_lock mutex, as reset can only happen in main thread
- move PREST reset handler sequence into main
- move therm trip handler to main thread
- use event to wake up main thread if main thread is sleeping, therm trip or perst could send wake up event


include/tenstorrent/bh_chip.h
```diff
@@ -51,6 +51,9 @@ struct bh_chip_data {
 	bool arc_just_reset;
 
 	unsigned int bus_cancel_flag;
+
+	/* notify the main thread to apply reset sequence */
+	atomic_t trigger_reset;
```
alewycky-tenstorrent, line 56: Needs init, see [ATOMIC_INIT](https://docs.zephyrproject.org/apidoc/latest/group__atomic__apis.html#gaadfbba86627ee7eeb07e04f712550f73)

include/tenstorrent/bh_chip.h
```diff
@@ -54,6 +54,9 @@ struct bh_chip_data {
 
 	/* notify the main thread to apply reset sequence */
 	atomic_t trigger_reset;
+
+	/* notify the main thread to handle therm trip */
+	atomic_t therm_trip_triggered;
```
alewycky-tenstorrent, line 59: Also needs ATOMIC_INIT

alewycky-tenstorrent:
Please rename commits, "move X handler to main thread" should be first line.

alewycky-tenstorrent:
Oops, atomic operations are faked using C code & spinlocks on some platforms including ours. Just use bool instead.

TTDRosen (commented):
Need to have reset.c check for trigger_reset to be asserted to cover the case where PERST triggers during the writing of the bootrom payload.

TTDRosen:
Broader question, is the intention to move the therm trip and reset request routines to a workqueue once we've done a proper review of priorities?

app/bmc/src/main.c
```diff
@@ -303,6 +303,31 @@ int main(void)
...
+					set_fan_speed(100);
+				}
+				bh_chip_reset_chip(chip, true);
+			}
+		}
+
+		/* handler for PERST */
+		ARRAY_FOR_EACH_PTR(BH_CHIPS, chip) {
+			if (atomic_test_and_clear_bit(&chip->data.trigger_reset, 0)) {
+				if (chip->data.workaround_applied) {
```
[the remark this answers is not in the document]
nxuTT: You want to move bh_chip_cancel_bus_transfer_clearfrom line 329 to after 320?

alewycky-tenstorrent:
> Broader question, is the intention to move the therm trip and reset request routines to a workqueue once we've done a proper review of priorities?

Probably? We'll determine that in the priority review. It sounds like the interaction between therm trip and reset is kind of subtle.

app/bmc/src/main.c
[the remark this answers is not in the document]
nxuTT: I'm doing i2c_bus_transfer_set in ISR

alewycky-tenstorrent:
The code changes in the commit that removes atomics are fine, but please put those changes into the original commits.

app/bmc/src/main.c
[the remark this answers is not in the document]
TTDRosen: Whoops my bad. Looking at the code again I made some really weird decisions. The reset lock was a lock that should have been inside the i2c driver to wait for the i2c transaction to actually complete before starting the reset proper. I think I probably made that decision so I could have one lock in case there were other busses that I wanted to be clear before starting the reset, but I don't know I would have used a mutex in that case.

lib/tenstorrent/bh_chip/bh_chip.c
```diff
@@ -113,6 +114,9 @@ void therm_trip_detected(const struct device *dev, struct gpio_callback *cb, uin
 	struct bh_chip *chip = CONTAINER_OF(cb, struct bh_chip, therm_trip_cb);
 	chip->data.therm_trip_triggered = true;
 	bh_chip_cancel_bus_transfer_set(chip);
+
+	uint8_t wakeup_msg = 0;
+	k_msgq_put(&main_queue, &wakeup_msg, K_NO_WAIT);
```
alewycky-tenstorrent: What if move this to the header, "static inline void wake_main_loop(void) { ... }"?

app/bmc/src/main.c
```diff
@@ -301,7 +305,8 @@ int main(void)
 		(bmStaticInfo){.version = 1, .bl_version = 0, .app_version = APPVERSION};
 
 	while (1) {
-		k_sleep(K_MSEC(20));
+		uint8_t wakeup_msg;
+		k_msgq_get(&main_queue, &wakeup_msg, K_MSEC(20));
```
alewycky-tenstorrent: What if move this to the header: "static inline void sleep_main_thread(k_timeout_t timeout) { ... }"?

app/bmc/src/main.c
```diff
@@ -23,10 +23,14 @@
 #include <tenstorrent/tt_smbus.h>
 #include <tenstorrent/bh_chip.h>
 #include <tenstorrent/bh_arc.h>
+#include <tenstorrent/bmc_main_queue.h>
 #include <tenstorrent/jtag_bootrom.h>
 
 LOG_MODULE_REGISTER(main, CONFIG_TT_APP_LOG_LEVEL);
 
+#define MAIN_QUEUE_SIZE 10
+K_MSGQ_DEFINE(main_queue, sizeof(uint8_t), MAIN_QUEUE_SIZE, _Alignof(uint8_t));
```
alewycky-tenstorrent, line 32: Remove MAIN_QUEUE_SIZE, replace with 1.

include/tenstorrent/bmc_main_queue.h
cfriedt, line 1: This is excessive. Can it just be put in the application?

nxuTT: I don't know where to put extern struct k_msgq main_queue, because there's multiple users in lib/tenstorrent

app/bmc/src/main.c
```diff
@@ -23,10 +23,13 @@
 #include <tenstorrent/tt_smbus.h>
 #include <tenstorrent/bh_chip.h>
 #include <tenstorrent/bh_arc.h>
+#include <tenstorrent/bmc_main_queue.h>
```
cfriedt, line 26: Can we not include this header please?

It would be cleaner to just call the functions directly.

There are also other ways to wait for an event. E.g. k_poll, k_event, etc.

Those allow you to wake on multiple kernel objects or multiple different types of events.

include/tenstorrent/bmc_main_queue.h
cfriedt: I would add another .h file e.g. `include/tenstorrent/bm_event.h` with

```cpp
/* both isr safe */
int bm_event_set(uint32_t event);
int bm_event_post(uint32_t events);

/* call this from main() */
int bm_event_wait(uint32_t events, k_timeout_t timeout);
```

and .c file e.g. `lib/tenstorrent/bm_event/bm_event.c`.

```cpp
static K_EVENT_DEFINE(bm_event);

int bm_event_set(uint32_t event)
{
  return k_event_set(&bm_event, event);
}

int bm_event_post(uint32_t events)
{
  return k_event_post(&bm_event, events);
}

int bm_event_wait(uint32_t events, k_timeout_t timeout)
{
  return k_event_wait(&bm_event, events, false, timeout);
}
```

That way, the app can depend on jtag_bootrom, bh_chip, and bm_event, preserving the direction of dependencies.

jtag_bootrom and bh_chip would then also depend on bm_event.

app/bmc/src/main.c
```diff
@@ -23,10 +23,13 @@
...
 #include <tenstorrent/jtag_bootrom.h>
 
 LOG_MODULE_REGISTER(main, CONFIG_TT_APP_LOG_LEVEL);
 
+K_MSGQ_DEFINE(main_queue, sizeof(uint8_t), 1, _Alignof(uint8_t));
```
cfriedt, line 31: A `k_event` object lets you pass in a bitmask of possible events (which might scale better if we want to handle other events in our `main()` function).

From a dependency point of view though, the app depends on the library. If the message queue is defined in the app, then the library would depend on the app, which is kind of the opposite.

So best to go with a mutual dependency that the app can depend on, as well as jtag_bootrom and bh_chip

cfriedt:
@nxuTT - definitely looks like a good fix to me - just some suggestions about where to put the object.

include/tenstorrent/bmc_main_queue.h
nxuTT: Thank you for your suggestions, I'll fix my MR.

lib/tenstorrent/bm_event/Kconfig
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+config TT_BM_EVENT
+	bool "BM event support"
+	help
+	  BM event support.
+
+module = TT_BM_EVENT
+module-str = TT BM event
+source "subsys/logging/Kconfig.template.log_config"
```
cfriedt, line 11: This is a really good idea - you can declare the bm_event log source with this in a header

```cpp

LOG_MODULE_DECLARE(bm_event);
```

And then in a c file (approximately) something like
```cpp

LOG_MODULE_REGISTER(bm_event, CONFIG_BM_EVENT_LOG_LEVEL);
```

and then you can `LOG_INFO(..)` anywhere. I don't know if there is any possibility of error, but it's always a good idea to log errors (at least until they are very rare, and then we can change them later to `LOG_DBG(...)`

lib/tenstorrent/bm_event/Kconfig
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2024 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+config TT_BM_EVENT
+	bool "BM event support"
+	help
+	  BM event support.
```
cfriedt, line 7: Didn't see this on Friday, but I'll approve anyway. Feel free to add it now or in a follow-up PR.

```suggestion
config TT_BM_EVENT
	bool "BM event support"
	depends on EVENTS
	help
	  BM event support.
```

cfriedt (dismissed):

lib/tenstorrent/bm_event/Kconfig
nxuTT: Fixed!

lib/tenstorrent/bm_event/bm_event.c
```diff
@@ -0,0 +1,29 @@
...
+
+uint32_t bm_event_wait(uint32_t events, k_timeout_t timeout)
+{
+	uint32_t ret = k_event_wait(&bm_event, events, false, timeout);
+
+	if (ret == 0) {
+		LOG_INF("Failed to wait for events: 0x%08X", events);
+	} else {
+		LOG_INF("Event wait successful: requested=0x%08X, received=0x%08X", events, ret);
+	}
```
cfriedt, line 26: Sorry I didn't catch this earlier, but logging if no events are received might spam the logs if our timeout is only 20ms. We would see this message 50x per second if no events are received.

```suggestion
	if (ret != 0) {
		LOG_INF("Event wait successful: requested=0x%08X, received=0x%08X", events, ret);
	}
```

nxuTT: Done!

app/bmc/src/main.c
alewycky-tenstorrent: Sorry, the above comment got stuck in a pending review. Given that ISR safety is the entire point of this change, using isr-ok routines is imperative.

cfriedt (approved):

PR #153, opened 2025-04-24
scripts: add rtt helpers to dump RTT data

danieldegrasse:
Add helpers to dump RTT data for SMC and BMC. These helpers will not reset the system, but simply dump state. They can be useful for debugging a hung SMC or BMC


scripts/bmc_rtt.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2025 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Script to open RTT console to SMC
```
cfriedt, line 7:
```suggestion
Script to open RTT console to BMC
```

cfriedt (approved):
One typo.

I guess this is a workaround for Anas NAK'ing re-using the existing `west rtt` support for `twister`?

cfriedt (approved):

PR #157, opened 2025-04-25
Revert "ci: flash a stable copy of BMFW and CMFW prior to HW CI runs"

danieldegrasse:
This reverts commit 34a1b20cf563390a2c7420950c173e1e8c8b446c.

Here we go again....


cfriedt (approved):
Our CI seems to behave a bit like Schrodinger's cat

PR #160, opened 2025-04-26
scripts: requirements: add tt-smi to required python packages

cfriedt:
Ensure that `tt-smi` is installed (this is typically done via `west packages pip --install`).

Python packages are installed as part of `zephyrproject-rtos/action-zephyr-setup`, so remove the additional manual `pip` invocation.

https://github.com/zephyrproject-rtos/action-zephyr-setup/blob/db1412e2ab8e2c07b1f866541e66aec41b20d244/action.yml#L141

PR #161, opened 2025-04-26
ci: temporarily disable p100a hw tests until it runs reliably

cfriedt:
p100a machines need manual recovery almost constantly. p100 seems fine. 

It's a massive waste of developer time and is blocking CI, so just disable p100a until we can reliably run tests on that platform as part of the regular test sequence (which I hope is ASAP).

Case in point:
* https://github.com/tenstorrent/tt-zephyr-platforms/pull/160
* the PR above simply adds `tt-smi` to `scripts/requirements.txt`
* there should be zero reason for CI failing on p100a hardware tests

Likely the same problem will plague p150a, p150b, p150c, p300a, p300b, and p300c systems, so it needs to be done right.

Temporary workaround for #162, but that will need a fix ASAP.

### DIFF
--- a/.github/workflows/hardware-smoke.yml
+++ b/.github/workflows/hardware-smoke.yml
@@ -139,9 +139,6 @@ jobs:
           - board: p100
             runs-on:
               - yyz-zephyr-lab-p100
-          - board: p100a
-            runs-on:
-              - yyz-zephyr-lab-p100a
     runs-on: ${{ matrix.config.runs-on }}
     env:
       "ZEPHYR_SDK_INSTALL_DIR": /opt/toolchains


cfriedt:
* rebased
* reduced scope to only disable `smoke-e2e-test` for `p100a` since it is still failing consistently on `main`

.github/workflows/hardware-smoke.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           - board: p100a
             runs-on:
               - yyz-zephyr-lab-p100a
-    runs-on: ${{ matrix.config.runs-on }}
+      runs-on: ${{ matrix.config.runs-on }}
```
nxuTT, line 33: This is a typo?

nxuTT (approved):

.github/workflows/hardware-smoke.yml
cfriedt: Yep, fixed

Issue #158, opened 2025-04-25
Build firmware as an artifact for Hardware Long workflow


kkabilarTT:
Currently, the firmware bundle is built from source each time the Hardware Long workflow is triggered. It needs to be changed to firmware already built as an artifact so other tests can run on top of it.

dimitri-tenstorrent:
Please see the building of the firmware from source and flashing directly in https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/.github/workflows/hardware-long.yml

Instead what we should be striving for is downloading a bundle from a previous step as an artifact. 

The bundle is already being built in another workflow https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/.github/workflows/release.yml so it is a matter of adapting the steps from the Release workflow in the Hardware long and other workflows.

This will allow us to test against the bundles that we will be releasing and improve the certainty and trust in the release bundles.

cfriedt:
@dimitri-tenstorrent - firmware bundles are built here https://github.com/tenstorrent/tt-zephyr-platforms/blob/22b4cd7b3761529e3de469055815b29a779ae9bc/.github/workflows/build-fw.yml#L149 and then uploaded as artifacts.

Expand "Upload Combined Firmware Bundle Assets" at the URL below.

https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14681360178/job/41204500608?pr=161

There is no definitely no reason to depend on a tagged release (the release workflow) to obtain a firmware bundle.

> Currently, the firmware bundle is built from source each time the Hardware Long workflow is triggered. It needs to be changed to firmware already built as an artifact so other tests can run on top of it.

Typically what CI does is build source code from a trigger and run tests though, right? The point of the `hardware-long` workflow is to detect firmware regressions against a (presumed stable) metal software stack.

Downloading a previously built artifact for firmware CI is a bit of an anti-pattern here, I think, no?

Can I close this issue?

dimitri-tenstorrent:
@cfriedt I was using Release workflow just as an example. I did not imply that we should rely on published and released bundles. 

tt-metal uses the pattern of building assets be it wheels or Cpp libs and then testing these assets in different workflows. I wanted to use the same pattern for firmware as well.

If you disagree, would you like to sync up over a call to understand the reservations a little better?

cfriedt:
[...]

PR #163, opened 2025-04-28
Build firmware as an artifact for Hardware Long workflow

kkabilarTT:
Referencing issue: https://github.com/tenstorrent/tt-zephyr-platforms/issues/158

Currently, the firmware bundle is built from source each time the Hardware Long workflow is triggered. It needs to be changed to firmware already built as an artifact so other tests can run on top of it.


.github/workflows/hardware-long.yml
```diff
@@ -109,12 +115,24 @@ jobs:
...
+      - name: Select existing firmware bundle
+        run: |
+          case "${{ matrix.config.board }}" in
+            p100) FW_PATH=firmware/p100.fwbundle;;
+            p100a) FW_PATH=firmware/p100a.fwbundle;;
+            p150a) FW_PATH=firmware/p150a.fwbundle;;
+            *) echo "Unknown board type"; exit 1;;
+          esac
+          echo "FW_PATH=$FW_PATH" >> $GITHUB_ENV
+
```
cfriedt, line 127: I think this can probably be removed

.github/workflows/hardware-long.yml
```diff
@@ -109,12 +115,24 @@ jobs:
...
       - name: Flash Firmware
         working-directory: zephyr
         run: |
           # This needs to be added to the github runner
           export PATH=$PATH:/opt/SEGGER/JLink/
 
+          cp $FW_PATH /opt/tenstorrent/fw/stable/
```
cfriedt, line 134: We definitely should not overwrite firmware generated in this PR what is considered stable firmware for CI recovery purposes.

I would have to double-check, but I think the target path is only mounted read-only.

cfriedt: To flash firmware, all that should be required is `tt-flash --fw-tar fw_pack*.fwbundle --force`

cfriedt (changes requested):
This entire workflow should probably be changed if we decide to switch to downloading a prebuilt fw artifact.

The prebuilt fw artifact that should be used is `combined-fwbundle`.

When redesigning this workflow, technically speaking, not even the source code for this repository is strictly necessary when using prebuilt firmware bundles. However, any recovery mechanisms in place would still need to be used, at least until SMC firmware is "unbrickable" (i.e. has `mcuboot` to orchestrate firmware updates).

When FW is "unbrickable", even bad firmware installed with `tt-flash` will not prevent the card from re-enumerating.

Please coordinate with @danieldegrasse .

cfriedt:
@kkabilarTT - closing since #136 was merged and I believe it addresses this concern.

https://github.com/tenstorrent/tt-zephyr-platforms/blob/c9892c584a92b14eff283c34b6e0b09ca980f85e/.github/workflows/hardware-long.yml#L41
https://github.com/tenstorrent/tt-zephyr-platforms/blob/c9892c584a92b14eff283c34b6e0b09ca980f85e/.github/workflows/hardware-long.yml#L50

Please reopen if you feel that isn't the case.

dimitri-tenstorrent:
@cfriedt There is still work to connect the Metal tests to bundles built from source.
Likely a combination of the build workflow and hardware long workflow. I'd like to re-open.

PR #164, opened 2025-04-28
ci: Add tt-smi -r smoke test

alexanderlayTT:
Add a test to hardware-smoke that performs a tt-smi reset and checks for successful reset and chip access

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -5,6 +5,8 @@
 
 import logging
 import os
+import subprocess
+
 import pyluwen
 import pytest
 import re
@@ -162,3 +164,25 @@ def test_fw_bundle_version(arc_chip):
         telemetry.fw_bundle_version == exp_bundle_version
     ), f"Firmware bundle version mismatch: {telemetry.fw_bundle_version:#010x} != {exp_bundle_version:#010x}"
     logger.info(f"FW bundle version: {telemetry.fw_bundle_version:#010x}")
+
+
+@pytest.mark.flash
+def test_smi_reset():
+    """
+    Checks that tt-smi resets are working successfully
+    """
+    smi_reset_cmd = "tt-smi -r"
+    total_tries = 10
+    fail_count = 0
+    for i in range(total_tries):
+        logger.info(f"Iteration {i}:")
+        smi_reset_result = subprocess.run(
+            smi_reset_cmd.split(), capture_output=True, check=False
+        ).returncode
+        logger.info(f"'tt-smi -r' returncode:{smi_reset_result}")
+
+        if smi_reset_result != 0:
+            fail_count += 1
+
+    logger.info(f"'tt-smi -r' failed {fail_count}/{total_tries} times.")
+    assert fail_count == 0, "'tt-smi -r' failed a non-zero number of times."


.github/workflows/hardware-smoke.yml
```diff
@@ -236,3 +236,23 @@ jobs:
...
+            --device-serial-pty rtt \
+            --outdir twister-e2e-reset
+
+      - name: Upload E2E Reset Results
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: E2E Reset results (${{ matrix.config.board }})
+          path: |
+            zephyr/twister-e2e-reset/**/update.fwbundle
```
cfriedt, line 258: I think this can be removed. We probably don't need to upload more firmware bundles.

```suggestion
      - name: Upload E2E Reset Results
        if: ${{ always() }}
        uses: actions/upload-artifact@v4
        with:
          name: E2E Reset results (${{ matrix.config.board }})
          path: |
            zephyr/twister-e2e-reset/**/update.fwbundle
```

cfriedt, line 258: I think this can probably be removed - we don't need to upload additional firmware bundles.

app/smc/pytest/e2e_smoke.py
```diff
@@ -188,3 +189,25 @@ def test_fw_bundle_version(arc_chip):
...
+        smi_reset_result = subprocess.run(
+            smi_reset_cmd.split(), capture_output=True, check=False
+        ).returncode
+        logger.info(f"'tt-smi -r' returncode:{smi_reset_result}")
+
+        if smi_reset_result != 0:
+            fail_count += 1
+
+    logger.info(f"'tt-smi -r' failed {fail_count}/{total_tries} times.")
+    assert fail_count <= 1, "'tt-smi -r' fail count greater than expected threshold"
```
cfriedt, line 213: We should probably require that 10 resets are successful. For longer duration tests where we reset e.g. 100-300 times, I think we can probably settle for 95% or better success rate.
```suggestion
    assert fail_count == 0, "'tt-smi -r' fail count greater than expected threshold"
```

app/smc/sample.yaml
```diff
@@ -16,7 +16,7 @@ tests:
         - pytest/e2e_smoke.py
       pytest_args:
         - "--dut-scope=session"
-        - "-m not flash"
+        - "-m not flash and not reset"
```
cfriedt, line 19: Is this change needed?

alexanderlayTT: I will remove this - initially I thought the intention was to only run the reset in a separate test

alexanderlayTT:
@cfriedt I've addressed the changes - thank you!

For the `hardware-long` workflow, is there a specific number of resets you are thinking of? I'll aim to implement this today

app/smc/pytest/e2e_smoke.py
```diff
@@ -188,3 +189,25 @@ def test_fw_bundle_version(arc_chip):
         telemetry.fw_bundle_version == exp_bundle_version
     ), f"Firmware bundle version mismatch: {telemetry.fw_bundle_version:#010x} != {exp_bundle_version:#010x}"
     logger.info(f"FW bundle version: {telemetry.fw_bundle_version:#010x}")
+
+
+@pytest.mark.reset
+def test_reset_accessible(arc_chip):
```
danieldegrasse, line 195: I like this for `tt-smi`, but I think we should also add a "BMC reset" test. What I mean by this is we should test an "unclean" reset, where the ARC core is reset while running. We can do this by resetting the BMC via SWD, and have a script to do it here: https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/scripts/bmc-reset.py. If we call this with the `--wait` parameter, the script should wait for SMC boot (and if it exits with an error code, we know the SMC hung)

I ask for this because I think most of our SMC hangs aren't during a tt-smi reset, but during an unclean reset (since these only happen frequently during CI)

cfriedt (dismissed):

app/smc/pytest/e2e_smoke.py
alexanderlayTT: Sorry I missed the above reply and pushed a change to add this.  If we would like this to happen in a separate PR I can drop the relevant commit

alexanderlayTT:
I believe the reason why I'm facing issues with implementing a BMC reset test is that on CI, the device will appear on `/dev/tenstorrent/0` after `rescan_pcie()`, but no device with vendor ID `0x1e52` exists in `/sys/bus/pci/devices/`, which causes `pyluwen.detect_chips()` to panic. Whereas when I run the script locally everything functions as expected.

I've dropped the commits associated with implementing the bmc-reset test and can continue to try to debug this on a separate branch, so that we can merge in the tt-smi test in the meantime.

cfriedt:
@danieldegrasse - do you want to take a look?

app/smc/sample.yaml
```diff
@@ -26,6 +26,17 @@ tests:
         - pytest/e2e_smoke.py
       pytest_args:
         - "--dut-scope=session"
+        - "-m not reset"
```
danieldegrasse, line 29: Why not just run this as part of the `smoke-flash` test? Also, don't we need to skip this test for the `e2e-smoke` test since it runs a BMC reset? Otherwise we'll be testing the firmware in the eeprom, not the firmware running from CSM.

alexanderlayTT: Initially, I thought the intent was to have this run in a separate test, but I will make the change to instead run this as part of the `smoke-flash` test. I will also add `not reset` into the args for the`e2e-smoke` test.

danieldegrasse: I think `not reset` is still missing from the `e2e-smoke` test in latest push- otherwise things look good

danieldegrasse: Missed this, my bad!

danieldegrasse (approved):

PR #168, opened 2025-04-29
lib: tenstorrent: improve I2C recovery function

danieldegrasse:
Add the following improvements to the I2C recovery function used when the I2C bus sticks:
- forcibly hold SCL low for 100ms, to induce a reset for PMBUS/SMBUS devices

This PR also adds a `reset-hammer` test, which runs as part of the long hardware testsuite. This test takes roughly 75 minutes to complete, and resets the SMC 1000 times, each time verifying there is no I2C hang. Each reset is performed twice in quick succession, because in my testing this had the best chance of reproducing I2C issues (presumably because regulator init occurs during early boot)

A link to a run of this test within the PR can be seen here, before I moved it to the nightly tests: [tenstorrent/tt-zephyr-platforms/actions/runs/14767700634/job/41462250566?pr=168](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14767700634/job/41462250566?pr=168)

### DIFF
--- a/lib/tenstorrent/bh_arc/dw_apb_i2c.c
+++ b/lib/tenstorrent/bh_arc/dw_apb_i2c.c
@@ -226,6 +226,14 @@ static void I2CRecoverBus(uint32_t id)
 	/* Set both pads to output low */
 	WriteReg(GetI2CPadDataAddr(id), 0x0);
 	/*
+	 * First, manually hold SCL low for 150 ms. Per the SMBUS spec,
+	 * we should only need to hold the line low for 25 ms, but that does
+	 * not work reliably and this does...
+	 */
+	i2c_cntl ^= RESET_UNIT_I2C_PAD_CTRL_TRIEN_SCL_MASK;
+	WriteReg(GetI2CPadCntlAddr(id), i2c_cntl);
+	Wait(150 * WAIT_1MS);
+	/*
 	 * Bitbang I2C reset to unstick bus. Hold SDA low, toggle SCL 32 times to create 16
 	 * clock cycles. Note we toggle the TRIEN bit, as when TRIEN is
 	 * set the bus will be released and external pullups will


danieldegrasse:
Moving to draft, this PR still fails some longer test runs

danieldegrasse:
Test I am running:
```
# Checkout to main branch, currently using SHA c9892c584a92b1
$ git checkout main
# Repeatedly reset the SMC twice in row (to try to reset the SMC while regulator init is happening), until tt-smi reports the ARC core did not init correctly. Let the flash run a few times after this to confirm the existing I2C Recovery function doesn't fix it.
$ for i in in {1..1000}; do echo "####### $i ###########"; west flash -d build/smc && west flash -d build/smc && sudo ./tt-zephyr-platforms/scripts/rescan-pcie.sh && tt-smi -s; done
# Checkout this branch
$ git checkout fix/i2c-recovery
# Run the same test, see if I2C recovers
$ for i in in {1..1000}; do echo "####### $i ###########"; west flash -d build/smc && west flash -d build/smc && sudo ./tt-zephyr-platforms/scripts/rescan-pcie.sh && tt-smi -s; done
```

Thus far I have run the test once locally and confirmed that the branch does recover the I2C hang.

danieldegrasse:
I've moved the `reset-hammer-test` to the long hardware tests (considering it takes an hour to execute), but you can see a run here where it passes: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14767700634/job/41462250566?pr=168

@cfriedt do you think we should make this test a nightly or weekly thing? The runtime is ~1h 15m looking at the logs, which is *quite* significant. We could move this to weekly once we are sure the I2C gremlins are gone maybe?

scripts/run-reset-hammer.sh
```diff
@@ -0,0 +1,37 @@
...
+	echo "############ Iteration $i ##################"
+	# Flash twice, this gives us the best chance of resetting the SMC while
+	# I2C init is ongoing
+	west flash -d build-reset-test
+	west flash -d build-reset-test
+	# Rescan PCI
+	sudo $TT_Z_P/scripts/rescan-pcie.sh
+	# Test with tt-smi
+	tt-smi -s
+done
```
cfriedt, line 37: It would be ideal if we could actually count and report how many times the process failed vs succeeded (even if the expectation is that it would always succeed).

Also, be careful about running with `set -e`. If any command has a non-zero exit status, then the script will die prematurely.

It would also be good to run the script through `shellcheck`. It will point out bashisms, so there are differences between running with `#!/bin/sh` vs `#!/bin/bash`.

cfriedt:
> @cfriedt do you think we should make this test a nightly or weekly thing? The runtime is ~1h 15m looking at the logs, which is _quite_ significant. We could move this to weekly once we are sure the I2C gremlins are gone maybe?

Yeah, that makes sense.

Can you update the PR description with before / after results so that it shows a numerical improvement in failure rate?

danieldegrasse:
> > @cfriedt do you think we should make this test a nightly or weekly thing? The runtime is ~1h 15m looking at the logs, which is _quite_ significant. We could move this to weekly once we are sure the I2C gremlins are gone maybe?
> 
> Yeah, that makes sense.
> 
> Can you update the PR description with before / after results so that it shows a numerical improvement in failure rate?

Test results from failing example are here: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14781473543/job/41511859046?pr=178. You can see that both cards *eventually* lock up, and they do not recover

danieldegrasse:
> Would it be worth it to also create a shorter test with e.g. several resets of a FW that just constantly polls a register from i2c?

Like a test that is designed to reproduce the error using a custom firmware, so that we maximize testing of the I2C recovery function? We definitely could, yeah, and I think it would be valuable- but I think it should coexist with this. Frankly even this test is creating somewhat arbitrary conditions, but it just happens those are similar conditions to what we create in CI.

IMO, the crash this test reproduces probably isn't super likely in the wild- but it is somewhat frequent in CI.

danieldegrasse:
https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14787606270/job/41518745613?pr=168

Here is a run with the I2C recovery fix. Without this fix, the test failed consistently at iteration 230 for the P100, and iteration 362 for the P100a. The test timed out after ~750 iterations.

With the fix, we see 5 failures out of 1000 for the P100a, and 4 out of 1000 for the P100. We recover from these failures each time.

Now for the big question- do we mark this test as failing in that case? Personally, I think yes- I'll increase the duration we hold the SCL line low for in the next push, and see if that gets us 100% passing rates

cfriedt:
@danieldegrasse - that's great. I love to see these kinds of numbers 

So to summarize based on success rate:

**Before**
p100: 30.666%
p100a: 48.266%

**After**
p100: 99.500%
p100a: 99.600%

danieldegrasse:
> @danieldegrasse - that's great. I love to see these kinds of numbers. Now to just make CI green again 😅
> 
> So to summarize based on success rate:
> 
> **Before** p100: 30.666% p100a: 48.266%
> 
> **After** p100: 99.500% p100a: 99.600%

I'll split the hammer test into another PR, and we can keep working on getting it to 100% before merging

cfriedt:
> I'll split the hammer test into another PR, and we can keep working on getting it to 100% before merging

@danieldegrasse - Hmm...

The idea is that we can use it to check for regressions. Getting to 100% can be a long-tail process. It could take months. 

What we want to be OK with is that it works e.g. 95% of the time or greater and that's a bar we can choose to increase as-needed.

So definitely keep it here - it's exactly the test commit (and kind of evidence) we want paired with the code commit that greatly improves things (not necessarily that it's fixed perfectly).

The passing criteria should be whether it works e.g. 95% of the time or greater.

If you want to make another code commit later that raises that bar to 100% and another commit that raises the bar, by all means, make that in another PR later.

cfriedt (approved):

cfriedt (approved):

PR #169, opened 2025-04-29
ci: build-fw: enable virtual uart by default for published binaries

cfriedt:
This enables the PCIe Virtual UART by default in the build-fw workflow only so that users can use `tt-console`.

This is a temporary solution until #72 is able to pass tests in CI.


danieldegrasse (approved):
Fine with me- we should target turning on vuart by default for the next release though, since these firmware bundles will be used to build the release, right? Or is the firmware rebuilt in `build_combined_fwbundle.sh`?

PR #170, opened 2025-04-30
hnaik/therm trip board fault

hnaikTT:
This PR asserts the board fault LED if any chips enter the therm trip condition


palexsonTT:
Would you be able to rebase the first commit onto the main branch and put your code in the therm trip handler part of the main thread (in `app/bmc/src/main.c`)?

TTDRosen:
We probably need some condition to remove the board fault if/when the therm trip is cleared. Is `chip->data.therm_trip_triggered` set back to false anywhere?

hnaikTT:
Closing this PR as it was merging from a fork into the main branch.

palexsonTT:
@TTDRosen In that case do we also need to detect and handle when the therm trip pin falls?

TTDRosen:
I was considering that the pin would probably fall on the reset but I figure we'd want to somehow confirm that the chip wasn't immediately going to therm trip again before turning off the fault

hnaikTT:
Something on the lines of an exit latency/ hysterisys?

PR #174, opened 2025-05-01
Add workflow_dispatch for build-fw.yml

kkabilarTT:
Add workflow_dispatch for build-fw.yml

### DIFF
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -5,6 +5,8 @@ on:
     secrets:
       SIGNATURE_KEY:
         required: true
+  workflow_dispatch:
+  
   push:
     branches:
       - main
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -6,7 +6,7 @@ on:
       SIGNATURE_KEY:
         required: true
   workflow_dispatch:
-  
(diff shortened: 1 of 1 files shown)


cfriedt:
Given our current commit rate of multiple times per day on `main`, and also I think that for all branches, our artifacts are retained for 2 weeks by default, I don't see the point of enabling on-demand artifact generation.

FWIU, the following should work as long as you are signed-in to github, but maybe I'm mistaken:

* bookmark the URL below 

https://github.com/tenstorrent/tt-zephyr-platforms/actions/workflows/build-fw.yml?query=branch%3Amain

* and then follow the link / PR title (e.g. "Fix metal reference")
* follow the link that says build-combined-fwbundle
Screenshot 2025-05-01 at 12 14 48 PM
* expand where it says "Upload Combined Firmware Bundle Assets" 
Screenshot 2025-05-01 at 12 18 59 PM
* Download the firmware bundle artifact
* https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14777637991/artifacts/3044949010

As far as I know, that's effectively the same thing as on-demand workflow dispatch but makes less work for our CI.

kkabilarTT:
Thanks for the solution - I'll close this PR

PR #175, opened 2025-05-01
Hnaik/therm trip flash led

hnaikTT:
Trigger the LED to flash mode after thermal trip interrupt is asserted


danieldegrasse (approved):
Looks fine to me- just FYI, jobs might take a while to start right now because I tied up the CI runners on a long test: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14781473543/job/41511859046?pr=178

cfriedt (approved):

PR #176, opened 2025-05-01
bmc_debug: disable reset on west attach

nxuTT:
Disable reset on west attach


cfriedt:
I think this was just merged in #180 - closing as duplicate.

PR #178, opened 2025-05-01
lib: tenstorrent: continued elimination of I2C gremlins

danieldegrasse:
Add a reset hammer test, intended to be run nightly. This test will reset the SMC repeatedly, and verify that I2C hangs do not occur when the SMC crashes (or is reset) during early init.

This PR also includes additional I2C recovery fixes, and should not be merged until the reset hammer test is passing at 1000 iterations on P100/P100a/P150a

### DIFF
--- a/scripts/run-reset-hammer.sh
+++ b/scripts/run-reset-hammer.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+# Copyright (c) 2025 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+BOARD=$1
+COUNT=$2
+
+if [ $# -ne 2 ]; then
+	echo "Incorrect number of arguments, expected board name and test count"
+	exit 1
+fi
+
+# Repeatedly resets the SMC. Intended to be run within CI.
+# Accepts two arguments:
+# 1. board name to use
+# 2. the number of times to run the reset test
+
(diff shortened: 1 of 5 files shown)


danieldegrasse:
Moving this PR to draft- the only intention here was to demonstrate that the SMC hammer workflow would fail on main, but pass with the I2C recovery fix

danieldegrasse:
CI run where reset test passes 1000 times on all cards: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14804605831/job/41570600868?pr=178

scripts/run-reset-hammer.sh
```diff
@@ -0,0 +1,63 @@
...
+	echo "Reset test ended: $pass_cnt passed, $fail_cnt failed"
+
+	if [ $fail_cnt -ne 0 ]; then
+		echo "Reset test failed with $fail_cnt errors"
+		exit 1
+	else
+		echo "Reset test passed successfully"
+		exit 0
+	fi
+}
```
cfriedt, line 36: It would be good to add a follow-up PR that reported a success rate in percent.

E.g.

```shell
success_pct="$(echo "$pass_count * 100 /  ($pass_count + $fail_count)")"

echo "Success Rate: $success_pct %"
```

cfriedt (approved):
LGTM

PR #180, opened 2025-05-01
remove reset behaviour for west attach

nxuTT:
- Remove MCU behaviour for west attach
- fixing CI with reset removal.

### DIFF
--- a/app/bmc/prj.conf
+++ b/app/bmc/prj.conf
@@ -40,6 +40,10 @@ CONFIG_TT_BOOT_BANNER=y
 CONFIG_TT_BOOT_BANNER_GIT_VERSION=y
 CONFIG_TT_BOOT_BANNER_SDK_VERSION=y
 
+# Raise RTT buffer size, so entire boot log is captured even if RTT isn't
+# attached
+CONFIG_SEGGER_RTT_BUFFER_SIZE_UP=2048
+
 # k_event APIs
 CONFIG_EVENTS=y
 CONFIG_TT_BM_EVENT=y
--- a/boards/tenstorrent/tt_blackhole/support/tt_blackhole_bmc.cfg
+++ b/boards/tenstorrent/tt_blackhole/support/tt_blackhole_bmc.cfg
@@ -5,14 +5,9 @@ transport select hla_swd
 source [find target/stm32g0x.cfg]
 # No reset signal is attached to the STM32 debug header
 reset_config none
+gdb_breakpoint_override hard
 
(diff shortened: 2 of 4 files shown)


danieldegrasse:
I can't approve this code without the appearance of bias- @cfriedt can you take a look?

cfriedt:
> we had too small of an RTT up buffer. This meant even when we did attach, the RTT subsystem had usually dropped the BMFW version print we are looking for in the output. I just increased the size to fix that

This is one of those cat-and-mouse type changes. It's a kind of necessary evil. Unfortunately, RTT kind of sucks for a lot of reasons. At some point, we need to forget about relying on RTT.

app/bmc/prj.conf
```diff
@@ -40,6 +40,10 @@ CONFIG_TT_BOOT_BANNER=y
 CONFIG_TT_BOOT_BANNER_GIT_VERSION=y
 CONFIG_TT_BOOT_BANNER_SDK_VERSION=y
 
+# Raise RTT buffer size, so entire boot log is captured even if RTT isn't
+# attached
+CONFIG_SEGGER_RTT_BUFFER_SIZE_UP=2048
```
cfriedt, line 45: Yeah... I hate wasting SRAM like this, but it's kind of a necessary evil. RTT unfortunately kind of sucks.

Next gen, let's just use a UART.

boards/tenstorrent/tt_blackhole/support/tt_blackhole_bmc.cfg
```diff
@@ -5,14 +5,9 @@ transport select hla_swd
 source [find target/stm32g0x.cfg]
 # No reset signal is attached to the STM32 debug header
 reset_config none
+gdb_breakpoint_override hard
 
-$_TARGETNAME configure -event gdb-attach {
-        echo "Debugger attaching: halting execution"
-        reset halt
```
cfriedt, line 11: This makes sense to me - normally, I wouldn't expect a gdb-attach to reset the target, it should just attach to it in whatever state it's in.

And I think the default detach behaviour is identical anyway.

No real need to echo a message in either case.

cfriedt (approved):

app/bmc/prj.conf
danieldegrasse: Very much agree- I also wonder why RTT doesn't just loop? IE why can't we overwrite the start of the buffer when RTT overflows

cfriedt: It does, but it's not as instantaneous as hardware

PR #181, opened 2025-05-01
ci: use generic tags for blackhole, enable P150

danieldegrasse:
Use generic tags for blackhole runners, as we now have two P100a runners online. Enable P150 runner in CI.

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -18,10 +18,13 @@ jobs:
         config:
           - board: p100
             runs-on:
-              - yyz-zephyr-lab-p100
+              - p100-jtag
           - board: p100a
             runs-on:
-              - yyz-zephyr-lab-p100a
+              - p100a-jtag
+          - board: p150a
+            runs-on:
+              - p150a-jtag
     runs-on: ${{ matrix.config.runs-on }}
     container:
       image: ghcr.io/tenstorrent/tt-metal/tt-metalium/ubuntu-22.04-dev-amd64:bd47135f35f7b26851eb09066076fb3e6074c54f # Change this to a tag when metal has release versions for this image.
--- a/.github/workflows/prepare-zephyr/action.yml
+++ b/.github/workflows/prepare-zephyr/action.yml
@@ -64,6 +64,11 @@ runs:
         app-path: ${{ inputs.app-path }}
         toolchains: arm-zephyr-eabi:riscv64-zephyr-elf:x86_64-zephyr-elf:arc-zephyr-elf
 
+    - name: Install app python dependencies
+      shell: bash
+      run: |
+        pip install -r ${{ inputs.app-path }}/scripts/requirements.txt
+
     - name: Verify binary blobs
       shell: bash
       run: |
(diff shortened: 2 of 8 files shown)


.github/workflows/prepare-zephyr/action.yml
```diff
@@ -64,6 +64,11 @@ runs:
         app-path: ${{ inputs.app-path }}
         toolchains: arm-zephyr-eabi:riscv64-zephyr-elf:x86_64-zephyr-elf:arc-zephyr-elf
 
+    - name: Install app python dependencies
+      shell: bash
+      run: |
+        pip install -r ${{ inputs.app-path }}/scripts/requirements.txt
+
```
cfriedt, line 71: This shouldn't be necessary since pip requirements are install by the upstream setup action.

`west packages pip --install` IIRC

Can you try without?

cfriedt (commented):
Looks mostly good. Just some minor comments.

Do we need to reregister GitHub runners or is there a safe backend way to add or change the label?

cfriedt (commented):
Also mega bonus points if we can somehow generate the list of boards for CI from a JSON or YAML file. I hate maintaining this in multiple places. 1 source of truth would be nice.

danieldegrasse:
> Do we need to reregister GitHub runners or is there a safe backend way to add or change the label?

Labels have been changed- if you look at the CI run for this PR it is being split between the P100a in aus and the one in yyz :)

> Also mega bonus points if we can somehow generate the list of boards for CI from a JSON or YAML file. I hate maintaining this in multiple places. 1 source of truth would be nice.

Within github actions? We might be able to do this, do you mean tracking the inventory of boards within the repo itself?

.github/workflows/prepare-zephyr/action.yml
danieldegrasse: Yes, pip requirements for building *zephyr* are- we have a separate requirements file that includes `tt-flash` and `tt-smi`. I did try this without, here is the run: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14784332664/job/41510027200

cfriedt:
> Within github actions? We might be able to do this, do you mean tracking the inventory of boards within the repo itself?

Maybe one yaml to list targets to build in CI. Ideally the same yaml could be used to determine which self-hosted runners were available (per org / repo).

I wish it were possible to add a rule to only run HW CI if certain runners were available, but that query requires additional admin privileges by the runner, which might introduce security concerns.

Actually, come to think of it, do we want to maybe use a syseng prefix to p100, p100a, etc to differentiate those from the rest of the runners available to the org?

.github/workflows/hardware-long.yml
```diff
@@ -18,10 +18,13 @@ jobs:
         config:
           - board: p100
             runs-on:
-              - yyz-zephyr-lab-p100
```
cfriedt, line 21: On second glance, and given that our runners are available to everyone in the org, do we maybe want to change the labels to be more specific so that it's clear they are used for zephyr firmware?

Eventually, we should be able to go fully generic anywhere on the cloud, but I think that it would require the "unbrickable" fw boot flow for CMFW.

Maybe until then we should just keep some kind of prefix.

danieldegrasse:
PR has been updated with the new `<board>-jtag` labels for runners

PR #185, opened 2025-05-02
Update version to 80.18.1.0

afongTT:
Release version 80.18.1.0.

Make a patch release to fix the reported bundle versions of WH and GS.


cfriedt (approved):
I'll approve / merge / tag

PR #186, opened 2025-05-02
tests: drivers: flash: performance: update test to run on MT25

danieldegrasse:
MT25 quad flash takes a bit longer to read because it uses quad
mode (~7 ms), so add a new testcase definition for the devices
using this flash driver.

Update CI invocations to properly run out of tree tests (thanks @cfriedt for catching this!)

This PR also includes a patch to west's telnet client to resolve a bug with decoding RTT output, which was causing failures on CI systems since they do not have netcat installed in the docker image.


cfriedt (approved):

danieldegrasse:
Who knows how I clicked the close button

cfriedt (approved):

PR #187, opened 2025-05-05
Test PR: Use a container in the e2e PR flow

dimitri-tenstorrent:
Use an existing container from Hardware Long.
The idea is to switch to Metal Upstream container
but one step at a time.


.github/workflows/hardware-smoke.yml
```diff
@@ -158,108 +158,44 @@ jobs:
...
-            p100) BMC_BOARD=tt_blackhole@p100/tt_blackhole/bmc;;
-            p100a) BMC_BOARD=tt_blackhole@p100a/tt_blackhole/bmc;;
-            p150a) BMC_BOARD=tt_blackhole@p150a/tt_blackhole/bmc;;
-            p300) BMC_BOARD=p300/tt_blackhole/bmc;;
-            *) echo "Unknown board: ${{ matrix.config.board }}"; exit 1;;
-          esac
-          echo "SMC_BOARD=$SMC_BOARD" >> "$GITHUB_ENV"
-          echo "BMC_BOARD=$BMC_BOARD" >> "$GITHUB_ENV"
-
-      - name: run-e2e-tests
```
danieldegrasse, line 200: We need to keep these e2e tests running somewhere, and they should run before we execute metal (as a set of basic sanity tests). Personally I think the best option is to use a separate job for metal tests, that depends on the completion of the E2E tests before running

.github/workflows/hardware-smoke.yml
```diff
@@ -158,108 +158,44 @@ jobs:
...
+      - name: Flash the firmware
+        run: |
+          sudo chmod -R a+rwX $HOME/.cargo $HOME/.cache
+          python -m venv .env
+          source .env/bin/activate
+          pip install git+https://github.com/tenstorrent/tt-flash.git
+          tt-flash --fw-tar fw_pack*.fwbundle --force
+      - name: Run Container Test
+        run: |
+          sh -c "$TEST_COMMAND"
```
danieldegrasse, line 196: Just wondering, is this command set as part of the environment of the container?

dimitri-tenstorrent: Yes.

danieldegrasse:
Changes look ok to me now- since this is using the firmware bundle from main, should we make this part of the nightly tests for now? I think that when we run metal per-PR it should use the firmware bundle from that PR

dimitri-tenstorrent:
> Changes look ok to me now- since this is using the firmware bundle from main, should we make this part of the nightly tests for now? I think that when we run metal per-PR it should use the firmware bundle from that PR
> 
> I know that for the CI runs we need this to be part of the per-PR workflow, just wondering what the final change will look like

Awesome! Here is PR #199 and Kannika and I will work on https://github.com/tenstorrent/tt-zephyr-platforms/pull/198 to merge it in before this one.

.github/workflows/hardware-smoke.yml
```diff
@@ -263,3 +263,61 @@ jobs:
...
+      ARCH_NAME: blackhole
+    defaults:
+      run:
+        shell: bash
+        working-directory: /home/user/tt-metal/
+    steps:
+      - name: Install Unzip
+        run: |
+          sudo apt update
+          sudo apt install -y unzip protobuf-compiler
```
cfriedt, line 299: Could these dependencies be added to the container image? That would probably simplify some things.

dimitri-tenstorrent:
Closing as #227 already incorporated the Metal tests into Smoke tests

PR #193, opened 2025-05-05
ci: use shortened Zephyr setup when SDK is already present

danieldegrasse:
When running within a Zephyr SDK docker image, we don't need to manually setup the Zephyr SDK every time. Add a basic setup step that runs when we already have an SDK installed. This should speed up CI execution time

### DIFF
--- a/.github/workflows/prepare-zephyr/action.yml
+++ b/.github/workflows/prepare-zephyr/action.yml
@@ -56,13 +56,40 @@ runs:
           rm -rf zephyr-sdk
         fi
 
-    # TODO: potentially improve caching strategy
-    # Note: this step performs "west init", "west update", and installs zephyr python dependencies
-    - name: Set up Zephyr project
+    - name: Check if Zephyr SDK is installed
+      shell: bash
+      id: check-sdk
+      run: |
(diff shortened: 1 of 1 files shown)


danieldegrasse:
This change saves us about a minute per CI run on hardware, so ~2 minutes faster for CI results (since hardware smoke and hardware e2e will be sped up)

Run without this optimization (3m 20s): https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14862278154/job/41730011692?pr=197
Run with this optimization (2m 18s): https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14862404306/job/41730448108?pr=193

danieldegrasse:
> Faster is better, so not going to block it.
> 
> However, this is a kind of bespoke hack to speed up our CI that we'll need to carry forward.
> 
> I do like the simplicity of what we had before. A PR to fix the upstream `action-zephyr-setup` would be a really good upgrade path. Maybe even passing in the existence of the SDK would be an option to consider (I thought you had a PR for that already?).

I do have a PR, it isn't moving forwards :/ The alternative here would be to use a fork of the action-zephyr-setup until my change merges

PR #194, opened 2025-05-05
ci: make sure docker container uses HTTPs for HTTP

danieldegrasse:
Self hosted docker containers need https urls for http. Patch the urls when installing curl, as this step is only running on docker containers anyways


cfriedt (approved):

danieldegrasse:
Closing this out, probably not needed and cleaning up old PRs...

PR #198, opened 2025-05-06
Update hardware-long.yml to build combined fwbundle and use the uploaded artifact

kkabilarTT:
Build combined fwbundle and use the uploaded artifact in hardware-long.yml

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -14,6 +14,7 @@ jobs:
   build-fw-artifact:
     uses: ./.github/workflows/build-fw.yml
     secrets: inherit
+
   hardware-metal-test:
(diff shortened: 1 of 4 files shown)


.github/workflows/hardware-long.yml
```diff
@@ -11,7 +11,12 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  build-fw-artifact:
```
danieldegrasse, line 14: Generally looks good- the github annotations here should provide a hint, but you need to remove the trailing whitespace here to pass compliance checks

danieldegrasse (commented):
Can you move the metal workflow into the hardware-smoke definition for one run, so we can verify that CI passes with this change?

PR #211, opened 2025-05-07
Use artifact built from build-fw

kkabilarTT:
Workflow: built artifact

Use artifact built from build-fw.yml

### DIFF
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -5,6 +5,10 @@ on:
     secrets:
       SIGNATURE_KEY:
         required: true
+    outputs:
+      combined-fwbundle-artifact:
+        description: "The combined firmware bundle artifact name"
+        value: ${{ jobs.build-combined-fwbundle.outputs.combined-fwbundle-artifact }}
   push:
     branches:
       - main
@@ -151,6 +155,8 @@ jobs:
   build-combined-fwbundle:
     runs-on: ubuntu-24.04
     needs: build
+    outputs:
+      combined-fwbundle-artifact: ${{ steps.set_combined_fwbundle_output.outputs.combined_fwbundle_artifact }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -170,9 +176,21 @@ jobs:
           BUNDLE_TEMP_PREFIX="$PWD/tt-zephyr-platforms/firmware-" \
             ./tt-zephyr-platforms/scripts/build-combined-fwbundle.sh
 
-      - name: Upload Combined Firmware Bundle Assets
+      - name: Find the generated fwbundle
+        id: find-fwbundle
+        run: |
+          echo "Finding .fwbundle file..."
+          zip fw_pack.zip fw_pack-*.fwbundle
+          FW_BUNDLE_PATH=$(find . -name 'fw_pack.zip' | head -n 1)
+          echo "Found firmware bundle at: $FW_BUNDLE_PATH"
+          echo "fw_bundle_path=$FW_BUNDLE_PATH" >> "$GITHUB_ENV"
+
+      - name: Upload Combined Firmware Bundle Asset
         uses: actions/upload-artifact@v4
         with:
           name: combined-fwbundle
-          path: |
-            fw_pack*.fwbundle
+          path: ${{ env.fw_bundle_path }}
+
+      - name: Set combined firmware bundle output
+        id: set_combined_fwbundle_output
+        run: echo "combined_fwbundle_artifact=combined-fwbundle" >> $GITHUB_OUTPUT
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -11,7 +11,12 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  build-fw-artifact:
+    uses: ./.github/workflows/build-fw.yml
+    secrets:
+      SIGNATURE_KEY: ${{ secrets.SIGNATURE_KEY }}
   hardware-metal-test:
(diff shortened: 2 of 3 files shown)


.github/workflows/hardware-smoke.yml
```diff
@@ -19,6 +19,67 @@ concurrency:
...
+          sudo apt update
+          sudo apt install -y unzip protobuf-compiler
+      - name: Download the latest firmware bundle
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ needs.build-fw-artifact.outputs.combined-fwbundle-artifact }}
+          path: .
+      - name: Run the rescan-pcie.sh script
+        run: |
+          curl -o /tmp/rescan-pcie.sh https://raw.githubusercontent.com/tenstorrent/tt-zephyr-platforms/${{ github.ref }}/scripts/rescan-pcie.sh
```
dimitri-tenstorrent, line 66: You can use `main` here for testing for now as it is causing an issue in the tests.

The proper fix might be switching from `github.ref` to `github.sha`

.github/workflows/hardware-long.yml
```diff
@@ -44,10 +49,10 @@ jobs:
           sudo apt update
           sudo apt install -y unzip protobuf-compiler
       - name: Download the latest firmware bundle
-        # Replace this with the reliance on the previous workflow that builds the firmware bundle.
-        run: |
-          curl -L -o combined-fwbundle.zip https://nightly.link/tenstorrent/tt-zephyr-platforms/workflows/build-fw/main/combined-fwbundle.zip
-          unzip -o combined-fwbundle.zip
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ needs.build-fw-artifact.outputs.combined-fwbundle-artifact }}
```
dimitri-tenstorrent, line 54: The logs suggest that this string is empty as the log say that the name is not provided: 
    ```Found 18 artifact(s)
No input name, artifact-ids or pattern filtered specified, downloading all artifacts
An extra directory with the artifact name will be created for each download
Preparing to download the following artifacts:
```

File: .github/workflows/build-fw.yml [hunk 3]
```diff
@@ -151,6 +151,8 @@ jobs:
   build-combined-fwbundle:
     runs-on: ubuntu-24.04
     needs: build
+    outputs:
+      combined-fwbundle-artifact: ${{ steps.set-output.outputs.combined_fwbundle_artifact }}
```
dimitri-tenstorrent, line 155: There is likely a dash vs underscore mismatch between what you expect and what you send to GITHUB_OUTPUT

.github/workflows/hardware-smoke.yml
```diff
@@ -19,6 +19,67 @@ concurrency:
...
+          source /tmp/rescan-pcie.sh
+      - name: Flash the firmware
+        run: |
+          sudo chmod -R a+rwX $HOME/.cargo $HOME/.cache
+          python -m venv .env
+          source .env/bin/activate
+          pip install git+https://github.com/tenstorrent/tt-flash.git
+          tt-flash --fw-tar fw_pack*.fwbundle --force
```
dimitri-tenstorrent, line 74: We need to make sure the name of the artifact matches with what we are trying to flash

dimitri-tenstorrent (changes requested):
Super close! Let's get this over the line :)

.github/workflows/hardware-long.yml
dimitri-tenstorrent: I think you have fixed this.

.github/workflows/hardware-smoke.yml
```diff
@@ -19,6 +19,76 @@ concurrency:
...
+        run: echo "Firmware bundle artifact is ${{ needs.build-fw-artifact.outputs.combined-fwbundle-artifact }}"
+      - name: Download the latest firmware bundle
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ needs.build-fw-artifact.outputs.combined-fwbundle-artifact }}
+          path: .
+      - name: Find the downloaded zip file
+        run: |
+          echo "Listing files in current directory:"
+          ls -la
```
dimitri-tenstorrent, line 69: From the logs - it seems that the files are downloaded to a different directory - https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14917845760/job/41907547275#step:6:16

dimitri-tenstorrent: Specifically, `Starting download of artifact to: /__w/tt-zephyr-platforms/tt-zephyr-platforms`

.github/workflows/hardware-smoke.yml
```diff
@@ -19,6 +19,70 @@ concurrency:
...
+    steps:
+      - name: Install Unzip
+        run: |
+          sudo apt update
+          sudo apt install -y unzip protobuf-compiler
+      - name: Download the latest firmware bundle
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ needs.build-fw-artifact.outputs.combined-fwbundle-artifact }}
+          path: .
```
dimitri-tenstorrent, line 63: I wonder if it is better to download into ` /home/user/tt-metal/` directory since we end up using it all the time.

dimitri-tenstorrent (commented):
This is starting to look great! 

We likely need to split this into a few PRs for merging. This way every step will be well-tested as per Daniel's instructions. 

1. Build-fw workflow
2. Integration of Build-fw and Hardware Smoke: you can use an existing PR https://github.com/tenstorrent/tt-zephyr-platforms/pull/187
3. Integration with Hardware Long

If you want, you can combine 1&2 and then I will close PR 187.

PR #212, opened 2025-05-09
smbus messaging: notify BMC FW when SMC FW is ready to receive msg

kikkiakki:
BMC FW attempts to send board power limit before SMC FW is guaranteed to be initialized. This results in unintentional throttling down to the default board power limit of 150W.

Add a cm2bm message to notify BMC FW when it's okay to start sending data.

### DIFF
--- a/include/tenstorrent/bh_chip.h
+++ b/include/tenstorrent/bh_chip.h
@@ -43,10 +43,11 @@ struct bh_chip_data {
 	bool workaround_applied;
 
 	/*
-	 * Flag set when need to send or receive 1 time info to chip.
+	 * Flags set when need to send or receive 1 time info to chip.
 	 * Could be used for static data or config of peripherals.
 	 */
 	bool arc_just_reset;
+	bool cm_ready_for_msgs;
 
 	unsigned int bus_cancel_flag;
 
--- a/include/tenstorrent/bh_chip.h
+++ b/include/tenstorrent/bh_chip.h
@@ -43,11 +43,10 @@ struct bh_chip_data {
 	bool workaround_applied;
 
 	/*
-	 * Flags set when need to send or receive 1 time info to chip.
+	 * Flag set when need to send or receive 1 time info to chip.
 	 * Could be used for static data or config of peripherals.
 	 */
-	bool arc_just_reset;
-	bool cm_ready_for_msgs;
+	bool arc_needs_init_msg;
 
 	unsigned int bus_cancel_flag;
 
--- a/lib/tenstorrent/bh_arc/cm2dm_msg.c
+++ b/lib/tenstorrent/bh_arc/cm2dm_msg.c
@@ -115,6 +115,15 @@ void UpdateFanSpeedRequest(uint32_t fan_speed)
 	EnqueueCm2DmMsg(&msg);
 }
 
+void Dm2CmReadyRequest(void)
+{
+	/* Send a message to dmfw to indicate ready to receive messages */
+	Cm2DmMsg msg = {
+		.msg_id = kCm2DmMsgIdReady,
+	};
+	EnqueueCm2DmMsg(&msg);
+}
+
 /* Report the current message and automatically acknowledge it. */
 int32_t ResetBoardByte(uint8_t *data, uint8_t size)
 {
--- a/app/dmc/src/main.c
+++ b/app/dmc/src/main.c
@@ -352,10 +352,10 @@ int main(void)
 		 */
 		ARRAY_FOR_EACH_PTR(BH_CHIPS, chip) {
 			if (chip->data.arc_needs_init_msg) {
-				if (bh_chip_set_static_info(chip, &static_info) == 0) {
+				if (bh_chip_set_static_info(chip, &static_info) == 0 &&
+				    bh_chip_set_board_pwr_lim(chip, max_pwr) == 0) {
 					chip->data.arc_needs_init_msg = false;
 				}
-				bh_chip_set_board_pwr_lim(chip, max_pwr);
 			}
 		}
 
--- a/app/dmc/src/main.c
+++ b/app/dmc/src/main.c
@@ -120,6 +120,9 @@ void process_cm2dm_message(struct bh_chip *chip)
 				set_fan_speed((uint8_t)message.data & 0xFF);
 			}
 			break;
+		case 0x4:
+			chip->data.cm_ready_for_msgs = true;
+			break;
 		}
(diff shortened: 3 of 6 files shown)


app/bmc/src/main.c
```diff
@@ -38,6 +38,8 @@ static const struct gpio_dt_spec board_fault_led =
 	GPIO_DT_SPEC_GET_OR(DT_PATH(board_fault_led), gpios, {0});
 static const struct device *const ina228 = DEVICE_DT_GET_OR_NULL(DT_NODELABEL(ina228));
 
+static bool cm_ready_for_msgs;
```
TTDRosen, line 41: You need to reset this back to false post reset

alewycky-tenstorrent, line 41: First, this has to be per-chip.

This starts out false and becomes true after kCm2BmMsgIdReady. Does it ever need to go false again? I think any kind of BH reset event should clear it.

But is there any case where the BH/CMFW can go through a reset without BMFW knowing?

kikkiakki: Is it okay if I do it here?
```
if (bh_chip_set_static_info(chip, &static_info) == 0) {
	chip->data.arc_just_reset = false;
	chip->data.cm_ready_for_msgs = false;
}
```

alewycky-tenstorrent: Aaron confirmed in other discussion that BMFW knows all resets because it has to participate in the boot workaround. (until BHA1)

alewycky-tenstorrent: I think this leaves a surprise for the future when somebody gates sending a message on cm_ready_for_msgs without realizing that it was cleared early rather than on reset.

app/bmc/src/main.c
```diff
@@ -115,6 +115,9 @@ void process_cm2bm_message(struct bh_chip *chip)
 				set_fan_speed((uint8_t)message.data & 0xFF);
 			}
 			break;
+		case 0x4:
+			chip->data.cm_ready_for_msgs = true;
```
afongTT, line 119: @DRvader do we need a `chip->data.arc_just_reset = true` here for the case where CMFW gets reloaded via JTAG?

TTDRosen: Yep, I don't think that should have any negative side effects...

afongTT: Maybe rename it to something like `arc_needs_init_msg` since that's all it's used for.

afongTT: Actually, if we only send the init msg after prompting by CMFW, then you only need one flag.

kikkiakki: So we should only keep one of `arc_just_reset` and `cm_ready_for_msgs` and only set it to true when this message is received?

TTDRosen: Yeah, I agree we should rename arc_just_reset to arc_needs_init_msg because I don't think it is used anywhere else.

kikkiakki: Okay, added a change to address this

app/bmc/src/main.c
```diff
@@ -343,9 +346,9 @@ int main(void)
 		/* TODO(drosen): Turn this into a task which will re-arm until static data is sent
 		 */
 		ARRAY_FOR_EACH_PTR(BH_CHIPS, chip) {
-			if (chip->data.arc_just_reset) {
+			if (chip->data.arc_needs_init_msg) {
 				if (bh_chip_set_static_info(chip, &static_info) == 0) {
-					chip->data.arc_just_reset = false;
+					chip->data.arc_needs_init_msg = false;
 				}
 				bh_chip_set_board_pwr_lim(chip, max_pwr);
```
afongTT, line 353: Can you keep sending until both messages are sent successfully?

kikkiakki: done

app/bmc/src/main.c
```diff
@@ -115,6 +115,9 @@ void process_cm2bm_message(struct bh_chip *chip)
 				set_fan_speed((uint8_t)message.data & 0xFF);
 			}
 			break;
+		case 0x4:
+			chip->data.arc_needs_init_msg = true;
```
cfriedt, line 119: I realize that there is a pattern being followed here, but I'm really confused about the pattern we're using.

Why are we setting a `bool` value in global state here for the `main` thread to later pick up when this function is itself called by the `main` thread?

Wouldn't it be more efficient to simply call the following without setting any global state?

```cpp
bh_chip_set_static_info(chip, &static_info);
bh_chip_set_board_pwr_lim(chip, max_pwr);
```

include/tenstorrent/bh_chip.h
```diff
@@ -46,7 +46,7 @@ struct bh_chip_data {
 	 * Flag set when need to send or receive 1 time info to chip.
 	 * Could be used for static data or config of peripherals.
 	 */
-	bool arc_just_reset;
+	bool arc_needs_init_msg;
```
cfriedt, line 49: I think we can remove this (and the comment above it).

app/bmc/src/main.c
TTDRosen: Those functions aren't guaranteed to succeed on the first attempt so we want to keep retrying if we fail to send the first time. Ideally we would be spawning a task so it could happen entirely asynchronously but that feels like too much added complexity for something that has to pass qual the first time through.

cfriedt: Ah, I see

lib/tenstorrent/bh_arc/cm2dm_msg.c
```diff
@@ -115,6 +115,15 @@ void UpdateFanSpeedRequest(uint32_t fan_speed)
 	EnqueueCm2DmMsg(&msg);
 }
 
+void Dm2CmReadyRequest(void)
+{
+	/* Send a message to bmfw to indicate ready to receive messages */
```
afongTT, line 120: Change comment to refer to dmfw?

afongTT (approved):
Fix the "bmfw" comment if you can

lib/tenstorrent/bh_arc/cm2dm_msg.c
kikkiakki: done

afongTT (approved):

cfriedt (approved):
Is this the only change required for aiclk?

ejouretTT:
@cfriedt @tcolacoTT is this FW ready to test with for fixed AI Clock?

PR #214, opened 2025-05-09
telemetry cleanup

alewycky-tenstorrent:
Mostly internal changes, but also telemetry entries without valid data have been removed.


alewycky-tenstorrent:
Needs rethink and update.

PR #218, opened 2025-05-12
ci: fixes for build artifact uploads

danieldegrasse:
While looking at test results for #164, I realized we aren't uploading the `twister_harness.log` file where pytest results are stored.

This PR fixes that issue, as well as a few other improvements:
- enable uploading hidden files so that the `.config` file gets uploaded
- combine the e2e-flash and e2e-smoke uploads, since we are going to run some tests exclusively within the `e2e-flash` run


danieldegrasse:
Example artifact link here: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/14976737457/artifacts/3107534441.

cfriedt (approved):

PR #222, opened 2025-05-13
Save the uploaded artifact as github output

kkabilarTT:
Save artifact from build-fw.yml

Signed-off-by: Kannika Kabilar kkabilar@tenstorrent.com

### DIFF
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -5,6 +5,10 @@ on:
     secrets:
       SIGNATURE_KEY:
         required: true
+    outputs:
+      combined-fwbundle-artifact:
+        description: "The combined firmware bundle artifact name"
+        value: ${{ jobs.build-combined-fwbundle.outputs.combined-fwbundle-artifact }}
   push:
     branches:
       - main
@@ -151,6 +155,8 @@ jobs:
   build-combined-fwbundle:
     runs-on: ubuntu-24.04
     needs: build
+    outputs:
+      combined-fwbundle-artifact: ${{ steps.set_combined_fwbundle_output.outputs.combined_fwbundle_artifact }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -176,3 +182,6 @@ jobs:
           name: combined-fwbundle
           path: |
             fw_pack*.fwbundle
+
+      - id: set_combined_fwbundle_output
+        run: echo "combined_fwbundle_artifact=combined-fwbundle" >> $GITHUB_OUTPUT


dimitri-tenstorrent:
This looks good to me as the first part of the 3-part series of PRs :) @danieldegrasse @cfriedt Kannika already tested this flow in https://github.com/tenstorrent/tt-zephyr-platforms/pull/211

danieldegrasse (approved):

.github/workflows/build-fw.yml
```diff
@@ -170,9 +176,21 @@ jobs:
...
 
-      - name: Upload Combined Firmware Bundle Assets
+      - name: Find the generated fwbundle
+        id: find-fwbundle
+        run: |
+          echo "Finding .fwbundle file..."
+          zip fw_pack.zip fw_pack-*.fwbundle
+          FW_BUNDLE_PATH=$(find . -name 'fw_pack.zip' | head -n 1)
+          echo "Found firmware bundle at: $FW_BUNDLE_PATH"
+          echo "fw_bundle_path=$FW_BUNDLE_PATH" >> "$GITHUB_ENV"
```
cfriedt, line 186: Is this step necessary?

The fwbundle itself is a compressed `.tar.gz` file which will not receive any additional compression from `zip`.

Aside from that, it would seem that this step is mainly for debug verbosity? It's otherwise equivalent to `fw_pack*.fwbundle` below.

There should never be more than one distinct match / file that exists for `fw_pack*.fwbundle` for a given commit, since the filename itself is fixed [here](https://github.com/tenstorrent/tt-zephyr-platforms/blob/b22ddf868f2c89674c88bfe366518eaf62c9e5c5/scripts/build-combined-fwbundle.sh#L83) and any existing file with the same name is erased [here](https://github.com/tenstorrent/tt-zephyr-platforms/blob/b22ddf868f2c89674c88bfe366518eaf62c9e5c5/scripts/tt_boot_fs.py#L847).

Have you ever run into a situation where more than one `fw_pack*.fwbundle` exists?

.github/workflows/build-fw.yml
```diff
@@ -170,9 +176,21 @@ jobs:
...
+
+      - name: Upload Combined Firmware Bundle Asset
         uses: actions/upload-artifact@v4
         with:
           name: combined-fwbundle
-          path: |
-            fw_pack*.fwbundle
+          path: ${{ env.fw_bundle_path }}
```
cfriedt, line 192:
```suggestion
          path: |
            fw_pack*.fwbundle
```

cfriedt (changes requested):
My preference would be to be further minimize this change slightly.

.github/workflows/build-fw.yml
```diff
@@ -170,9 +176,21 @@ jobs:
...
+
+      - name: Set combined firmware bundle output
+        id: set_combined_fwbundle_output
+        run: echo "combined_fwbundle_artifact=combined-fwbundle" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
kkabilarTT: Done

.github/workflows/build-fw.yml
kkabilarTT: Ok, I've removed it

.github/workflows/build-fw.yml
kkabilarTT: Done

cfriedt (approved):

cfriedt:
LGTM - thank you @kkabilarTT and @dimitri-tenstorrent 🙏

danieldegrasse (dismissed):

danieldegrasse (approved):

PR #223, opened 2025-05-13
 app: smc: pytest: e2e_smoke: add "dirty reset" test

danieldegrasse:
Add test that resets the SMC via the DMC reset line, which is considered a "dirty reset" as the SMC and tt-kmd are not involved. Verify the SMC can boot after this reset


scripts/dmc_reset.py
```diff
@@ -275,7 +279,7 @@ def main():
                 logger.debug(f"Found {dut['platform']} with JTAG ID {dut['id']}")
                 break
 
-    ret = reset_dmc(args)
+    ret = reset_dmc(vars(args))
```
cfriedt, line 282: Would be kind of nice to just keep args here, I think. There is an alternative though (which I'll mention where it's needed).

app/smc/pytest/e2e_smoke.py
```diff
@@ -186,3 +185,37 @@ def test_smi_reset():
...
+    total_tries = 10
+    timeout = 60  # seconds to wait for SMC boot
+    fail_count = 0
+    # Use dmc-reset script as a library to reset the DMC
+    args = {
+        "config": dmc_reset.DEFAULT_DMC_CFG,
+        "debug": 0,
+        "openocd": dmc_reset.DEFAULT_OPENOCD,
+        "scripts": dmc_reset.DEFAULT_SCRIPTS_DIR,
+    }
```
cfriedt, line 206:
```suggestion
    args = lambda: None
    args.config = dmc_reset.DEFAULT_DMC_CFG
    args.debug = 0
    args.openocd = dmc_reset.DEFAULT_OPENOCD
    args.scripts = dmc_reset.DEFAULT_SCRIPTS_DIR
```

scripts/dmc_reset.py
```diff
@@ -127,18 +130,18 @@ def parse_args():
 
 def reset_dmc(args):
```
cfriedt, line 131: Would be kind of nice to keep `reset_dmc(args)`, I think, rather than rework that function to use dicts everywhere. I get that a dict is convenient to compose arguments on the fly in a test, but there is an alternative that I'll suggest in that file.

cfriedt (approved):

PR #225, opened 2025-05-13
Save and use fw artifact

kkabilarTT:
Use artifact from build-fw


dimitri-tenstorrent:
@kkabilarTT The last push is to use build fw inside of Hardware Long.

cfriedt:
@kkabilarTT - is this needed anymore since #227 and #256 were merged?

cfriedt:
I'm assuming that this is no longer needed. Please re-open if that's not the case.

PR #226, opened 2025-05-13
ci: switch back to hw-map.yml after dmc rename

cfriedt:
Change workflows to use hw-map.yml instead of hw-map-dmc.yml after renaming bmc to dmc.

Additionally:
* ci: hardware-smoke: mount /opt/tenstorrent volume once
* ci: hardware-smoke: only use openocd-rtt for arc

### DIFF
--- a/.github/workflows/hardware-smoke.yml
+++ b/.github/workflows/hardware-smoke.yml
@@ -77,7 +77,7 @@ jobs:
           ./scripts/twister -i --retry-failed 3 \
             --retry-interval 5 \
             -p $DMC_BOARD --device-testing \
-            --hardware-map /opt/tenstorrent/twister/hw-map-dmc.yml --west-flash \
+            --hardware-map /opt/tenstorrent/twister/hw-map.yml --west-flash \
             --tag smoke \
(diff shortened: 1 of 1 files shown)


.github/workflows/hardware-smoke.yml
```diff
@@ -267,4 +263,4 @@ jobs:
           echo "DMC RTT logs:"
           python3 ./scripts/dmc_rtt.py -n --openocd /opt/tenstorrent/bin/openocd-rtt
           echo "SMC RTT logs:"
-          python3 ./scripts/smc_rtt.py -n --openocd /opt/tenstorrent/bin/openocd-rtt
+          python3 ./scripts/smc_rtt.py -n
```
danieldegrasse, line 266: SMC needs to use openocd RTT, only BMC will be safe with the default executable

cfriedt: Oh sorry - I had this backwards. Good catch

cfriedt: Fixed!

danieldegrasse (approved):

PR #227, opened 2025-05-14
Integrate build-fw into hardware-smoke

kkabilarTT:
Call build-fw in hardware-smoke

Integrate build-fw into hardware-smoke

### DIFF
--- a/.github/workflows/hardware-smoke.yml
+++ b/.github/workflows/hardware-smoke.yml
@@ -275,9 +275,10 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - board: p100
-            runs-on:
-              - p100-jtag
+          # disabled until #253 is resolved
+          # - board: p100
+          #  runs-on:
+          #    - p100-jtag
           - board: p100a
             runs-on:
               - p100a-jtag
--- a/.github/workflows/hardware-smoke.yml
+++ b/.github/workflows/hardware-smoke.yml
@@ -264,3 +264,64 @@ jobs:
           python3 ./scripts/dmc_rtt.py -n
           echo "SMC RTT logs:"
           python3 ./scripts/smc_rtt.py -n --openocd /opt/tenstorrent/bin/openocd-rtt
+
+  build-fw-artifact:
+    uses: ./.github/workflows/build-fw.yml
+    secrets:
+      SIGNATURE_KEY: ${{ secrets.SIGNATURE_KEY }}
+  metal-test:
+    needs: build-fw-artifact
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - board: p100
+            runs-on:
+              - p100-jtag
+          - board: p100a
+            runs-on:
+              - p100a-jtag
+          - board: p150a
+            runs-on:
+              - p150a-jtag
+    runs-on: ${{ matrix.config.runs-on }}
+    container:
+      image: ghcr.io/tenstorrent/tt-metal/upstream-tests-bh:v0.58.0-rc26-5-g668476a4d2
+      volumes:
+        - /dev/hugepages-1G:/dev/hugepages-1G
+        - /dev/hugepages:/dev/hugepages
+      options: '--device /dev/tenstorrent --device /dev/bus/usb --privileged'
+    env:
+      ARCH_NAME: blackhole
+    defaults:
+      run:
+        shell: bash
+        working-directory: /home/user/tt-metal/
+    steps:
+      - id: mkdir-home-user-tt-metal
+        run: |
+          mkdir -p /home/user/tt-metal
+      - name: Download the latest firmware bundle
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ needs.build-fw-artifact.outputs.combined-fwbundle-artifact }}
+          path: /home/user/tt-metal
+      - name: Run the rescan-pcie.sh script
+        run: |
+          curl -o /tmp/rescan-pcie.sh https://raw.githubusercontent.com/tenstorrent/tt-zephyr-platforms/${{ github.sha }}/scripts/rescan-pcie.sh
+          source /tmp/rescan-pcie.sh
+      - name: Flash the firmware
+        run: |
+          sudo chmod -R a+rwX $HOME/.cargo $HOME/.cache
+          python -m venv .env
+          source .env/bin/activate
+          pip install git+https://github.com/tenstorrent/tt-flash.git
+          tt-flash --fw-tar /home/user/tt-metal/fw_pack-*.fwbundle --force
+      - name: Run Container Test
+        run: |
+          sh -c "$TEST_COMMAND"
+      - name: cleanup
+        if: ${{ always() }}
+        run: |
+          # Clean out metal
+          rm -f /home/user/tt-metal/fw_pack-*.fwbundle


.github/workflows/hardware-smoke.yml
```diff
@@ -268,3 +268,65 @@ jobs:
           python3 ./scripts/dmc_rtt.py -n --openocd /opt/tenstorrent/bin/openocd-rtt
           echo "SMC RTT logs:"
           python3 ./scripts/smc_rtt.py -n --openocd /opt/tenstorrent/bin/openocd-rtt
+
+  build-fw-artifact:
+    uses: ./.github/workflows/build-fw.yml
+    secrets:
+      SIGNATURE_KEY: ${{ secrets.SIGNATURE_KEY }}
+  hardware-metal-test:
```
dimitri-tenstorrent, line 276: Could we call this `metal-test` for visual ease of identification?

.github/workflows/hardware-smoke.yml
```diff
@@ -268,3 +268,65 @@ jobs:
...
+      ARCH_NAME: blackhole
+    defaults:
+      run:
+        shell: bash
+        working-directory: /home/user/tt-metal/
+    steps:
+      - name: Install Unzip
+        run: |
+          sudo apt update
+          sudo apt install -y unzip protobuf-compiler
```
dimitri-tenstorrent, line 308: We do not need unzip anymore, likely

.github/workflows/hardware-smoke.yml
```diff
@@ -268,3 +268,65 @@ jobs:
...
+          pip install git+https://github.com/tenstorrent/tt-flash.git
+          tt-flash --fw-tar /__w/tt-zephyr-platforms/tt-zephyr-platforms/fw_pack-*.fwbundle --force
+      - name: Run Container Test
+        run: |
+          sh -c "$TEST_COMMAND"
+      - name: cleanup
+        if: ${{ always() }}
+        run: |
+          # Clean out metal
+          rm -rf /__w/tt-zephyr-platforms/tt-zephyr-platforms/fw_pack-*.fwbundle ./extracted_fwbundle/fw_pack*.fwbundle
```
dimitri-tenstorrent, line 332: Do we need ./extracted... folder removal?

.github/workflows/hardware-smoke.yml
kkabilarTT: removed it

.github/workflows/hardware-smoke.yml
kkabilarTT: changed it to: rm -rf /__w/tt-zephyr-platforms/tt-zephyr-platforms/fw_pack-*.fwbundle ./extracted_fwbundle

.github/workflows/hardware-smoke.yml
kkabilarTT: done

dimitri-tenstorrent (approved):
Approved on my end :)

.github/workflows/hardware-smoke.yml
```diff
@@ -268,3 +268,61 @@ jobs:
...
+        run: |
+          curl -o /tmp/rescan-pcie.sh https://raw.githubusercontent.com/tenstorrent/tt-zephyr-platforms/${{ github.sha }}/scripts/rescan-pcie.sh
+          source /tmp/rescan-pcie.sh
+      - name: Flash the firmware
+        run: |
+          sudo chmod -R a+rwX $HOME/.cargo $HOME/.cache
+          python -m venv .env
+          source .env/bin/activate
+          pip install git+https://github.com/tenstorrent/tt-flash.git
+          tt-flash --fw-tar /__w/tt-zephyr-platforms/tt-zephyr-platforms/fw_pack-*.fwbundle --force
```
danieldegrasse, line 320: Why use an absolute path here? Shouldn't the fwbundle be in the current working directory, or am I reading these steps wrong?

kkabilarTT: That's where the fwbundle gets downloaded so I'm just pointing to that location on line 320

danieldegrasse:
Not clear why the job is failing, but it seems like there is a legitimate issue. Could the issue be that we build the firmware bundle as part of multiple jobs (and produce multiple artifacts as a result) Maybe try turning the code in `build-fw` into a reusable action, and see if that fixes things? Not sure if github supports including non-reusable workflows like this

dimitri-tenstorrent:
@danieldegrasse I think the issue with P100 job has two do with the permissions on P100 runner setup differently
See the error for downloading the file `/__w/tt-zephyr-platforms/tt-zephyr-platforms/firmware-p300b` on p100 machine that same file is downloaded successfully elsewhere on machines P100a and 150a.

@kkabilarTT Could you please see a). that we can download to some sort of temp folder (let's clean it up after) b). that we download only the needed for our test likely `combined-fwbundle` or `firmware-p100`.

Let's get this over the finish line guys :)

danieldegrasse:
> @danieldegrasse I think the issue with P100 job has two do with the permissions on P100 runner setup differently
See the error for downloading the file /__w/tt-zephyr-platforms/tt-zephyr-platforms/firmware-p300b on p100 machine that same file is downloaded successfully elsewhere on machines P100a and 150a.

@cfriedt can you check the permissions on the P100 runner to see if it is different than the P100a/P150a?

.github/workflows/build-fw.yml
cfriedt, line 1: There should really not be any changes to this file

cfriedt (dismissed):
Please undo changes to `build-fw.yml`

cfriedt:
It's hard to be find the root cause of the error that was happening before because the branch was recently force-pushed, but given that builds were failing on github runners and not on self-hosted runners, there is likely something to do with this PR that was causing builds to fail, rather than some infrastructure issue.

Corollary: this PR is passing https://github.com/tenstorrent/tt-zephyr-platforms/pull/249

.github/workflows/build-fw.yml
kkabilarTT: Ok, I removed the changes to build-fw.yml

cfriedt:
@kkabilarTT - the errors in CI look really weird - like DNS and other network errors.

I'm just going to try to re-run the jobs to see if we can get this past the finish line

cfriedt:
There are no differences in permissions.

@kkabilarTT - it's definitely failing inside of a container. I cleared the runner working directory, and there is no output when re-running this job in isolation.

https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15148134675/job/42640153306?pr=227

However, I did note that after this job ran, `tt-smi` stopped working.

```shell
Info : flash mode : dual-bank
Info : Flash write discontinued at 0x08008c60, next section at 0x0800c000
Warn : Adding extra erase range, 0x08008c60 .. 0x08008fff
Info : Padding image section 1 at 0x08016066 with 2 bytes (bank write end alignment)
Warn : Adding extra erase range, 0x08016068 .. 0x080167ff
auto erase enabled
wrote 77000 bytes from file /opt/tenstorrent/fw/stable/tt_blackhole_tt_blackhole_bmc/bmc_3_1_0.hex in 1.817159s (41.381 KiB/s)

Info : Unable to match requested speed 2000 kHz, using 1800 kHz
Info : Unable to match requested speed 2000 kHz, using 1800 kHz
shutdown command invoked
Rescanning the PCI bus
1
```

Normally, at the end of running `rescan-pcie.sh`, I expect to see something like this:

```shell
./scripts/rescan-pcie.sh
Removing PCI device /sys/bus/pci/devices/0000:01:00.0 with vendor id 1e52
1
Rescanning the PCI bus
1
```

But the log from your job above does not show that.

.github/workflows/hardware-smoke.yml
```diff
@@ -268,3 +268,61 @@ jobs:
...
+    defaults:
+      run:
+        shell: bash
+        working-directory: /home/user/tt-metal/
+    steps:
+      - name: Download the latest firmware bundle
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ needs.build-fw-artifact.outputs.combined-fwbundle-artifact }}
+          path: .
```
cfriedt, line 309: @kkabilarTT - this step seems to be failing consistently, and it appears that the directory probably does not exist in the container.

```shell
Run actions/download-artifact@v4
...
Starting download of artifact to: /__w/tt-zephyr-platforms/tt-zephyr-platforms
...
```
Can you try this instead?
```suggestion
    steps:
      - name: Download the latest firmware bundle
        uses: actions/download-artifact@v4
        with:
          name: ${{ needs.build-fw-artifact.outputs.combined-fwbundle-artifact }}
          path: /home/user/tt-metal
```

dimitri-tenstorrent: This is a great suggestion @cfriedt !

.github/workflows/hardware-smoke.yml
cfriedt: I agree - I think that this should be changed.

cfriedt, line 320: Try this out please

```suggestion
          tt-flash --fw-tar /home/user/tt-metal/fw_pack-*.fwbundle --force
```

.github/workflows/hardware-smoke.yml
```diff
@@ -268,3 +268,61 @@ jobs:
...
+          pip install git+https://github.com/tenstorrent/tt-flash.git
+          tt-flash --fw-tar /__w/tt-zephyr-platforms/tt-zephyr-platforms/fw_pack-*.fwbundle --force
+      - name: Run Container Test
+        run: |
+          sh -c "$TEST_COMMAND"
+      - name: cleanup
+        if: ${{ always() }}
+        run: |
+          # Clean out metal
+          rm -rf /__w/tt-zephyr-platforms/tt-zephyr-platforms/fw_pack-*.fwbundle
```
cfriedt, line 328:
```suggestion
          rm -f /home/user/tt-metal/fw_pack-*.fwbundle
```

.github/workflows/hardware-smoke.yml
cfriedt: @kkabilarTT - I'm not 100% certain if `/home/user/tt-metal` already exists in the container, but if it does not, it would be necessary to create the directory in advance of using it.

So something like this above this rule.

```yaml
      - id: mkdir-home-user-tt-metal
        run: |
          mkdir -p /home/user/tt-metal
```

kkabilarTT:
@cfriedt I made changes to the path to download the artifact in `/home/user/tt-metal` directory as mentioned but there still seems to be a permission denied error for p100: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15170077045/job/42657985202?pr=227#step:5:22

cfriedt:
Currently, there are zero differences between the CI configuration and file permission structure between runners.

The one difference that I have seen is that the runner with the failing jobs is still on Ubuntu 20.04 whereas the runners with the passing jobs are on Ubuntu 24.04.

I'll see if I can do an upgrade and whether that will fix the issue.

dimitri-tenstorrent:
> Currently, there are zero differences between the CI configuration and file permission structure between runners.
> 
> However, "Download the latest firmware bundle" is [still failing](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119) with the extremely cryptic error message below.
> 
> ```shell
> Run actions/download-artifact@v4
>   with:
>     name: combined-fwbundle
>     path: /home/user/tt-metal
>     merge-multiple: false
>     repository: tenstorrent/tt-zephyr-platforms
>     run-id: [1](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119#step:4:1)5172081226
>   env:
>     ARCH_NAME: blackhole
> /usr/bin/docker exec  3e6573c0a9949f0ce086c0ac3b0a[2](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119#step:4:2)727067ab4ad[3](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119#step:4:3)bead26146ffef5f932d949f sh -c "cat /etc/*release | grep ^ID"
> Downloading single artifact
> Preparing to download the following artifacts:
> - combined-fwbundle (ID: 3172090260, Size: 28033[4](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119#step:4:4)1, Expected Digest: sha256:d8988[5](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119#step:4:5)44047f17ad5e9e50b153e724039dd23[6](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119#step:4:6)ef25d5acf992419c3967dd9519)
> Redirecting to blob download url: https://productionresultssa1.blob.core.windows.net/actions-results/bfe064ee-a22e-408e-bd04-[7](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119#step:4:7)d4a2709199d/workflow-job-run-5740d0aa-6df[8](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119#step:4:8)-53db-2472-d6464abbc1[9](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119#step:4:9)b/artifacts/2e6b2eb17497423f756776423828775dd58ab59ee494840e24ee90ef4b017811.zip
> Starting download of artifact to: /home/user/tt-metal
> (node:22) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
> (Use `node --trace-deprecation ...` to show where the warning was created)
> SHA256 digest of downloaded artifact is d8988544047f17ad5e9e50b[15](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15172081226/job/42709975119#step:4:16)3e724039dd236ef25d5acf992419c3967dd9519
> Artifact download completed successfully.
> Total of 1 artifact(s) downloaded
> Error: Unable to download artifact(s): EACCES: permission denied, open '/__w/_temp/_runner_file_commands/set_output_5c3fd3ca-343f-4ce8-8950-21f0503e31e2'
> ```
> 
> Presumably `_temp/` files are all cleaned after jobs run, or perhaps it's because this job is being run inside of a container, but the `_work/_temp/_runner_file_commands` subdirectory is empty on all of the GitHub runners (although they all have the same permissions).
> 
> The one difference that I have seen is that the runner with the failing jobs is still on Ubuntu 20.04 whereas the runners with the passing jobs are on Ubuntu 24.04.
> 
> I'll see if I can do the required upgrades (upgrade first from 20.04 to 22.04, and then from 22.04 to 24.04) and whether that will fix the issue.

My understanding is that the permission issue is not the downloading the file but rather storing the log output of the action. This is based on the file that is failing to get created - `/__w/_temp/_runner_file_commands/set_output_*`

cfriedt:
> My understanding is that the permission issue is not the downloading the file but rather storing the log output of the action. This is based on the file that is failing to get created - `/__w/_temp/_runner_file_commands/set_output_*`

The directory structure / permission scheme is identical. Maybe downloading the artifact via HTTPS makes sense, but it shouldn't be necessary.

cfriedt:
The bug that we are encountering using `actions/download-artifact` is a known issue.

https://github.com/actions/upload-artifact/issues/640

It's possible that the self-hosted runner may have been installed with slightly different parameters; it's the oldest one that we have running. That would explain why the other self-hosted runners are not encountering the same issue (because they were installed after the exact instructions were documented).

In the future, I think we should have an ansible playbook for this, but I also don't want to prevent this PR from being merged any longer.

cfriedt:
Filed https://github.com/tenstorrent/tt-zephyr-platforms/issues/253 and will disable the test on p100 until we can resolve it on that platform.

cfriedt (approved):
LGTM - thanks @kkabilarTT and @dimitri-tenstorrent for this addition to our CI

PR #229, opened 2025-05-14
blobs: Updating MRISC release version.

abuckTT:
Each board type now uses own GDDR parameter table.
2.7 MRISC changelog:
- Added messaging capability
- Added a message to kick off memory test.
- Creates parameter table for each board.
- Improvements for Galaxy
- Added option to use FW based power management


abuckTT:
FYI @hnaikTT

cfriedt:
Similarly to the recent ERISC FW update from @bingliTT , we were trying to add a basic testsuite for ARC that could be run in CI to verify the new mailbox interface.

@hnaikTT - you may need to clone this branch assuming there is not enough time to add a test to this PR.

cfriedt:
Please rebase to get rid of the merge commit. @hnaikTT - you may want to clone this branch and let me know when you are ready to add tests.

cfriedt (changes requested):
Remove merge commit and add tests.

cfriedt:
rebased - just waiting to make sure that the builds and at least a few tests succeed

cfriedt (dismissed):
The build needs to work for BH Galaxy as well. If that needs a separate FW, then we need to have a new blob added. Ping me when you are able to @hnaikTT

hnaikTT:
@cfriedt I'll add the galaxy blob in the next release.

cfriedt:
Also, just a note, that all that needs to succeed in this PR are the blobs.

It may be the case that some of the hardware workflows fail, but that has been because firmware may be left in a bad state by previous CI jobs.

cfriedt (approved):
Tentative approval - I'd like to see CI green after this, hopefully.

cfriedt:
> @cfriedt I'll add the galaxy blob in the next release.

Just for clarification, we added the missing blob in this PR. It's impossible to make a release if the build is failing, so the blob was necessary.

PR #232, opened 2025-05-14
scripts: pcie_utils.py: rescan even if no card is on bus

danieldegrasse:
Commit d33106d (app: pytest: e2e_smoke: support rescan if no card is on bus) added this behavior to the e2e smoke test, but commit d1ea6e1 (boards: tt_blackhole: rename bmc to dmc to avoid confusion) did not port the behavior into the new pcie_utils.py script. Reintroduce it.

### DIFF
--- a/scripts/pcie_utils.py
+++ b/scripts/pcie_utils.py
@@ -27,18 +27,23 @@ def rescan_pcie():
     """
     # First, we must find the PCIe card to power it off
     dev = find_tt_bus()
-    if dev is None:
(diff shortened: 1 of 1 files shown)


cyberDrake:
there is a change in behavior here if there is no card found - previously there was an error message stating `No tenstorrent card found to power off`, now if no card is found it looks like it just silently ignores it.

if that is what is desired, LGTM. going ahead and marking as approved so this isn't blocked.

cfriedt (approved):

PR #234, opened 2025-05-14
scripts: tt-console: support accessing console via JTAG

danieldegrasse:
This PR adds support into tt-console for accessing the console data via JTAG. It includes the following changes (broadly):
- rework of tt-console to split memory access code out from console logic
- inclusion of libjaylink into project as module, to support JTAG access to ARC memory
- support for reading/writing ARC memory within tt-console program
- additional testcases for tt-console program to verify memory access drivers (PCIE and JTAG based) are working
- additional vuart testcase to verify JTAG functionality

> [!WARNING]
> This PR updates the license of tt-console to be GPL-2.0, because we now link it with libjaylink (which is GPL-2.0 only).
> This is required unless we want to use SEGGER's JLink libary (which may have more licensing restrictions and require
> making tt-console closed source)

Once we have tt-console updated on CI machines, we should be able to move to make the VUART serial the default console for the SMC, because we can access the console data even without PCIe enumeration

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -27,8 +27,19 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          path: tt-zephyr-platforms
           ref: ${{ github.ref }}
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+
+      - name: Fetch dependencies
+        run: |
+          pip install west
+          west init -l tt-zephyr-platforms
+          west update libjaylink
+
       - name: Run the script
         run: |
-          gcc -Iinclude -O0 -g -Wall -Wextra -Werror -std=gnu11 -o tt-console \
-            scripts/tt-console/console.c
+          cd tt-zephyr-platforms/scripts/tt-console
+          ./build.sh
--- a/.github/workflows/prepare-zephyr/action.yml
+++ b/.github/workflows/prepare-zephyr/action.yml
@@ -17,12 +17,19 @@ runs:
       with:
         python-version: '3.10'
 
-    - name: Install curl
+    - name: Install dependencies
       shell: bash
       run: |
         if ! command -v curl &> /dev/null ; then
-          apt-get update
-          apt-get install -y curl
+          sudo apt-get update
+          sudo apt-get install -y curl
+        fi
+        if pkg-config --exists libusb-1.0; then
+          echo "Found libusb"
+        else
+          echo "Installing libusb"
+          sudo apt-get update
+          sudo apt-get install -y libusb-1.0-0-dev
         fi
 
     - name: Install Rust
--- a/MAINTAINERS.yml
+++ b/MAINTAINERS.yml
@@ -63,6 +63,13 @@ Release Notes:
   description: |
     Maintained upstream.
 
+"West project: libjaylink":
+  status: odd fixes
+  files: []
+  labels: []
+  description: |
+    Maintained upstream.
+
 "West project: zephyr":
   status: odd fixes
   files: []
--- a/app/smc/pytest/vuart.py
+++ b/app/smc/pytest/vuart.py
@@ -55,8 +55,8 @@ def test_boot_banner_pcie(tt_console: Path, arc_chip):
 def test_boot_banner_jtag(tt_console: Path, arc_chip):
     arc_chip.axi_read32(ARC_STATUS)
 
-    # run 'tt-console' in quiet mode, and timeout after 10 seconds
-    cmd = f"{tt_console} -q -j -w 10000"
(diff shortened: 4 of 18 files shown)


cfriedt (changes requested):
This is generally really great work.

However, it's probably best to keep it on hold for a while because

a) we should really not introduce anything other than Apache-2 licensed code in here unless you want to go down a legal rabbit hole for several months (so that is a NAK), and

b) the tooling team should probably own it.

Ultimately, we also want to rarely, if ever, have to use JTAG on our cards. So the vuart, or something like it, should be the only thing people ever need externally.

The other eventuality I would hope we get to, is to use a more standard industry solution that will work well with the kernel, e.g. virtio.

I'm hoping that tt-console is kind of a stop-gap solution that is done in conjunction with the tools team.

cfriedt:
Please don't get me wrong - I just don't want to go too far down this path.

> > This is generally really great work.
> That is fair- one way to work around this would be to link dynamically with `libjaylink`- that way we could keep this code apache2 licensed, right?

A quick survey on Google says that even dynamically linking to GPL code means that the app needs to be GPL too because the app constitutes a "derived work".

Apparently, if libjaylink were LGPL, then it would be possible to dynamically link an Apache 2.0 binary to it.

I'm not sure if I agree with the first point above, but I'm also not a lawyer 🫠 

> > b) the tooling team should probably own it.
> 
> CC @warthog9 - essentially what `tt-console` is currently is a program for accessing a virtual console (IE a circular ram buffer) on the SMC, which allows us to dump logs from running firmware. IMHO this is something we could (and should probably) consider moving into the tooling team as it stabilizes. I don't think this tool is at that point yet though.

I think the tools team is already aware of it - there was a separate chat somewhere where it became an issue and we were asked to revert it. I don't think that's going to happen, but also, I don't want to push things. It's a tool, we need it to extract data efficiently from ARC as developers. We shouldn't need permission for that, but at the same time, there is no better alternative. Still, we can't distribute it as an official tool.

> > Ultimately, we also want to rarely, if ever, have to use JTAG on our cards. So the vuart, or something like it, should be the only thing people ever need externally.
> 
> I agree that we don't want to use JTAG as a first-class solution for accessing the cards. However, I strongly feel the support should be available. 

It's impossible to do that in cloud, in user deployments, where workloads will be, where the cards will be stressed in ways that we haven't anticipated. Nobody ever anticipates that they need to buy a JTAG adapter when they purchase a PCIe card.

> * debugging a NOC hang- PCIe is down, JTAG will still work
> * debugging a card failing to enumerate on PCIe

Ideally, we would ensure that firmware and kernel code would work together to ensure that the NOC is mostly used in ways that guarantee mutual exclusion, no possibility of contention, etc. It's a long-tail problem (meaning it might not require one or two changes to fix permanently, and it could take a while). This is one of those SLA things - so the NOC should work 99.999% of the time.

We should fix the issue of a card failing to enumerate by using mcuboot, so 99.999% of the time, pcie always enumerates.

We need to be able to rely on post-mortem debug info for diagnosing and remediating issues. It's maybe optimistic, but we'll have to aim for it anyway.

> > The other eventuality I would hope we get to, is to use a more standard industry solution that will work well with the kernel, e.g. virtio.
> > I'm hoping that tt-console is kind of a stop-gap solution that is done in conjunction with the tools team.
> 
> I agree here- however I think even when we switch to that solution, we should still offer tt-console as a method to access the VIRTIO ram buffers via JTAG, for the same fallback reasons I mentioned above.

Except JTAG should only really be used during bringup. We can't offer JTAG as a solution to anybody.

> The broader goal of this PR is to get to a point where tt-vuart is the default UART for all BH cards. There are two paths I see to making this a reality:
> 
> * turn on PCIe enumeration by default for all SMC applications. This will _usually_ work, but still won't cover debugging in early boot time crashes before enumeration completes
> * enable JTAG support in tt-console
> 
> Which of the above do you prefer as a path forwards? Ideally I'd prefer to turn on JTAG in tt-console, but if there is a strong argument as to why we shouldn't have that support we can look at enabling PCIe instead.

I think we will have to rely on PCIe enumeration by default.

One thing I've done previously is use `devmux` to enable dynamic switching between a physical uart and a virtual pcie uart. Additionally, there can be multiple shell interfaces too, and dynamically adjustable log verbosity.

cc: @ycsin

Our firmware is stil relatively young, so it will take a bit of time to get there (and for sure the more people we have working on it, the better).

> I'm fine to keep this PR on hold if we need to, but I do _not_ want to let the broader issue of moving to tt-vuart as the default console slip- the sooner we get good quality debug support for our production builds, the less pain we will be in.

> I'm aware that we currently enable tt-vuart in release builds, but our testing scope is still pretty limited- I'd like to get to the point where we use this tool by default, so we can increase testing scope to all applications running on the SMC

💯 I agree - both of us have very little bandwidth at the moment to work on tasks that will have a large impact.

PR #239, opened 2025-05-15
scripts: build-combined-fwbundle: use semantic version in filename

### LINKED ISSUE #228: version: use semantic versioning string for bundle-version file
**Is your enhancement proposal related to a problem? Please describe.**
Previously, there was a `MAJ.MIN.PATCH.RC` versioning convention for .fwbundle files.

However, given that we have recently dropped the 80 prefix and are moving closer to fully semantic versions, it would make sense to also use the semantic version string format for .fwupdate files rather than the bespoke versioning convention above.

**Describe the solution you'd like**
Change scripts, github actions, etc, to use a fully semantic bundle version.

So instead of `fw_pack-18.3.0.1.fwbundle` being generated for the next `-rc1` tag, use `fw_pack-18.3.0-rc1.fwbundle`.

Final releases would then be the same, minus the `-rcN`.

The binary integer version number associated with the bundle version will still have the RC number encoded in the least-significant byte.

`18.3.0-rc1` would be encoded as `(18 << 24) | (3 << 16) | (0 << 8) | (1 << 0)` and `18.3.0` would be encoded as `(18 << 24) | (3 << 16) | (0 << 8) | (0 << 0)`

**Describe alternatives you've considered**

**Additional context**

This ticket set up to track
https://github.com/tenstorrent/tt-firmware/issues/16

Side note: we probably don't want people filing firmware issues at `tt-firmware` @warthog9

cfriedt:
Use the semantic version in the filename of the firmware bundle rather than the quad-dotted versioning scheme historically used.

Going forward, release candidate firmware packs will have names like `fw_pack-18.3.0-rc1.fwbundle` rather than `fw_pack-18.3.0.1.fwbundle`, and
final releases will have names like `fw_pack-18.3.0.fwbundle` rather than
`fw_pack-18.3.0.0.fwbundle`.

Fixes #228

Testing done (manual, but also from the "build combined fwbundle" job in this PR):
```shell
Creating fw_pack-18.3.0-rc1.fwbundle
Wrote fwbundle to fw_pack-18.3.0-rc1.fwbundle
```

### DIFF
--- a/doc/release/migration-guide-18.3.0.md
+++ b/doc/release/migration-guide-18.3.0.md
@@ -5,3 +5,8 @@
 This document lists recommended and required changes for those migrating from the previous v18.2.0 firmware release to the new v18.3.0 firmware release.
 
 [comment]: <> (UL by area, indented as necessary)
+
+* Going forward, and following the more standard semantic versioning of firmware, the filename
+  of the combined firmware bundles will be of the form
+  * `fw_pack-<major>.<minor>.<patch>.fwbundle` for final releases, and
+  * `fw_pack-<major>.<minor>.<patch>-rc<N>.fwbundle` for release candidates (pre-releases)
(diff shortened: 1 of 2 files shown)


knauth (approved):
LGTM, thanks!

PR #242, opened 2025-05-15
doc: release: v18.3.0: note usage of input power instead of current

afongTT:
Update release notes and migration guide to note usage of input power instead of input current.


cfriedt (approved):
Awesome - thank you @afongTT 🙏

PR #248, opened 2025-05-19
lib: tenstorrent: support reading telemetry tags via SMBUS

danieldegrasse:
This PR enables reading telemetry tags via SMBUS, using the following scheme:

| Register       | Address | Width   | Usage                                                                                                                             |
|----------------|---------|---------|-----------------------------------------------------------------------------------------------------------------------------------|
| TELEMETRY_TAG  | 0x26    | 8 bits  | Write only. Write with telemetry tag value to select which telemetry field will be read when reading from TELEMETRY_DATA register |
| TELEMETRY_DATA | 0x27    | 32 bits | Read only. Read to get the latest telemetry data for the tag programmed to TELEMETRY_TAG register                                 |

The `TELEMETRY_TAG` register is "sticky"- so a controller only needs to program it once, and then can poll the `TELEMETRY_DATA` register for new telemetry data for that tag.

This PR also includes two other notable changes:
- remove the `tag_table` definition- telemetry tags now map directly to their offset in the telemetry table. This way, we don't need to perform any remapping logic on the SMC to report telemetry values based on their tag. The tag table is still defined so existing tools like tt-smi will function, but it is now effectively a 1:1 map
- add an SMBUS test sequence to the DMFW. This sequence is only enabled for test builds, but it will allow us to test SMBUS messages from the controller side, with the SMC running as the peripheral. Currently it is only used to test the telemetry tags support, but we could also use this to test fan control or board power limits.

### DIFF
--- a/lib/tenstorrent/bh_arc/telemetry.c
+++ b/lib/tenstorrent/bh_arc/telemetry.c
@@ -44,13 +44,70 @@ struct telemetry_table {
 };
 
 /* Global variables */
-static struct telemetry_table telemetry_table;
+static struct telemetry_table telemetry_table = {
+	.tag_table = {
+		[0] = {TAG_BOARD_ID_HIGH, TELEM_OFFSET(TAG_BOARD_ID_HIGH)},
+		[1] = {TAG_BOARD_ID_LOW, TELEM_OFFSET(TAG_BOARD_ID_LOW)},
+		[2] = {TAG_ASIC_ID, TELEM_OFFSET(TAG_ASIC_ID)},
+		[3] = {TAG_HARVESTING_STATE, TELEM_OFFSET(TAG_HARVESTING_STATE)},
+		[4] = {TAG_UPDATE_TELEM_SPEED, TELEM_OFFSET(TAG_UPDATE_TELEM_SPEED)},
+		[5] = {TAG_VCORE, TELEM_OFFSET(TAG_VCORE)},
+		[6] = {TAG_TDP, TELEM_OFFSET(TAG_TDP)},
+		[7] = {TAG_TDC, TELEM_OFFSET(TAG_TDC)},
+		[8] = {TAG_VDD_LIMITS, TELEM_OFFSET(TAG_VDD_LIMITS)},
+		[9] = {TAG_THM_LIMITS, TELEM_OFFSET(TAG_THM_LIMITS)},
+		[10] = {TAG_ASIC_TEMPERATURE, TELEM_OFFSET(TAG_ASIC_TEMPERATURE)},
+		[11] = {TAG_VREG_TEMPERATURE, TELEM_OFFSET(TAG_VREG_TEMPERATURE)},
+		[12] = {TAG_BOARD_TEMPERATURE, TELEM_OFFSET(TAG_BOARD_TEMPERATURE)},
+		[13] = {TAG_AICLK, TELEM_OFFSET(TAG_AICLK)},
+		[14] = {TAG_AXICLK, TELEM_OFFSET(TAG_AXICLK)},
+		[15] = {TAG_ARCCLK, TELEM_OFFSET(TAG_ARCCLK)},
+		[16] = {TAG_L2CPUCLK0, TELEM_OFFSET(TAG_L2CPUCLK0)},
+		[17] = {TAG_L2CPUCLK1, TELEM_OFFSET(TAG_L2CPUCLK1)},
+		[18] = {TAG_L2CPUCLK2, TELEM_OFFSET(TAG_L2CPUCLK2)},
+		[19] = {TAG_L2CPUCLK3, TELEM_OFFSET(TAG_L2CPUCLK3)},
+		[20] = {TAG_ETH_LIVE_STATUS, TELEM_OFFSET(TAG_ETH_LIVE_STATUS)},
+		[21] = {TAG_GDDR_STATUS, TELEM_OFFSET(TAG_GDDR_STATUS)},
+		[22] = {TAG_GDDR_SPEED, TELEM_OFFSET(TAG_GDDR_SPEED)},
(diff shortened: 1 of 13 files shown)


lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -39,19 +39,76 @@ struct telemetry_entry {
 struct telemetry_table {
 	uint32_t version;
 	uint32_t entry_count;
-	struct telemetry_entry tag_table[TELEM_ENUM_COUNT];
-	uint32_t telemetry[TELEM_ENUM_COUNT];
+	struct telemetry_entry tag_table[TAG_COUNT];
```
cfriedt, line 42: We should be able to remove this. It provides zero information gain, FWICT

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -39,19 +39,76 @@ struct telemetry_entry {
...
+		[45] = {TAG_GDDR_4_5_CORR_ERRS, TELEM_OFFSET(TAG_GDDR_4_5_CORR_ERRS)},
+		[46] = {TAG_GDDR_6_7_CORR_ERRS, TELEM_OFFSET(TAG_GDDR_6_7_CORR_ERRS)},
+		[47] = {TAG_GDDR_UNCORR_ERRS, TELEM_OFFSET(TAG_GDDR_UNCORR_ERRS)},
+		[48] = {TAG_MAX_GDDR_TEMP, TELEM_OFFSET(TAG_MAX_GDDR_TEMP)},
+		[49] = {TAG_ASIC_LOCATION, TELEM_OFFSET(TAG_ASIC_LOCATION)},
+		[50] = {TAG_BOARD_POWER_LIMIT, TELEM_OFFSET(TAG_BOARD_POWER_LIMIT)},
+		[51] = {TAG_INPUT_POWER, TELEM_OFFSET(TAG_INPUT_POWER)},
+		[52] = {TAG_TELEM_ENUM_COUNT, TELEM_OFFSET(TAG_TELEM_ENUM_COUNT)},
+	},
+};
```
cfriedt, line 103: Can this entire table be deleted?

Maybe I'm missing something here too, and I believe I've asked about it a long time ago, but there also seems to be zero information gain here.

```
[0] = {1, 1},
[1] = {2, 2},
...
```

cfriedt:
This change looks fine to me - I think the table is maybe wasting space though, not sure.

danieldegrasse:
> This change looks fine to me - I think the table is maybe wasting space though, not sure.

I *think* tt-smi relies on the existence of this table: https://github.com/tenstorrent/luwen/blob/a8357d0d2a747c9f0596401bf37ec81fa13ae1b2/crates/luwen-if/src/chip/blackhole.rs#L702. @DRvader, can you confirm if we could drop the table? We could also implement some sort of migration to drop the table long term

lib/tenstorrent/bh_arc/telemetry.c
cfriedt: cc @DRvader

cfriedt:
> We could also implement some sort of migration to drop the table long term

That would be great @DRvader.

Specifically one of the reasons why RPC is kind of ideal for telemetry is to avoid bespoke artifacts / tech debt like this.

cfriedt (approved):

lib/tenstorrent/bh_arc/telemetry.c
TTDRosen: The table, has the extra level of indirection to support explicit field deletion. And fields have already been removed so the mapping is slightly offset. I know that's the case because some code did assume the 1:1 mapping and broke when the fields were removed 😆

TTDRosen:
Doesn't this break the tags?

afongTT:
Requesting to revert this in #250 to restore the layer of indirection we had for tag and telemetry.

afongTT:
> Specifically one of the reasons why RPC is kind of ideal for telemetry is to avoid bespoke artifacts / tech debt like this.

Could you link to an example of something like this? We can keep it in mind for future projects / features.

cfriedt:
> Doesn't this break the tags? BOARD_ID_HIGH was one previously (because of a removed tag, and now it's zero again)? Am I reading the code wrong?

TAG_BOARD_ID_HIGH := 1 [here](https://github.com/tenstorrent/tt-zephyr-platforms/blob/61a57b4ecb92e8b302e56938d995642aaf3f5ce5/lib/tenstorrent/bh_arc/telemetry.h#L18) which is unchanged before and after this commit.

The zero'th entry in the telemetry table is `[0] = { 1, 1 }` [here](https://github.com/tenstorrent/tt-zephyr-platforms/blob/61a57b4ecb92e8b302e56938d995642aaf3f5ce5/lib/tenstorrent/bh_arc/telemetry.c#L49) which is exactly the same as it was [before this change](https://github.com/tenstorrent/tt-zephyr-platforms/blob/c5a864e647fde9b6cade688eaf231b99564ddcf8/lib/tenstorrent/bh_arc/telemetry.c#L273).

cfriedt:
> > Specifically one of the reasons why RPC is kind of ideal for telemetry is to avoid bespoke artifacts / tech debt like this.
> 
> Could you link to an example of something like this? We can keep it in mind for future projects / features.

First step would be using a straight linear translation.

For examples of RPC, see

https://docs.zephyrproject.org/latest/samples/modules/nanopb/README.html#nanopb
https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/modules/nanopb

https://docs.zephyrproject.org/latest/samples/modules/thrift/hello/README.html#thrift-hello
https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/modules/thrift

There is also CBOR, but I can't seem to find a link to a sample.e

PR #252, opened 2025-05-21
ci: use download-artifact@v3

cfriedt:
There is an important note on [the landing page](https://github.com/actions/download-artifact) for actions/download-artifact in that it warns users of GHES that v4+ is not currently supported on GitHub Enterprise Server.

Screenshot 2025-05-21 at 4 31 51 PM

It's not 100% certain if that applies to us, but I guess it doesn't hurt to try? ¯\_(ツ)_/¯.


cfriedt:
v3 does not work

PR #257, opened 2025-05-23
bh_arc: expose fused ASIC ID in telemetry

kikkiakki:
Each BH ASIC is fused with a unique 64 bit ASIC ID during testing. Add two new fields to telemetry to expose this, ASIC_ID_HIGH and ASIC_ID_LOW.

Remove ASIC_ID from telemetry table, which was added when a 32 bit value was expected.

### DIFF
--- a/lib/tenstorrent/bh_arc/telemetry.h
+++ b/lib/tenstorrent/bh_arc/telemetry.h
@@ -69,10 +69,12 @@
 #define TAG_ASIC_LOCATION        52
 #define TAG_BOARD_POWER_LIMIT    53
 #define TAG_INPUT_POWER          54
+#define TAG_ASIC_ID_HIGH         62
+#define TAG_ASIC_ID_LOW          63
 /* Not a real tag, signifies the last tag in the list.
  * MUST be incremented if new tags are defined
  */
-#define TAG_COUNT                55
+#define TAG_COUNT                64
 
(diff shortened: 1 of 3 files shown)


cfriedt (dismissed):
LGTM -  I restarted the failing CI jobs because the p100a that we have in the yyz lab is not re-enumerating and has been taken offline for servicing. Hopefully we have some redundancy for p100a in Austin?

lib/tenstorrent/bh_arc/telemetry.h
```diff
@@ -69,10 +69,12 @@
 #define TAG_ASIC_LOCATION        52
 #define TAG_BOARD_POWER_LIMIT    53
 #define TAG_INPUT_POWER          54
+#define TAG_ASIC_ID_HIGH         55
+#define TAG_ASIC_ID_LOW          56
```
afongTT, line 73: Wait, these conflict with existing tags. Please reserve tag ID internally here for now:
https://tenstorrent.atlassian.net/wiki/spaces/syseng/pages/666665152/Telemetry#Tag-IDs

afongTT (changes requested):
Update tag numbers to not conflict with existing.

lib/tenstorrent/bh_arc/telemetry.h
kikkiakki: Done

afongTT (approved):

PR #258, opened 2025-05-24
lib: jtag_bootrom: add timeout wiating for jtag id

cfriedt:
Instead of looping indefinitely, which prevents any other action from the DMC to be taken, timeout after a default period of 5000 ms and return an error.

### DIFF
--- a/lib/tenstorrent/jtag_bootrom/jtag_bootrom.c
+++ b/lib/tenstorrent/jtag_bootrom/jtag_bootrom.c
@@ -36,19 +36,22 @@ bool jtag_axiwait(const struct device *dev, uint32_t addr)
 	return !jtag_axi_read32(dev, addr, &value);
 }
 
-void jtag_bitbang_wait_for_id(const struct device *dev)
+static int jtag_bitbang_wait_for_id(const struct device *dev, k_timeout_t timeout)
 {
 	uint32_t reset_id = 0;
 
-	while (true) {
+	for (k_timepoint_t end = sys_timepoint_calc(timeout); !K_TIMEOUT_EQ(timeout, K_NO_WAIT);
+	     timeout = sys_timepoint_timeout(end)) {
 		jtag_reset(dev);
 		jtag_read_id(dev, &reset_id);
 
 		if (reset_id == 0x138A5) {
-			break;
+			return 0;
 		}
 		k_yield();
 	}
+
+	return -ETIMEDOUT;
 }
 
 static const __maybe_unused struct gpio_dt_spec arc_rambus_jtag_mux_sel =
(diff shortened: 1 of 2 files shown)


lib/tenstorrent/jtag_bootrom/jtag_bootrom.c
```diff
@@ -36,19 +36,30 @@ bool jtag_axiwait(const struct device *dev, uint32_t addr)
...
-		if (reset_id == 0x138A5) {
-			break;
+		if (reset_id == id) {
+			return 0;
 		}
 		k_yield();
-	}
+
+		timeout = sys_timepoint_timeout(end);
+	} while (!K_TIMEOUT_EQ(timeout, K_NO_WAIT));
```
alewycky-tenstorrent, line 60: This is equivalent to `while (!sys_timeout_expired(end));`. Did you want the updated `timeout` for any purpose?

cfriedt:
> Is the intention of this to allow future code to do something when the JTAG ID polling times out?

Our production `app/dmc` definitely requires blackhole to be "up", so I would assume even _today's_ code should do _something_ when that doesn't happen. Being able to act on that error situation would definitely be required to assume that there is some kind of determinism. 

> I don't think the current code has any handling for the error case?

The error handling would kind of be implicit in the BIST and bootloader. I.e. if the system doesn't come up, then "die". In case of a pending firmware upgrade, that would ensure that we don't mark the firmware upgrade as successful and we revert back to the previous stable firmware.

If that isn't the order of events that's happening today, then there has definitely been a regression. It's kind of hard to tell with all of the `bh_chip` state not being encapsulated in a driver and kind of being spread out everywhere.

I also thought it was just generally known that infinite loops are bad in what is effectively driver code (i.e. code to enable BH).

cfriedt:
> also please fix "wiating" in the title

@alewycky-tenstorrent - fixed!

lib/tenstorrent/jtag_bootrom/jtag_bootrom.c
afongTT: @cfriedt just wanted to check if you wanted to change anything here

cfriedt: @alewycky-tenstorrent - that's a good suggestion. Sorry - I missed that originally. I'll add it. Thank you! 🙏

app/dmc/src/main.c
```diff
@@ -403,7 +403,11 @@ int main(void)
 				 */
 				bh_chip_cancel_bus_transfer_clear(chip);
 
-				jtag_bootrom_reset_asic(chip);
+				if (jtag_bootrom_reset_asic(chip) < 0) {
+					chip->data.performing_reset = false;
+					LOG_ERR("Failed to reset bh_chip %p", chip);
+					break;
```
danieldegrasse, line 409: Do we want to break out here, or just return from `main` (effectively to an idle loop)? We should also enable CONFIG_LOG if we want to be able to timeout here. Otherwise, it is going to be challenging to figure out where a BMFW failed- currently you can just attach with a debugger and see that the FW is hung here.

cfriedt: @danieldegrasse  I originally considered returning from main, but then our firmware would effectively do nothing in response to stimuli.

At least by using a break in the PERST handler, we can go on to run the PGOOD handler, etc. I.e. I think it's better for the firmware to be able to respond _somehow_ than to be in an infinite loop or idle and unresponsive.

danieldegrasse: Sure, that makes sense. Break is ok with me, so long as we turn on CONFIG_LOG.

cfriedt: Yikes - enabling logging adds 6k to the binary size. That seems excessive.
```
53248 is < the binary size 59052
```

PR #259, opened 2025-05-26
drivers: i2c: design ware i2c: removing i2c driver files 

mshimozatoTT:
removed lib/tenstorrent/bh_arc/dw_apb_i2c.c and lib/tenstorrent/bh_arc/dw_apb_i2c.h

upstream zephyr files are already in place


cfriedt:
I see now why this can't simply be merged. Documented in #261. Thanks for looking into it.

mshimozatoTT:
> I see now why this can't simply be merged. Documented in #261. Thanks for looking into it.

PR #274, opened 2025-05-29
lib: telemetry: reduce tag values to comply with pyluwen tag parsing

danieldegrasse:
pyluwen parses tag indices as a u8 on blackhole, and each tag occupies 32 bits. When the tag index is converted into an offset by multiplying by 4, we hit an effective limit of 62 tag values (max tag value of 62), as otherwise the pyluwen tag parsing code hits an overflow when slicing from the tag data array (63 * 4 + 4 = 256).

To correct this, reduce tag values for asic IDs to fit within the boundary. Also, properly populate the tag count within the telemetry block.


kikkiakki (approved):

cfriedt:
> I guess this is ok if no one has started using these tags yet (60/61/62), but ideally we wouldn't change them after publishing.

Agreed - we should consider these tags fixed after the next release.

PR #275, opened 2025-05-29
bh_arc: add support for tensix reset message

afongTT:
Tensix resets are needed for internal testing purposes. CMFW needs to know when the Tensix reset is happening, since this also resets the NOC. If NOC traffic from CMFW is in flight while NOC gets reset, this can cause CMFW's ARC processor to hang. Having CMFW do the Tensix reset is the simplest way of solving this.

Also adds an SMC e2e test to repeatedly issue the Tensix reset.


app/smc/pytest/e2e_smoke.py
```diff
@@ -236,3 +239,69 @@ def args():
...
+    SOFT_RESET_ADDR = 0xFFB121B0
+    SOFT_RESET_DATA = 0x47800
+
+    # Force AICLK to safe frequency
+    arc_chip.arc_msg(ARC_MSG_TYPE_FORCE_AICLK, arg0=250, arg1=0)
+    # Clear RISC reset registers
+    for addr in TENSIX_RISC_RESET_ADDR:
+        arc_chip.axi_write32(addr, 0)
+    for addr in TENSIX_RISC_RESET_ADDR:
+        print(f"{addr:#010x} = {arc_chip.axi_read32(addr):#010x}")
```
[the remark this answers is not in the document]
afongTT: Fixed, this print was not supposed to be left in.

lib/tenstorrent/bh_arc/init.c
```diff
@@ -361,6 +361,26 @@ static void EthInit(void)
 }
 
 #ifndef CONFIG_TT_SMC_RECOVERY
+static uint8_t ToggleTensixReset(uint32_t msg_code, const struct request *req, struct response *rsp)
```
cfriedt, line 364: We do have a number of reset devices for the tensixes already.

https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/drivers/reset/reset_tt_bh.c
https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/dts/arc/tenstorrent/tt_blackhole_smc.dtsi#L178
https://docs.zephyrproject.org/latest/doxygen/html/group__reset__controller__interface.html

It seems the only shortcoming with the Zephyr API is that you can only reset one line at a time.

Can you make a follow-up task to patch the Zephyr API to support resetting via a mask as well? No rush on that, but it would be nice to use one driver.

afongTT: I'm confused who implements the driver for that compatible?
`tenstorrent,blackhole-reset`

I'll create a JIRA to use the reset controller for Tensix in general.

cfriedt: i wrote it a long time ago

PR #281, opened 2025-05-29
Drosen/18.3.1 bump

DRvader:
Bump fw versions for smc and dmc to 0.12.1 and 0.6.1 respectively and release fw version 18.3.1


danieldegrasse:
For version bumps on backport branches (which this is, if I'm understanding right) do we want to overload the PATCHLEVEL field like this? right now the patch level is only ever set to 99 (in between releases) or 0 (for a release). I think we either need to move to using the TWEAK field for the 99/0 flip, or use TWEAK for backport version bumps like this. @cfriedt, any thoughts?

the core issue I see here is that once we make this release, what will we set the PATCHLEVEL field back to? 99 like it was before? Then the version of a change before the release and after it would be the same

cfriedt:
Patch level is correct. After the 18.3.1 release here, we have to change the patch level back to 99

PR #284, opened 2025-05-30
drivers: Add AVS driver

alayander:
Zephyrize library AVS driver


alayander:
To be revisited after a proper PLL driver exists for initialization dependency. Will also modify the implementation to be an MFD and utilize upstream driver APIs

PR #285, opened 2025-05-30
lib: tenstorrent: fan_ctrl: disable initial fan spin-up

kikkiakki:
Disable fan spin-up and set initial fan speed to 25% on active-cooled boards. This is to address customer concerns about the fan spin-up being loud.

### DIFF
--- a/app/dmc/src/main.c
+++ b/app/dmc/src/main.c
@@ -259,7 +259,7 @@ int main(void)
 
 	if (IS_ENABLED(CONFIG_TT_FAN_CTRL)) {
 		ret = init_fan();
-		set_fan_speed(100); /* Set fan speed to 100 by default */
+		set_fan_speed(25); /* Start fan speed at 25% */
 		if (ret != 0) {
 			LOG_ERR("%s() failed: %d", "init_fan", ret);
 			return ret;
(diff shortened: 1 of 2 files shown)


afongTT:
Should we start at 35% since that's the floor of our fan curve?

afongTT (approved):

cfriedt (approved):

cfriedt:
There was a CI issue with p150a earlier. I'll check on it again shortly.

cfriedt:
I would guess that this is related. @kikkiakki - can you comment on the tt-kmd issue? If this change fixes that but did not make it into 18.4.0, then the issue will be fixed in the next release.

https://github.com/tenstorrent/tt-kmd/issues/83

kikkiakki:
@cfriedt Fan speed being stuck at 100% but reporting 0 RPM really sounds like a communication problem with the STM32, is there a system we have that reproduces this issue?

If that is the case, this change will fix the fan being loud but fan RPM will probably still report as 0 and a therm trip might happen because there won't be any fan speed updates received.

cfriedt:
@kikkiakki - please comment on the other issue. I don't have any information myself.

PR #286, opened 2025-06-02
blobs: update wormhole fw blob

afongTT:
CMFW 2.33.1.0
- Fix telemetry tag conflict with BH CMFW


cfriedt:
Created a follow-up issue
https://github.com/tenstorrent/tt-zephyr-platforms/issues/289

PR #294, opened 2025-06-03
Track firmware loading time for PCIe in CI

nxuTT:
Track firmware loading time for PCIe in CI


nxuTT:
> I'll have another look later, but I'm confused about why there are so many changes all over the codebase.
> 
> Is it possible to better encapsulate data rather than having state and control in so many places?

You'll find out once you review the code. I don't like it in DMFW that we use number in addresses, so I fixed all of that. And basically putting a register define in include/tenstorrent that is shared between cmfw and dmfw. Because of that, I need to change all the includes in cmfw.

nxuTT:
> I'm concerned that the extra jtag read/writes would be extending the total time because the bootrom would be executing during that jtag read and write. Maybe it would be better to take the total time from reset signal received to soft reset deasserted from the dmc timer and then report that via the state info over i2c to the cmfw.

But you can't really track it in CI, because CI only does smi reset, it doesn't do sudo reboot. The time from AISC reset to PCIE FW loading complete is 40ms in sudo reboot case, but is 60ms in tt-smi reset case. And the smi reset case is not what I'm concerned about. I also made sure that by adding this jtag write/read, it doesn't increase the PCIe FW loading time.

nxuTT:
> Can you split it into two value?
> 
> 1. Time from signal received to execution start
> 2. Total time of execution
>    That should give a better idea, as most of the time should be spent in the second value and the first can in theory be mostly constant for each case (smi reset, reboot).

This PR will get you 2. For 1 since you say that 1 should be constant for smi reset and reboot then why would you care?

danieldegrasse:
> But you can't really track it in CI, because CI only does smi reset, it doesn't do sudo reboot. The time from AISC reset to PCIE FW loading complete is 40ms in sudo reboot case, but is 60ms in tt-smi reset case. And the smi reset case is not what I'm concerned about. I also made sure that by adding this jtag write/read, it doesn't increase the PCIe FW loading time.

I *think* we could track the full time from PERST (or something close to it) by resetting the DMC with the `dmc_reset.py` script. If the DMC also reboots, we can write the DMC timer's timestamp to a scratch register when resetting the ARC, then add to that value once we complete the PCIe FW load. Since the `dmc_reset.py` script will toggle the DMC RST pin, it should be the same measurement as PERST, right?

That being said this PR is an improvement as is, so I'm more than happy to move forwards with it right now.

nxuTT:
> > But you can't really track it in CI, because CI only does smi reset, it doesn't do sudo reboot. The time from AISC reset to PCIE FW loading complete is 40ms in sudo reboot case, but is 60ms in tt-smi reset case. And the smi reset case is not what I'm concerned about. I also made sure that by adding this jtag write/read, it doesn't increase the PCIe FW loading time.
> 
> I _think_ we could track the full time from PERST (or something close to it) by resetting the DMC with the `dmc_reset.py` script. If the DMC also reboots, we can write the DMC timer's timestamp to a scratch register when resetting the ARC, then add to that value once we complete the PCIe FW load. Since the `dmc_reset.py` script will toggle the DMC RST pin, it should be the same measurement as PERST, right?
> 
> That being said this PR is an improvement as is, so I'm more than happy to move forwards with it right now.

Agree, if we implement what @TTDRosen suggested using DMC timer to track the exact time from PERST signal received to execution start. But for the smi reset or DMC reset case, we need to be careful because they don't share the same code path as sudo reboot. PERST only got asserted when sudo reboot/cold boot occurs. The way we handle PERST is through interrupt handler but tt-smi reset is a message based implementation and no interrupt involved. We could still trace in the code and see where we could approximately get the similar enough duration.

lib/tenstorrent/bh_arc/gpio.c
```diff
@@ -8,9 +8,10 @@
 
 #include "reg.h"
 
+#include <tenstorrent/control_reg.h>
```
cfriedt, line 11: `RESET_UNIT_GPIO_PAD_TRIEN_CNTL_REG_ADDR` is mostly a detail that should be encapsulated inside of the gpio neighbourhood (or possibly pinctrl). I could see maybe naming this file `zephyr/drivers/gpio/gpio_tt_bh.h` to follow the pattern that exists already with [most Zephyr GPIO drivers](https://github.com/zephyrproject-rtos/zephyr/tree/main/include/zephyr/drivers/gpio).

For reference #265

 Is this register location specific to Blackhole or all Tenstorrent products?

lib/tenstorrent/jtag_bootrom/jtag_bootrom.c
```diff
@@ -4,13 +4,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#include "blackhole_offsets.h"
-
 #include <stdint.h>
 
 #include <tenstorrent/bh_chip.h>
+#include <tenstorrent/control_reg.h>
```
cfriedt, line 10: Could we maybe just copy the register definitions into the jtag workaround, since it's (hopefully) a 1-off thing? Otherwise, it's a bit unusual to share platform-specific headers between two different FW domains.

include/tenstorrent/control_reg.h
cfriedt, line 1: https://github.com/tenstorrent/tt-zephyr-platforms/pull/294/files#r2127020981

include/tenstorrent/status_reg.h
```diff
@@ -4,11 +4,13 @@
...
 
 /* Scratch registers used for status and error reporting */
-#ifndef STATUS_REG_H
-#define STATUS_REG_H
+#ifndef TESNTORRENT_STATUS_REG_H_
+#define TENSTORRENT_STATUS_REG_H_
 
 #include <stdint.h>
 
+#define RESET_UNIT_REFCLK_CNT_LO_REG_ADDR 0x800300E0
```
cfriedt, line 12: This isn't exactly a status register. It belongs to the timer, so by moving this here we are leaking between subsystems or drivers and adding tech debt.

include/tenstorrent/status_reg.h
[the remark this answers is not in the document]
cfriedt: Sure - that works.

lib/tenstorrent/bh_arc/status_reg.h
```diff
@@ -35,12 +35,15 @@
...
+#define DMFW_INIT_DONE_REG_ADDR              RESET_UNIT_SCRATCH_RAM_REG_ADDR(16)
+#define PERST_TO_DMFW_INIT_DONE_REG_ADDR     RESET_UNIT_SCRATCH_RAM_REG_ADDR(17)
 #define I2C0_TARGET_DEBUG_STATE_REG_ADDR     RESET_UNIT_SCRATCH_RAM_REG_ADDR(19)
 #define I2C0_TARGET_DEBUG_STATE_2_REG_ADDR   RESET_UNIT_SCRATCH_RAM_REG_ADDR(20)
 
-#define STATUS_FW_VUART_REG_ADDR(n)          RESET_UNIT_SCRATCH_RAM_REG_ADDR(40 + (n))
+#define STATUS_FW_VUART_REG_ADDR(n) RESET_UNIT_SCRATCH_RAM_REG_ADDR(40 + (n))
 /* SCRATCH_RAM_40 - SCRATCH_RAM_41 reserved for virtual uarts */
-#define STATUS_FW_SCRATCH_REG_ADDR           RESET_UNIT_SCRATCH_RAM_REG_ADDR(63)
+#define STATUS_FW_SCRATCH_REG_ADDR  RESET_UNIT_SCRATCH_RAM_REG_ADDR(63)
```
alewycky-tenstorrent, line 46: Revert this bit.

include/tenstorrent/bh_arc.h
```diff
@@ -28,6 +28,8 @@ typedef struct dmStaticInfo {
 	uint32_t version;
 	uint32_t bl_version;
 	uint32_t app_version;
+	uint32_t dm_init_cmpl_ts;
+	uint32_t dm_init_duration;
```
alewycky-tenstorrent, line 32: Please add a comment, at least say units.

include/tenstorrent/tt_smbus_regs.h
```diff
@@ -19,7 +19,7 @@ enum CMFWSMBusReg {
 	/* WO, 16 bits. Write with sequence number and message ID to ack cm2dmMessage */
 	CMFW_SMBUS_ACK = 0x11,
 	/* WO, 96 bits. Write with dmStaticInfo struct including DMFW version */
```
alewycky-tenstorrent, line 21: I don't like that the comment contains the size, but I think the size is wrong now.

include/tenstorrent/bh_arc.h
```diff
@@ -28,6 +28,8 @@ typedef struct dmStaticInfo {
 	uint32_t version;
 	uint32_t bl_version;
 	uint32_t app_version;
+	uint32_t dm_init_cmpl_ts;
```
alewycky-tenstorrent, line 31: How about arc_start_time? I prefer because it directly describes the event.

lib/tenstorrent/jtag_bootrom/jtag_bootrom.c
```diff
@@ -266,6 +289,14 @@ void jtag_bootrom_soft_reset_arc(struct bh_chip *chip)
 	/* Write reset_vector (rom_memory[0]) */
 	jtag_axi_write32(dev, ROM_MEMORY_MEM_BASE_ADDR, 0x84);
 
+	/* store DMC init done timestamp */
+	if (perst_start_time_valid) {
```
alewycky-tenstorrent, line 293: I don't know the point of this. Can jtag_bootrom_soft_reset_arc happen before gpio_asic_reset_callback?
Maybe change the bool to "perst_seen" because it's really about perst, not about some start time.

alewycky-tenstorrent:
Second commit title misspells "init"

alewycky-tenstorrent (approved):

nxuTT:
> Second commit title misspells "init"

Fixed :)

PR #295, opened 2025-06-04
drivers: clock_control: add tenstorrent blackhole clock control driver

alexapostolu:
Add Tenstorrent Blackhole PLL driver using Zephyr's clock control API. Integration of this driver with the rest of the code will be done in a follow up commit.

### DIFF
--- a/doc/release/release-notes-18.6.md
+++ b/doc/release/release-notes-18.6.md
@@ -62,6 +62,8 @@ Major enhancements with this release include:
 
 * add `maxim,max6639` driver + tests
 
+* add Zephyr PLL driver + tests
+
 [comment]: <> (UL Libraries)
 
 [comment]: <> (H2 Other Notable Changes, if applicable)


cfriedt:
@afongTT @danieldegrasse - the program perf issue where the duration exceeds some expected value indicates to me that one branch of the clock tree might not be set to the optimal frequency. IIRC, Aaron mentioned that there is a kind of ramp-up where we start at one frequency, and then configure the oll to a higher frequency for flash. Could that be what's missing here?

drivers/clock_control/clock_control_tt_bh.c
```diff
@@ -0,0 +1,726 @@
...
+
+typedef struct {
+	pll_cntl_1_reg pll_cntl_1;
+	pll_cntl_2_reg pll_cntl_2;
+	pll_cntl_3_reg pll_cntl_3;
+	pll_cntl_5_reg pll_cntl_5;
+	pll_use_postdiv_reg use_postdiv;
+} PLLSettings;
+
+static const PLLSettings pll_initial_settings[PLL_COUNT] = {
```
cfriedt, line 149: Can this be added via devicetree instead?

dts/arc/tenstorrent/tt_blackhole_smc.dtsi
```diff
@@ -95,6 +95,23 @@
...
+	ref_clk: ref_clk {
+		compatible = "fixed-clock";
+
+		clock-frequency = <1000000000>; /* 1 GHz */
+		#clock-cells = <0>;
+
+		status = "okay";
+	};
+
+	pll: pll@80020100 {
```
cfriedt, line 107: I think technically, there should be several instances of this. Can it be converted in a follow-up PR?

alexapostolu: I thought it would be better to have one PLL device since the init function depends on initializing all the PLLs together. I wasn't sure if I could separate the initialization for each PLL

cfriedt: In the `init()` function, if something has to be done on the first call to all of the PLLs, or some common registers, then the driver could probably have a static `bool common_init_complete;` that gets set after a common init function is done once. Something like

```cpp
static void clock_control_tt_bh_init_common(void)
{
  static bool complete;
  
  if (complete) {
    return;
  }
  
  /* common init here */
}
```

drivers/clock_control/clock_control_tt_bh.c
```diff
@@ -0,0 +1,624 @@
...
+#include <zephyr/sys_clock.h>
+#include <zephyr/sys/util.h>
+#include <stdint.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(clock_control_tt_bh);
+
+#define PLL_COUNT 5
+
+#define PLL_LOCK_TIMEOUT 400
```
cfriedt, line 23: Would be good to add units to this
```suggestion
#define PLL_LOCK_TIMEOUT_MS 400
```

drivers/clock_control/clock_control_tt_bh.c
```diff
@@ -0,0 +1,624 @@
...
+
+#define PLL_CNTL_0_OFFSET             0x00
+#define PLL_CNTL_1_OFFSET             0x04
+#define PLL_CNTL_2_OFFSET             0x08
+#define PLL_CNTL_3_OFFSET             0x0C
+#define PLL_CNTL_4_OFFSET             0x10
+#define PLL_CNTL_5_OFFSET             0x14
+#define PLL_CNTL_6_OFFSET             0x18
```
cfriedt, line 31: is `PLL_CNTL_7_OFFSET` missing?

tests/drivers/clock-control/src/main.c
```diff
@@ -0,0 +1,127 @@
...
+
+	ret = clock_control_set_rate(pll, gddr_subsys, (clock_control_subsys_rate_t)target_rate);
+	zassert_ok(ret, "set_rate for GDDR failed with %d", ret);
+
+	ret = clock_control_get_rate(pll, gddr_subsys, &new_rate);
+	zassert_ok(ret, "get_rate for GDDR failed with %d", ret);
+
+	zassert_within(new_rate, target_rate, new_rate / 100.f * CLOCK_RATE_TOLERANCE_PERCENT,
+			 "Expected ~%d Hz but got %d Hz", target_rate, new_rate);
+}
```
cfriedt, line 50: woops - I think this is the only other clang-format issue.

danieldegrasse (commented):
This isn't a blocking request (this driver is already a massive improvement 😄 ) but have we considered treating each of the 4 PLLs in the system as an independent instance? Ideally we could make the PLL driver pretty generic in that case, and only need to pass in the postdiv index when we were getting a rate for a specific clock (IE ARCCLK with postdiv index of 4 on PLL1)

alexapostolu:
The only thing I'm uncertain about is how to handle setting the rate for each clock? Since the old code had two different ways to set the rate for AI and GDDR Mem CLK, I'm not sure how to make that more generic for this driver

soc/tenstorrent/tt_blackhole/Kconfig.defconfig.tt_blackhole_smc
```diff
@@ -73,6 +73,9 @@ config 2ND_LVL_INTR_03_OFFSET
 config GPIO
 	default y
 
+config CLOCK_CONTROL
+	default y
```
cfriedt, line 77: I think we should not enable clock control for SMC until the subsequent integration PR.

```suggestion
```

test-conf/tests/drivers/clock-control/boards/tt_blackhole_tt_blackhole_smc.conf
cfriedt, line 1: Don't forget to use underscores instead of dashes.
`test-conf/tests/drivers/clock_control`

include/zephyr/drivers/clock-control/clock_control_tt_bh.h
cfriedt, line 1: `include/zephyr/drivers/clock_control`

soc/tenstorrent/tt_blackhole/Kconfig.defconfig.tt_blackhole_smc
cfriedt: Test it first, and if that's the case, then probably, yea.

tests/drivers/flash/performance/testcase.yaml
```diff
@@ -12,7 +12,7 @@ tests:
       # The below values are set experimentally. A test failure may
       # indicate a performance regression in the flash driver.
       - CONFIG_EXPECTED_READ_TIME=5
-      - CONFIG_EXPECTED_PROGRAM_TIME=400
+      - CONFIG_EXPECTED_PROGRAM_TIME=460
```
afongTT, line 15: @danieldegrasse do we expect to need to change this?

danieldegrasse: We shouldn't have to change it, no. @alexapostolu, you were seeing some cases where the programming time took longer than expected, right?

drivers/clock_control/clock_control_tt_bh.c
```diff
@@ -0,0 +1,622 @@
...
+	return 0;
+}
+
+static enum clock_control_status clock_control_tt_bh_get_status(const struct device *dev,
+								clock_control_subsys_t sys)
+{
+	return CLOCK_CONTROL_STATUS_UNKNOWN;
+}
+
+static int clock_control_tt_bh_set_rate(const struct device *dev, clock_control_subsys_t sys,
```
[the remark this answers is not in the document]
danieldegrasse: The clock_control API *doesn't* handle these cases. I have some *opinions* on the API (there is a reason I'm trying to rework it upstream). For the time being, I'd recommend some way of indicating to the driver "yes, you can change FBDIV to get my requested clock rate" versus "no, only touch the postdiv setting for this clock"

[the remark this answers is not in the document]
cfriedt: Would it make sense to work on that in a follow-up PR?

cfriedt (approved):

PR #298, opened 2025-06-05
boards: tenstorrent: encode a default board ID into flash images

danieldegrasse:
Encode a default board ID into the output flash image. This won't be present in the generated tt_boot_fs binary, but can be manually included with tt_boot_fs.py. This is useful for board recovery cases where the board ID has been erased.

### DIFF
--- a/boards/tenstorrent/tt_blackhole/spirom_data_tables/GALAXY/read_only.txt
+++ b/boards/tenstorrent/tt_blackhole/spirom_data_tables/GALAXY/read_only.txt
@@ -19,3 +19,4 @@ reference_values_table {
 }
 vendor_id: 0x1e52
 asic_location: 0
+board_id: 0x47100000000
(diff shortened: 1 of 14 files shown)


cfriedt (dismissed):

cfriedt:
Would be good to get another approval here.

TTDRosen:
Looks good as well, my only slight concern (which I'll deferr) to Aaron on is if we care about the board rev anywhere in BH

DRvader:
Looks good as well, my only slight concern (which I'll deferr) to Aaron on is if we care about the board rev anywhere in BH

afongTT:
> Looks good as well, my only slight concern (which I'll deferr) to Aaron on is if we care about the board rev anywhere in BH

True, let's just make everything board rev 1. e.g.

```diff
- board_id: 0x47000000000
+ board_id: 0x47100000000
```

danieldegrasse:
> > Looks good as well, my only slight concern (which I'll deferr) to Aaron on is if we care about the board rev anywhere in BH
> 
> True, let's just make everything board rev 1. e.g.
> 
> ```diff
> - board_id: 0x47000000000
> + board_id: 0x47100000000
> ```

Updated

afongTT (approved):

cfriedt (approved):

PR #299, opened 2025-06-06
lib: tenstorrent: add a print if PGOOD isn't high during boot

danieldegrasse:
Add a print statement if PGOOD isn't high during boot. This should help diagnose cases where a card isn't enumerating because PGOOD has gone low, since CI will dump the DMC RTT logs.


danieldegrasse:
Closing and reopening this to restart CI. I still think this print would be useful to add

cfriedt (approved):

PR #301, opened 2025-06-06
zephyr: blobs: Update ERISC FW to v1.4.1

bingliTT:
- Fixed a bug in ETH training for P300 that would cause ETH tests to fail to train on certain ports correctly


afongTT (approved):

cfriedt:
I reset the CI machines - it seemed like things were hung

PR #303, opened 2025-06-06
drivers: fwtable: integrating fwtable as psuedo-device

mshimozatoTT:
Add Tenstorrent Blackhole FW Table binding.


danieldegrasse:
Personally I think the fwtable makes more sense as a subsystem rather than a device- is there a specific reason we'd use the device model here? It is typically only needed for psuedo-devices when you might have multiple instances of the device (which won't be the case for fwtables)

cfriedt:
> Personally I think the fwtable makes more sense as a subsystem rather than a device- is there a specific reason we'd use the device model here? It is typically only needed for psuedo-devices when you might have multiple instances of the device (which won't be the case for fwtables)

Yeah, that was my suggestion. 

So normally, we have the settings subsystem in Zephyr for stuff like this. The main difference is that that is typically involved in software-level configuration.

I think for the high-speed I/O devices, where we need them to be initialized very deterministically, having the strongly-ordered implicit and explicit dependencies of DT + built-in priorities would be really helpful to hit those guarantees. With the device model, pcie, ethernet, gddr I believe all need to pull some read-only config out of SPI.

If the subsystem is being initialized, there is a less strongly-ordered dependency information, and probably more room for error.

In the end, they all use the same init system, but I personally just like the added security of the device model.

cfriedt:
I have a suspicion that this might depend on the PLL as well.

app/smc/CMakeLists.txt
```diff
@@ -6,5 +6,5 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 project(app LANGUAGES C)
 
 target_link_libraries(app PRIVATE bh_arc)
+target_link_libraries(app PRIVATE bh_fwtable)
```
cfriedt, line 9: Neither of these lines should be necessary

CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 zephyr_include_directories(include)
+
+zephyr_include_directories( $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/modules/tt-zephyr-platforms/drivers/misc/bh_fwtable>)
```
cfriedt, line 5: This line shouldn't be necessary either.

app/smc/CMakeLists.txt
```diff
@@ -6,5 +6,5 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
...
 target_sources(app PRIVATE src/main.c)
-target_include_directories(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../../lib/tenstorrent/bh_arc)
```
cfriedt, line 10: This might still be necessary

app/smc/prj.conf
```diff
@@ -38,3 +38,5 @@ CONFIG_BOOT_BANNER=n
 CONFIG_TT_BOOT_BANNER=y
 CONFIG_TT_BOOT_BANNER_GIT_VERSION=y
 CONFIG_TT_BOOT_BANNER_SDK_VERSION=y
+
+CONFIG_BH_FWTABLE=y
```
cfriedt, line 42: This shouldn't be necessary

drivers/misc/bh_fwtable/Kconfig
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2024 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+config BH_FWTABLE
+	bool "Tenstorrent Blackhole Firmware Table"
+	default y
+	depends on DT_HAS_TENSTORRENT_BH_FWTABLE_ENABLED
+	depends on TT_BOOT_FS
```
cfriedt, line 8: We might be able to use `select TT_BOOT_FS` here, which would be a bit better, since people using the fwtable don't particularly need to care that there is a "filesystem" used to store the tables.

include/zephyr/drivers/misc/bh_fwtable.h
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright (c) 2025 Tenstorrent AI ULC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef BH_FWTABLE_H
+#define BH_FWTABLE_H
+
+#include <stdint.h>
+#include <zephyr/device.h>
+
```
cfriedt, line 11:
```suggestion
#ifdef __cplusplus
extern "C" {
#endif
```

include/zephyr/drivers/misc/bh_fwtable.h
```diff
@@ -0,0 +1,36 @@
...
+	PcbTypeP100 = 1,
+	PcbTypeP150 = 2,
+	PcbTypeP300 = 3,
+	PcbTypeUBB = 4,
+	PcbTypeUnknown = 0xFF
+} PcbType;
+
+PcbType tt_bh_fwtable_get_pcb_type(const struct device *dev);
+uint32_t tt_bh_fwtable_get_asic_location(const struct device *dev);
+
```
cfriedt, line 35:
```suggestion

#ifdef __cplusplus
}
#endif

```

lib/tenstorrent/bh_arc/CMakeLists.txt
```diff
@@ -58,23 +55,7 @@ if(CONFIG_ARC)
...
-set(SPIROM_PROTOBUFS ${ZEPHYR_CURRENT_MODULE_DIR}/lib/tenstorrent/bh_arc/spirom_protobufs)
-
-message(STATUS "Generating Python files from protobufs")
-set(OUTPUT_DIR ${ZEPHYR_BINARY_DIR}/python_proto_files)
-file(MAKE_DIRECTORY ${OUTPUT_DIR})
-set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
-	COMMAND ${PROTOC} --python_out=${OUTPUT_DIR} ${SPIROM_PROTOBUFS}/fw_table.proto -I ${SPIROM_PROTOBUFS}
-	COMMAND ${PROTOC} --python_out=${OUTPUT_DIR} ${SPIROM_PROTOBUFS}/read_only.proto -I ${SPIROM_PROTOBUFS}
-	COMMAND ${PROTOC} --python_out=${OUTPUT_DIR} ${SPIROM_PROTOBUFS}/flash_info.proto -I ${SPIROM_PROTOBUFS}
-)
```
cfriedt, line 77: I think we still need this bit

lib/tenstorrent/bh_arc/init.c
```diff
@@ -421,13 +422,14 @@ static int InitHW(void)
...
-		load_flash_info_table(large_sram_buffer, SCRATCHPAD_SIZE);
-	}
+	/* if (!IS_ENABLED(CONFIG_TT_SMC_RECOVERY)) {
+	 * load_fw_table(large_sram_buffer, SCRATCHPAD_SIZE);
+	 * }
+	 * load_read_only_table(large_sram_buffer, SCRATCHPAD_SIZE);
+	 * if (!IS_ENABLED(CONFIG_TT_SMC_RECOVERY)) {
+	 * load_flash_info_table(large_sram_buffer, SCRATCHPAD_SIZE);
+	 * }
+	 */
```
cfriedt, line 432: I think we can remove this - except we should preserve the behaviour of loading only the read-only table when `CONFIG_TT_SMC_RECOVERY` is set in `bh_fwtable.c::init()`

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -46,72 +47,77 @@ struct telemetry_table {
...
-		[48] = {TAG_MAX_GDDR_TEMP, TELEM_OFFSET(TAG_MAX_GDDR_TEMP)},
-		[49] = {TAG_ASIC_LOCATION, TELEM_OFFSET(TAG_ASIC_LOCATION)},
-		[50] = {TAG_BOARD_POWER_LIMIT, TELEM_OFFSET(TAG_BOARD_POWER_LIMIT)},
-		[51] = {TAG_INPUT_POWER, TELEM_OFFSET(TAG_INPUT_POWER)},
-		[52] = {TAG_ASIC_ID_HIGH, TELEM_OFFSET(TAG_ASIC_ID_HIGH)},
-		[53] = {TAG_ASIC_ID_LOW, TELEM_OFFSET(TAG_ASIC_ID_LOW)},
-		[54] = {TAG_THERM_TRIP_COUNT, TELEM_OFFSET(TAG_THERM_TRIP_COUNT)},
-		[55] = {TAG_TELEM_ENUM_COUNT, TELEM_OFFSET(TAG_TELEM_ENUM_COUNT)},
-	},
+		.tag_table = {
```
cfriedt, line 50: It looks like most of this section could be left unchanged.

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -80,64 +83,51 @@ typedef struct {
 
 static Throttler throttler[kThrottlerCount] = {
 	[kThrottlerTDP] = {
-
```
cfriedt, line 83: looks like all of these formatting changes were unnecessary as well.

lib/tenstorrent/bh_arc/CMakeLists.txt
mshimozatoTT: Since we would eventually be removing the library drivers, I thought it might be helpful to move the protobuf functionality to the bh_fwtable cmake file ?

app/smc/CMakeLists.txt
mshimozatoTT: without it, app/smc/src/main.c is unable to locate the protobuf header files once they are generated by bh_fwtable

lib/tenstorrent/bh_arc/CMakeLists.txt
cfriedt: Sorry 67-77

cfriedt: But if it works to remove them - that's better as long as nothing is broken. I thought they might still be needed for something though.

boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc.dts
```diff
@@ -19,7 +19,6 @@
 		zephyr,console = &uart0;
 		zephyr,shell-uart = &uart0;
 	};
-
```
cfriedt, line 22: remove this change.

dts/bindings/misc/tenstorrent,bh-fwtable.yaml
```diff
@@ -0,0 +1,17 @@
...
+include: base.yaml
+
+properties:
+
+  flash_dev:
+    type: phandle
+    required: true
+    description: |
+      Phandle to the flash device the firmware table resides in. If this property is
+      not specified, the firmware table will not be read from SPI flash.
```
cfriedt, line 17: This is a TODO item - probably best handled in a follow-up. It means that the `flash_dev` property is not absolutely required, but we can revisit that after the PR is in.

lib/tenstorrent/bh_arc/aiclk_ppm.c
```diff
@@ -31,6 +31,8 @@ typedef enum {
 
 AiclkPPM aiclk_ppm;
```
cfriedt, line 32: This should probably be static.

lib/tenstorrent/bh_arc/init.c
```diff
@@ -81,7 +81,7 @@ static void AssertSoftResets(void)
 				GetGddrNocCoords(gddr_inst, noc_node_inst, kNocRing, &x, &y);
 				NOC2AXITlbSetup(kNocRing, kNocTlb, x, y, kSoftReset0Addr);
 				NOC2AXIWrite32(kNocRing, kNocTlb, kSoftReset0Addr,
-					kAllRiscSoftReset);
```
cfriedt, line 84: This change can be removed.

lib/tenstorrent/bh_arc/init.c
```diff
@@ -135,8 +135,7 @@ static int CheckGddrTraining(uint8_t gddr_inst, k_timepoint_t timeout)
 			MRISC_INIT_TIMEOUT, gddr_inst, post_code);
 		return -ETIMEDOUT;
 	}
-	LOG_ERR("GDDR instance %d failed to initialize. Post code: 0x%x\n",
-		gddr_inst, post_code);
+	LOG_ERR("GDDR instance %d failed to initialize. Post code: 0x%x\n", gddr_inst, post_code);
```
cfriedt, line 138: This change can be removed.

lib/tenstorrent/bh_arc/init.c
```diff
@@ -353,8 +354,8 @@ static void EthInit(void)
 
 	for (uint8_t eth_inst = 0; eth_inst < MAX_ETH_INSTANCES; eth_inst++) {
 		if (tile_enable.eth_enabled & BIT(eth_inst)) {
-			LoadEthFwCfg(eth_inst, ring, tile_enable.eth_enabled,
-				large_sram_buffer, fw_size);
+			LoadEthFwCfg(eth_inst, ring, tile_enable.eth_enabled, large_sram_buffer,
+				     fw_size);
```
cfriedt, line 358: This change can be removed.

lib/tenstorrent/bh_arc/pcie.c
```diff
@@ -37,10 +36,12 @@
...
 #define PCIE_SII_A_REG_MAP_BASE_ADDR    0xFFFFFFFFF0000000LL
 
-#define PCIE_SII_A_NOC_TLB_DATA_62__REG_OFFSET       0x0000022C
-#define PCIE_SII_A_NOC_TLB_DATA_0__REG_OFFSET        0x00000134
-#define PCIE_SII_A_APP_PCIE_CTL_REG_OFFSET           0x0000005C
-#define PCIE_SII_A_LTSSM_STATE_REG_OFFSET            0x00000128
+#define PCIE_SII_A_NOC_TLB_DATA_62__REG_OFFSET 0x0000022C
+#define PCIE_SII_A_NOC_TLB_DATA_0__REG_OFFSET  0x00000134
+#define PCIE_SII_A_APP_PCIE_CTL_REG_OFFSET     0x0000005C
+#define PCIE_SII_A_LTSSM_STATE_REG_OFFSET      0x00000128
```
cfriedt, line 42: This change can be removed.

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -46,72 +47,77 @@ struct telemetry_table {
...
+			},
 };
 static uint32_t *telemetry = &telemetry_table.telemetry[0];
 
 static struct k_timer telem_update_timer;
 static struct k_work telem_update_worker;
 static int telem_update_interval = 100;
 
-
-
```
cfriedt, line 114: This change can be removed

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -230,15 +236,17 @@ int GetMaxGDDRTemp(void)
...
 static void write_static_telemetry(uint32_t app_version)
 {
-	telemetry_table.version = TELEMETRY_VERSION;    /* v0.1.0 - Only update when redefining the
-							 * meaning of an existing tag
-							 */
-	telemetry_table.entry_count = TAG_COUNT; /* Runtime count of telemetry entries */
+	telemetry_table.version = TELEMETRY_VERSION; /* v0.1.0 - Only update when redefining the
+						      * meaning of an existing tag
+						      */
+	telemetry_table.entry_count = TAG_COUNT;     /* Runtime count of telemetry entries */
```
cfriedt, line 242: This change can be removed.

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -275,13 +284,12 @@ static void write_static_telemetry(uint32_t app_version)
 	telemetry[TAG_ENABLED_GDDR] = tile_enable.gddr_enabled;
 	telemetry[TAG_ENABLED_L2CPU] = tile_enable.l2cpu_enabled;
 	telemetry[TAG_PCIE_USAGE] =
-		((tile_enable.pcie_usage[1] & 0x3) << 2) |
-		(tile_enable.pcie_usage[0] & 0x3);
+		((tile_enable.pcie_usage[1] & 0x3) << 2) | (tile_enable.pcie_usage[0] & 0x3);
```
cfriedt, line 287: This change can be removed.

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -302,18 +310,20 @@ static void update_telemetry(void)
...
-	telemetry[TAG_VDD_LIMITS] = 0x00000000;      /* VDD limits - Not Available yet */
-	telemetry[TAG_THM_LIMITS] = 0x00000000;      /* THM limits - Not Available yet */
+	telemetry[TAG_TDP] =
+		telemetry_internal_data
+			.vcore_power; /* reported in W, will be truncated to uint32_t */
+	telemetry[TAG_TDC] =
+		telemetry_internal_data
+			.vcore_current;         /* reported in A, will be truncated to uint32_t */
+	telemetry[TAG_VDD_LIMITS] = 0x00000000; /* VDD limits - Not Available yet */
+	telemetry[TAG_THM_LIMITS] = 0x00000000; /* THM limits - Not Available yet */
```
cfriedt, line 320: This change can be removed.

app/smc/CMakeLists.txt
cfriedt: Could add `target_include_directories(app PRIVATE path/to/generated/headers)` as well.

lib/tenstorrent/bh_arc/CMakeLists.txt
cfriedt: Let's keep it here, since it's not required for the `bh_fwtable` driver.

drivers/misc/bh_fwtable/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+set(SPIROM_PROTOBUFS ${ZEPHYR_CURRENT_MODULE_DIR}/lib/tenstorrent/bh_arc/spirom_protobufs)
+
+message(STATUS "Generating Python files from protobufs")
+set(OUTPUT_DIR ${ZEPHYR_BINARY_DIR}/python_proto_files)
+file(MAKE_DIRECTORY ${OUTPUT_DIR})
+set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
+	COMMAND ${PROTOC} --python_out=${OUTPUT_DIR} ${SPIROM_PROTOBUFS}/fw_table.proto -I ${SPIROM_PROTOBUFS}
+	COMMAND ${PROTOC} --python_out=${OUTPUT_DIR} ${SPIROM_PROTOBUFS}/read_only.proto -I ${SPIROM_PROTOBUFS}
+	COMMAND ${PROTOC} --python_out=${OUTPUT_DIR} ${SPIROM_PROTOBUFS}/flash_info.proto -I ${SPIROM_PROTOBUFS}
+)
```
cfriedt, line 34: This shouldn't be necessary for the `bh_fwtable` driver, so I would just lines 24-34 where they were originally.

In the long run, they should probably go into `boards/tenstorrent/tt_blackhole`, but that doesn't need to be done in this PR.

app/smc/CMakeLists.txt
```diff
@@ -6,5 +6,6 @@ find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
 project(app LANGUAGES C)
 
 target_link_libraries(app PRIVATE bh_arc)
+target_include_directories(app PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../../drivers/misc/bh_fwtable)
+target_link_libraries(app PRIVATE bh_fwtable)
```
cfriedt, line 10:
```suggestion
target_include_directories(app PRIVATE ${ZEPHYR_TT_ZEPHYR_PLATFORMS_MODULE_DIR}/lib/tenstorrent/bh_arc)
target_include_directories(app PRIVATE ${CMAKE_BINARY_DIR}/modules/tt-zephyr-platforms/drivers/misc/bh_fwtable)

```

tests/drivers/flash/performance/prj.conf
```diff
@@ -4,3 +4,4 @@ CONFIG_FLASH=y
 CONFIG_FLASH_PAGE_LAYOUT=y
 CONFIG_MAIN_STACK_SIZE=2048
 CONFIG_ZTEST_STACK_SIZE=4096
+CONFIG_STD_C11=y
```
cfriedt, line 7:
```suggestion

# no need to build the fwtable library for this test
CONFIG_BH_FWTABLE=n
```

cfriedt:
@mshimozatoTT - one thing you could do in this PR, much like what others have done, is split it into 2 - one pr for implementation and a follow-up pr for integration.

The first pr just needs to have a test that can be built and makes some reasonable checks.

mshimozatoTT:
> @mshimozatoTT - one thing you could do in this PR, much like what others have done, is split it into 2 - one pr for implementation and a follow-up pr for integration.
> 
> The first pr just needs to have a test that can be built and makes some reasonable checks.

ok, I'll try doing that today!

lib/tenstorrent/bh_arc/CMakeLists.txt
```diff
@@ -80,6 +68,8 @@ set_property(GLOBAL APPEND PROPERTY extra_post_build_commands
 	COMMAND ${PROTOC} --python_out=${OUTPUT_DIR} ${SPIROM_PROTOBUFS}/flash_info.proto -I ${SPIROM_PROTOBUFS}
 )
 
+# target_link_libraries(bh_arc PRIVATE bh_fwtable)
```
cfriedt, line 71: Probably safe to remove this comment.

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -45,22 +50,65 @@ struct bh_fwtable_data {
...
 const FwTable *tt_bh_fwtable_get_fw_table(const struct device *dev)
 {
 	struct bh_fwtable_data *data = dev->data;
 
+	/* Load tables on first access */
+	if (!data->initialized) {
+		if (tt_bh_fwtable_load_tables(dev) != 0) {
+			return NULL;
+		}
+	}
```
cfriedt, line 66: This is a little bit racey. It should be OK if you want to synchronize with a semaphore though.

To reduce duplication, I would suggest putting this chunk directly inside `tt_bh_fwtable_load_tables()`, and then each function can simply call `tt_bh_fwtable_load_tables()` without any external checks.
```suggestion
	SYS_SEM_LOCK(&data->lock) {
		/* Load tables on first access */
		if (!data->initialized) {
			if (tt_bh_fwtable_load_tables(dev) != 0) {
				return NULL;
			}
		}
	}
```

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -45,22 +50,65 @@ struct bh_fwtable_data {
...
+const ReadOnly *tt_bh_fwtable_get_read_only_table(const struct device *dev)
+{
+	struct bh_fwtable_data *data = dev->data;
+
+	/* Load tables on first access */
+	if (!data->initialized) {
+		if (tt_bh_fwtable_load_tables(dev) != 0) {
+			return NULL;
+		}
+	}
```
cfriedt, line 94:
```suggestion
	if(tt_bh_fwtable_load_tables(dev) != 0) {
		return NULL;
	}
```

drivers/misc/bh_fwtable/bh_fwtable.c
cfriedt: After the above logic has been added to `tt_bh_fwtable_load_tables()`, this section is a bit simpler.
```suggestion
	if(tt_bh_fwtable_load_tables(dev) != 0) {
		return NULL;
	}
```

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -45,22 +50,65 @@ struct bh_fwtable_data {
...
 {
 	PcbType pcb_type;
 	struct bh_fwtable_data *data = dev->data;
 
+	/* Load tables on first access */
+	if (!data->initialized) {
+		if (tt_bh_fwtable_load_tables(dev) != 0) {
+			return PcbTypeUnknown;
+		}
+	}
```
cfriedt, line 110:
```suggestion
	if(tt_bh_fwtable_load_tables(dev) != 0) {
		return NULL;
	}
```

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -96,11 +144,29 @@ PcbType tt_bh_fwtable_get_pcb_type(const struct device *dev)
...
 {
 	struct bh_fwtable_data *data = dev->data;
 
-	return data->read_only_table.asic_location;
+	/* Load tables on first access */
+	if (!data->initialized) {
+		if (tt_bh_fwtable_load_tables(dev) != 0) {
+			return 0;
+		}
+	}
```
cfriedt, line 162:
```suggestion
	if(tt_bh_fwtable_load_tables(dev) != 0) {
		return 0;
	}
```

dts/bindings/misc/tenstorrent,bh-chip.yaml
```diff
@@ -8,9 +8,9 @@ compatible: "tenstorrent,bh-chip"
 properties:
   flash:
     type: phandle
-  bh_smbus:
+  bh-smbus:
     type: phandle
   arc:
     type: phandle
-  gpio_straps:
+  gpio-straps:
```
cfriedt, line 15: These changes should be made in a separate commit.

dts/bindings/misc/tenstorrent,bh-fwtable.yaml
```diff
@@ -9,8 +9,9 @@ include: base.yaml
 
 properties:
 
-  flash-dev:
+  flash_dev:
     type: phandle
+    required: true
```
cfriedt, line 14: This should probably be reverted (dt properties should use dashes, and I think we want to allow not specifying the flash device (as a special means of indicating that we should skip load firmware tables).

Issue #292, opened 2025-06-03
ci: add nightly action to remove feature branches in the main repo


cfriedt:
**Is your enhancement proposal related to a problem? Please describe.**
In order to keep a cleaner git history and to discourage developers with write access from pushing directly to the repository, create a nightly CI rule that deletes branches other than `main` and `v*-branch`. 

This is consistent with the open-source GH module of forking a repository and making PRs from forks.

This is similar to what is done upstream with the exception that upstream allows "collaboration" branches.

**Describe the solution you'd like**
Create a nightly CI job to delete feature branches.

**Describe alternatives you've considered**

Keeping feature branches indefinitely

**Additional context**

PR #307, opened 2025-06-09
.github: workflows: Add nightly action to remove feature branches

alexapostolu:
Add nightly action to remove feature branches. Keep main and version branches.

Just outputting the branch names for now, and once confirmed it works, then I'll add the actual deletion

Fixes: #292

### DIFF
--- a/.github/workflows/cleanup-branches.yml
+++ b/.github/workflows/cleanup-branches.yml
@@ -1,0 +1,44 @@
+name: 'Nightly Branch Cleanup'
+
+on:
+  schedule:
+    - cron: '0 5 * * *'
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        required: true
+        type: boolean
+        default: true
+
+jobs:
+  cleanup-branches:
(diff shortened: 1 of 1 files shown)


cfriedt:
I think we should be able to merge this after the RC. I'll give it a manual test as well later.

cfriedt:
This looks good to me. I gave people a heads up on Thursday or Friday last week that this would be going in, so as long as @danieldegrasse and @afongTT don't have any concerns, we should be able to merge it.

alexapostolu:
Dry run is enabled by default so it will only print the branches to be deleted without actually deleting them. I tested it on a separate repository and it worked as expected, but just to be extra safe I could make another PR to disable dry run once this is merged and we know it works?

Screenshot 2025-06-17 at 5 02 58 PM

danieldegrasse (approved):
Looks fine to me

cfriedt:
@alexapostolu - sounds good - make another PR to flip the switch

PR #308, opened 2025-06-09
scripts:verify-blob: add blob sha256sum checks

lzhengTT:
Merging outdated blob sha256sum in module.yml can break CI.

Implemented blob sha256sum checks.

sample output when mismatched blob is detected
```
Blob tt_blackhole_gddr_init.bin has changed. Expected SHA256: df1ceefa477e5d8fbb02f37b1e5e0a0c7aabb875669167e6db2f7ce7cb9e6325, but got 97372fed110dd3fe8d3feba4260e6f48db8e30b0bf08ced31a0628b0ba4ed7e8.
Error: Process completed with exit code 1.
```

### DIFF
--- a/.github/workflows/verify-blob.yml
+++ b/.github/workflows/verify-blob.yml
@@ -1,0 +1,46 @@
+name: Verify Blob
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - edited
+    branches:
+      - main
+      - v*-branch
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  verify_blob:
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


danieldegrasse:
I *think* the specific issue right now is that we fetch the blobs from the main branch, so the SHA256 in a PR would always need to match the SHA on the main branch or other checks would fail. As far as I know, the only way we can fix this is to stop fetching blobs using `west blobs fetch` in CI. The blob fetch command is intended to download from another url, not the same repo (which is why we run into this ordering issue)

lzhengTT:
> I don't really understand what is up with the compliance failures here. Editing the files, I see no errors.

I went back into my first commit and fixed the formatting there and now it works! I think it is because the CI test runs through each commit of the PR?

cfriedt:
> Sure- will this PR fix it though? I'm not clear what the solution is there- a PR changing blobs can't also change the location the blobs are fetched from, or we'd need two PRs (one to change the SHA and fetch the blobs from the PR, and another to fetch the blobs from the mainline again)

West blob fetch does not re-fetch blobs if they are already present. It will check sha sums though.

PR #277, opened 2025-05-29
drivers:pinctrl: Reimplementing I2C pin initialization

lzhengTT:
Implementing I2C pins configuration in pinctrl driver using upstream Zephyr library functions.

Added tests for pinctrl and I2C drivers.

PR #320, opened 2025-06-16
app: smc: fall-back to null if wdt0 is disabled

cfriedt:
Previouosly, `app/smc` would unconditionally try to get a DT handle on `wdt0`. This would fail, however, when the watchdog timer was not enabled in DT.

Use `DEVICE_DT_GET_OR_NULL()` so that we are not left with an undefined symbol. i.e. get rid of the error that says

```shell
error: '__device_dts_ord_9' undeclared here
```

https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/15690400288/job/44203933940?pr=319

### DIFF
--- a/app/smc/src/main.c
+++ b/app/smc/src/main.c
@@ -27,7 +27,7 @@
 
 LOG_MODULE_REGISTER(main, CONFIG_TT_APP_LOG_LEVEL);
 
-static const struct device *const wdt0 = DEVICE_DT_GET(DT_NODELABEL(wdt0));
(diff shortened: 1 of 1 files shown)


danieldegrasse:
IMO the solution to this isn't to allow the watchdog to be NULL- it is to make sure the watchdog is enabled for any build of the SMC app. If you add an overlay for the SMC app that doesn't enable the watchdog there should be a build error- we need the watchdog enabled for the auto reset feature

cfriedt:
I incorporated the suggested changes into #319 but that's only for testing. @lzhengTT should pull them into #277

PR #321, opened 2025-06-16
spi: enable quad ddr mode

nxuTT:
enable quad ddr spi mode for SPI by setting bootstraps


cfriedt (commented):
LGTM, but can you also please add release notes? I think this should be categorized under performance improvements.

cfriedt (approved):

PR #324, opened 2025-06-17
soc + board: tt_blackhole: add skeleton dt bindings and nodes for missing devices

cfriedt:
Add skeleton DT bindings and nodes for a number of missing devices in the SMC SoC and board-level devicetrees.

The bindings and node properties should be elaborated when the drivers are implemented.

### DIFF
--- a/boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc.dts
+++ b/boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc.dts
@@ -43,6 +43,48 @@
 
 &i2c1 {
 	status = "disabled";
+
+	regulator0: max20816@64 {
+		compatible = "maxim,max20816";
+		reg = <0x64>;
+		status = "disabled";
+	};
+
+	regulator1: max20816@65 {
+		compatible = "maxim,max20816";
+		reg = <0x65>;
+		status = "disabled";
+	};
+
+	regulator2: mpm3695@33 {
+		compatible = "mps,mpm3695";
+		reg = <0x33>;
+		status = "disabled";
+	};
+
+	regulator3: max20730@54 {
+		compatible = "maxim,max20730";
+		reg = <0x54>;
+		status = "disabled";
+	};
+
+	regulator4: max20730@55 {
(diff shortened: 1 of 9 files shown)


boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc.dts
```diff
@@ -39,6 +39,48 @@
...
+
+	regulator2: mpm3695@33 {
+		compatible = "mps,mpm3695";
+		reg = <0x33>;
+		status = "disabled";
+	};
+
+	regulator3: max20730@54 {
+		compatible = "maxim,max20730";
+		reg = <0x54>;
```
afongTT, line 63: All the regulators are connected to i2c1

dts/arc/tenstorrent/tt_blackhole_smc.dtsi
```diff
@@ -177,6 +177,66 @@
...
 
+	eth0: bh_eth0 {
+		compatible = "tenstorrent,bh-eth";
+		status = "disabled";
+	};
+
+	eth1: bh_eth1 {
+		compatible = "tenstorrent,bh-eth";
+		status = "disabled";
+	};
```
afongTT, line 188: There are 14 of these, but not sure if that matters right now.

danieldegrasse (approved):
Fine to merge this, but we need to figure out what is causing the tensix reset test to fail. It is already failing on `yyz-zephyr-02`, which I took offline as that runner is redundant with `syseng-gh-01`. The P100 runner has no backup though, and the fact this is occurring across multiple BH ASICs makes me slightly concerned

boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc.dts
cfriedt: Fixed!

dts/arc/tenstorrent/tt_blackhole_smc.dtsi
cfriedt: Fixed!

PR #336, opened 2025-06-19
ci: concurrency: ensure key is unique per job

cfriedt:
For the build-fw reusable workflow, we observed the second job canceling the first job. This was because the `concurrency` key was not unique for each job in the workflow.

Add `${{ github.job }}` as another item to key from, so that two jobs within the same workflow do not cancel each other.


.github/workflows/hardware-long.yml
```diff
@@ -7,7 +7,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}${{ github.job }}
```
danieldegrasse, line 10: I don't think we want concurrency groups with the job ID for these jobs- if we add this, than a force push over a pull request won't cancel the previous job, will it? If you tested that and the job cancellation works, than I'm fine with this change. We do need it for the build-fw job though

cfriedt:
Closing, since this isn't a priority atm.

PR #338, opened 2025-06-20
app: smc: test_arc_watchdog: increase delay before polling DMC

danieldegrasse:
Increase delay before checking if SMC has communication with DMC, to improve reliability of the arc watchdog test


danieldegrasse:
I *really* dislike that this fixes the problem, but it seems to solve it

cfriedt (approved):

PR #340, opened 2025-06-20
Updated tt-console to use the kmd tlb allocation API

DRvader:
As of kmd 1.34+ tt-kmd provides an API for allocating TLBs instead of manually mapping the PCIe bar space. This gives us the ability to run alongside other programs that wish to also interact with the hardware.

This patch adds support for this capability and removes support for the older manually mapped TLBs.


DRvader:
Yes! kmd-2.0 is what I tested with

PR #343, opened 2025-06-23
scripts: tt_runner: add new west runner for flashing via stm32

cfriedt:
Add a new west runner for flashing via stm32.


cfriedt:
I think this is maybe obsolete now by #479

PR #345, opened 2025-06-23
ci: assigner: use set_assignee.py scripts from main zephyr repo

cfriedt:
Use the set_assignees.py script from the main Zephyr repo. This also avoids the import error observed in CI about 'get_maintainer'.


.github/workflows/assigner.yml_ignore
danieldegrasse, line 1: On this could we just edit the "Run Assignment Script" step to return 0 instead of calling the script here: https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/.github/workflows/assigner.yml#L31? I assume github will ignore files without a `.yml` extension, but I feel like `return 0` is more explicit. We also can just disable the workflow until it is fixed: https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/disabling-and-enabling-a-workflow#disabling-a-workflow

PR #353, opened 2025-06-24
[WIP] lib: tenstorrent: fan_ctrl: refactor fan_ctrl library to use driver

alayander:
Refactor fan_ctrl to use Zephyrized maxim,max6639 fan controller driver


cfriedt:
Yay passing tests 🎉 Nice work!

Is your plan to then replace all calls to `set_fan_speed()`, `get_fan_duty_cycle()`, and `get_fan_rpm()` in the DMC FW with PWM API calls?

cfriedt:
> @cfriedt Initially I was planning to leave the functions in `lib/tenstorrent/fan_ctrl/fan_ctrl.c` (`set_fan_speed()`, `get_fan_duty_cycle()`, and `get_fan_rpm()`) the way they currently are in the PR to act as wrappers to the PWM/Sensor API calls. Would it be more preferable to completely remove the library and directly use the APIs within the DMC FW?

I think in the long run, we should get rid of both `fan_ctrl.c` and `fan_ctrl.h` and just use PWM APIs (and e.g. fixed-point math) directly, but I think this is OK for now.

cfriedt (approved):

PR #358, opened 2025-06-25
drivers: i2c: add tenstorrent blackhole i2c controller

cfriedt:
Add a driver for Tenstorrent Blackhole I2C controllers.

This driver is somewhat unique in that it's mainly only ever used for recoverying the i2c bus and it acts as a bit of a "shell" for the Designware i2c controller.


cfriedt:
obsolete

PR #364, opened 2025-06-26
app: smc: e2e_test: arc_watchdog: vastly extend the time before DMC ping

danieldegrasse:
Extend the time before DMC ping to an unreasonably long value. If CI fails after this point, we can safely assume that there is an intermittent failure with the DMC pinging the SMC after watchdog reset


cfriedt:
Super weird how the one test is failing even with this. Man, it would be great to get vuart working everywhere by default.

PR #371, opened 2025-07-02
tt_boot_fs: move to devicetree + python script to generate yaml file

alexapostolu:
Move tt_boot_fs definitions to devicetree and add a Python script to generate `tt_boot_fs.yaml` at build time

Creates a file `tt_boot_fs.yaml` and also outputs,

Screenshot 2025-07-02 at 1 03 33 PM


dts/arc/tenstorrent/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -0,0 +1,75 @@
...
+ * Copyright (c) 2025 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/dts-v1/;
+
+/ {
+	partitions {
+		compatible = "tenstorrent,fixed-partitions";
```
cfriedt, line 11: Are there any gains by using `tenstorrent` instead of `zephyr` for the vendor in this compatible? I think we should be able to use the upstream one.

dts/bindings/mtd/tenstorrent,fixed-partitions.yaml
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2025 Tenstorrent AI ULC
```
cfriedt, line 1: I think we can likely drop this binding, since "padto" (i.e. partition size) is already captured by the upstream bindings.

It might be possible to use a partition-label-based convention for the executable bit as well, which we should be able to handle in python.

dts/arc/tenstorrent/tt_blackhole_fixed_partitions.dtsi
alexapostolu: Daniel said we want custom properties like `executable` and `padto` which aren't defined upstream

scripts/generate_yaml.py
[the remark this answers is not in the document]
alexapostolu: Would this script be generating YAML files for other devicetrees as well, or just the bootfs?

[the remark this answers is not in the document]
cfriedt: We should point it at the finalized devicetree (`zephyr.dts`) when it gets integrated into the build process.

dts/arc/tenstorrent/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) 2025 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/dts-v1/;
+
+/ {
```
cfriedt, line 9: I would just remove these bits

dts/arc/tenstorrent/tt_blackhole_fixed_partitions.dtsi
cfriedt, line 1: probably ok to merge now, and then perform some integration testing later. We want to make these partition sizes as large as they can ever be expected to grow so that the fixed partition mapping never needs to change. Any flashing algorithm should also be able to accommodate writing a binary that is smaller than a given partition etc.

So likely, the bmfw partition should be changed later to 440 k to accommodate #372

cfriedt (approved):
LGTM

PR #378, opened 2025-07-03
app: smc: pytest: add a e2e stress test

danieldegrasse:
Add an e2e stress test, which runs many of the same tests used in the
smoke test but over many more iterations. This should help catch
stability issues with the firmware that only occur sporadically.


danieldegrasse:
Run where hardware long tests pass on this PR (achieved by modifying `hardware-long.yaml` to run per-PR: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/16349591709/job/46193205492?pr=378

app/smc/pytest/e2e_stress.py
```diff
@@ -0,0 +1,275 @@
...
+    total_tries = min(MAX_TEST_ITERATIONS, 100)
+    for i in range(total_tries):
+        logger.info(f"Starting ARC watchdog test iteration {i}/{total_tries}")
+        # Setup ARC watchdog for a 1000ms timeout
+        arc_chip.arc_msg(ARC_MSG_TYPE_SET_WDT, True, False, wdt_timeout, 0, 1000)
+        # Sleep 1500, make sure we can still ping arc
+        time.sleep(1.5)
+        response = arc_chip.arc_msg(ARC_MSG_TYPE_TEST, True, False, 1, 0, 1000)
+        assert response[0] == 2, "SMC did not respond to test message"
+        assert response[1] == 0, "SMC response invalid"
```
cfriedt, line 118: Mentioned a long time ago, but for stress tests, we don't want to fail the test if a single iteration doesn't work.

We want to record the failure and move on to the next iteration of the test. That way, if a test passes 999/1000 tries, or is above a specific threshold, then the test will still pass.

What should be asserted is whether the success rate is above a certain threshold, not that a single iteration failed.

Also, it would be great to record results for this so we can see it over time on a graph.

Twister has a feature for recording results, in tests/benchmarks

danieldegrasse: I'll update this PR to record statistics, I agree that will be useful to have. However I think we should initially aim for 100% pass rate. This is how our current long tests are designed, and they pass nightly. If we are routinely failing a test than we can evaluate the impact of allowing a 99% or 99.9% pass rate

PR #379, opened 2025-07-03
lib: tenstorrent: bh_arc: extend time waiting for DMC ping response

danieldegrasse:
Extend the time waiting for a DMC ping response. It seems that the DMC sometimes does not respond within 50 ms, and this may be the root cause of the periodic failure in the ARC watchdog test.


danieldegrasse:
Closing in favor of the fix in #378

PR #389, opened 2025-07-04
lib: tenstorrent: remove fanctrl library

alayander:
Replace internal fan_ctrl library with upstream PWM & sensor APIs.

Follow up to: https://github.com/tenstorrent/tt-zephyr-platforms/pull/353


cfriedt:
@alayander - can you also update `tests/lib/Tenstorrent/bh_arc`?

That testsuite does not appear to be running in CI at the moment, otherwise I guess it would be caught. Can you double check and prepend a commit to `.github/workflows` to activate them if necessary? I thought it would have coverage under `unit-tests`.

alayander:
> @alayander - can you also update `tests/lib/tenstorrent/bh_arc`?
> 
> That testsuite does not appear to be running in CI at the moment, otherwise I guess it would be caught. Can you double check and prepend a commit to `.github/workflows` to activate it? I thought it would have coverage under the `run-unit-tests` workflow.

I added the verbose flag to the twister run command, which can be seen in this workflow: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/16124869407/job/45499286012?pr=389.

It reports that the tests under `tests/lib/tenstorrent/bh_arc` are running and passing properly:

> INFO    - 2/8 native_sim/native         lib/tenstorrent/bh_arc/lib.tenstorrent.bh_arc      PASSED (native 0.015s <host>)

alayander:
> @alayander - is `tests/lib/tenstorrent/bh_arc/src/fan_ctrl.c` still there? Maybe I missed it.
> 
> I think it should be possible to remove it if it's still there though.
> 
> It's weird that the test would still compile if the fan ctrl header is removed

Oh, that test exists to verify the functionality of the `fan_ctrl` library within `lib/tenstorrent/bh_arc`, which handles the messaging interface to update fan speed and the fan curve that updates speed based on temperatures. Specifically that test suite verifies the fan curve, so perhaps I should rename it to `.../fan_curve.c`?

Initially there were two `fan_ctrl` libraries, one that handles the above from smc and one that handles controlling the actual fan controller device from the dmc. This PR replaces the latter with direct pwm/sensor API calls.

cfriedt:
Ah, that makes sense now.

cfriedt (approved):

PR #391, opened 2025-07-05
bh_arc: msgqueue: do not use the system workqueue

cfriedt:
It's a very slippery slope to use the system workqueue without explicitly synchronizing all other threads that use it, in the presence of blocking operations. A quick survey of our code shows that no such synchronization exists.

Use a dedicated workqueue thread for dealing with interrupts / messages from the host.


cfriedt:
Someone at FB thought using the SWQ as a kind of synchronization primitive to serialize tasks was a good idea too. However, that ensured that the application could never properly scale.

I would strongly advise not to make the same mistake.

cfriedt:
@afongTT @danieldegrasse @alewycky-tenstorrent - is there a reason to not merge this?

cfriedt:
Not entirely sure what was wrong with this PR, but closing since it's really old now.

PR #393, opened 2025-07-07
ci: workflows: upload device log

danieldegrasse:
Upload device log from twister runs, so we can debug why flashing steps failed


cfriedt (approved):
LGTM

PR #398, opened 2025-07-09
lib: tenstorrent: bh_arc: implement aiclk_ppm sweep handler

alayander:
Adds a handler for aiclk_ppm sweep, which sets target aiclk to be randomized between set min and max freq.

Continuation of: https://github.com/tenstorrent/tt-zephyr-platforms/pull/98


lib/tenstorrent/bh_arc/aiclk_ppm.h
```diff
@@ -7,6 +7,7 @@
 #define AICLK_PPM_H
 
 #include <stdint.h>
+#include <stdlib.h>
```
afongTT, line 10: Is this needed here?

lib/tenstorrent/bh_arc/aiclk_ppm.c
```diff
@@ -162,6 +190,11 @@ uint8_t ForceAiclk(uint32_t freq)
 	return 0;
 }
 
+uint32_t GetAiclkPPM(void)
+{
+	return aiclk_ppm.targ_freq;
+}
+
```
afongTT, line 197: Can you make this `GetAiclkTarg()`?

afongTT (changes requested):

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -180,7 +180,7 @@ static void UpdateThrottlerArb(ThrottlerId id)
 {
 	Throttler *t = &throttler[id];
 
-	float arb_val = aiclk_ppm.arbiter_max[t->arb_max];
+	float arb_val = GetThrottlerArbMax(t->arb_max);
```
cfriedt, line 183: one day, I hope we can get rid of floats here. It would be nice to use integer arithmetic only.

lib/tenstorrent/bh_arc/aiclk_ppm.c
alayander: Fixed in latest push

lib/tenstorrent/bh_arc/throttler.c
alayander: I can try looking into that in a follow up PR :)

cfriedt (approved):

PR #399, opened 2025-07-09
Refactoring tt-boot-fs

mshimozatoTT:
tt-boot-fs is never initialized at the beginning of the boot process which prevents tables (fwtable, flash_only table, read-only table) from being loaded at boot-up. 

solution: tt_boot_fs_ng

rewriting the tt-boot-fs API to be centered around a C struct (holding all necessary information to operate on the filesystem, including a pointer to the physical flash device). this eliminates the need for a separate initialization step at runtime by moving the filesystem setup from runtime to compile-time.


lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -171,44 +152,81 @@ static tt_checksum_res_t calculate_and_compare_checksum(uint8_t *data, size_t nu
...
+	return count;
+}
 
-		/* Found the right file descriptor */
-		*fd_data = boot_fs_cache[i];
-		return TT_BOOT_FS_OK;
+int tt_bootfs_ng_find_fd_by_tag(const tt_boot_fs_ng *fs, const uint8_t *tag, tt_boot_fs_fd *fd_data)
+{
+	/* Use the new cacheless approach to get the list of all fds */
+	tt_boot_fs_fd fds[32];
```
cfriedt, line 200: we should be able to get away with a cache of 16 fds

lib/tenstorrent/fwupdate/fwupdate.c
```diff
@@ -186,43 +167,32 @@ int tt_fwupdate(const char *tag, bool dry_run, bool reboot)
...
+		LOG_DBG("Device %s is not ready", tt_boot_fs.dev->name);
 		return -ENODEV;
 	}
 
-	LOG_DBG("Parsing SPI flash %s", flash1_dev->name);
+	LOG_DBG("Parsing SPI flash %s", tt_boot_fs.dev->name);
 
-	for (uint32_t addr = TT_BOOT_FS_OFFSET + TT_BOOT_FS_FD_HEAD_ADDR, i = 0;
-	     i < CONFIG_TT_BOOT_FS_IMAGE_COUNT_MAX; addr = tt_boot_fs_next(addr), ++i) {
+	tt_boot_fs_fd fds[32];
```
cfriedt, line 177: Would be best to use a common Kconfig value here.

E.g. `tt_boot_fs_fd fds[CONFIG_TT_BOOT_FS_MAX_FDS];`

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -171,44 +152,81 @@ static tt_checksum_res_t calculate_and_compare_checksum(uint8_t *data, size_t nu
...
+	int fd_count = tt_boot_fs_ls(fs, fds, ARRAY_SIZE(fds));
+
+	if (fd_count < 0) {
+		return TT_BOOT_FS_ERR;
+	}
+
+	for (int i = 0; i < fd_count; i++) {
```
cfriedt, line 207:
```cpp
ARRAY_FOR_EACH(fds, i) {
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -171,44 +152,81 @@ static tt_checksum_res_t calculate_and_compare_checksum(uint8_t *data, size_t nu
...
+		if (memcmp(fds[i].image_tag, tag, TT_BOOT_FS_IMAGE_TAG_SIZE) == 0) {
```
cfriedt, line 208: to be honest, I think it's probably safe to use `strncmp()` here.

include/tenstorrent/tt_boot_fs.h
```diff
@@ -57,21 +57,9 @@ typedef struct {
...
-typedef int (*tt_boot_fs_erase)(uint32_t addr, uint32_t size);
-
-typedef struct {
-	tt_boot_fs_read hal_spi_read_f;
-	tt_boot_fs_write hal_spi_write_f;
-	tt_boot_fs_erase hal_spi_erase_f;
-} tt_boot_fs;
-
 typedef struct {
 	uintptr_t magic; /* TT_BOOT_FS_NG_MAGIC */
```
cfriedt, line 61: As long as the boot fs is initialized earlier than its callers, we can probably just pass around the `const struct device *dev` that represents the flash device without needing to pass around the `tt_boot_fs_ng` pointer or checking for magic.

So likely, we can get rid of all references to `magic` or `TT_BOOT_FS_NG_MAGIC`.

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -14,34 +14,31 @@
...
-tt_boot_fs boot_fs_data;
-static tt_boot_fs_fd boot_fs_cache[16];
-
 uint32_t tt_boot_fs_next(uint32_t last_fd_addr)
 {
 	return (last_fd_addr + sizeof(tt_boot_fs_fd));
 }
 
-int tt_boot_fs_ng_read(const tt_boot_fs_ng *fs, uint32_t addr, uint8_t *buffer, size_t size)
+static int tt_bootfs_ng_initial_check(const tt_boot_fs_ng *fs)
```
cfriedt, line 22: When is this called that it isn't ready? Just wondering, because maybe we can get rid of the `tt_boot_fs_ng` struct entirely and just use the flash device pointer.

mshimozatoTT: yes sorry I think that's a better idea, I will do that

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -72,11 +65,9 @@ const FlashInfoTable *tt_bh_fwtable_get_flash_info_table(const struct device *de
...
 	struct bh_fwtable_data *data = dev->data;
 
-	/* Load tables on first access */
-	if (!data->initialized) {
-		if (tt_bh_fwtable_load_tables(dev) != 0) {
-			return NULL;
-		}
+	if (!device_is_ready(dev)) {
+		LOG_ERR("%s device is not ready", "bh_fwtable");
+		return NULL;
```
nxuTT, line 70: I'm not sure if it works, if tt_bh_fwtable_init fails, device_is_ready(dev) will return false. Is that correct?
It is possible for you to get an image without fw table during assembly, then your get_fwtable function is going to return NULL. 

For getting vendor_info and board_id, you don't check whether this function returns a NULL pointer, it could result in a seg fault.

mshimozatoTT: oh yes, thank you so much for catching this! I think your solution makes sense.

PR #402, opened 2025-07-10
lib: bh_arc: reduce scratchpad size down from 64 kB to 36kB

cfriedt:
We had some difficulty building our application with debug information before because the linked binary would not fit into memory reserved for cmfw.

Reduce the large scratchpad size down from 64 kB to 36kB, which is just larger than the size of ERISC firmware. The size can be further reduced by adjusting the firmware loading algorithms to be flash-aware. If using malloc is a possibility then the buffers can be freed when they are no longer required, which would allow for additional runtime memory usage.

Technically, I think we should also be able to extend the boundary within CSM (SRAM) where CMFW may be loaded, but if we can get away with using a smaller static buffer and still meet the same performance and functional requirements, we might as well do that too.

### DIFF
--- a/lib/tenstorrent/bh_arc/Kconfig
+++ b/lib/tenstorrent/bh_arc/Kconfig
@@ -77,9 +77,13 @@ config TT_BH_ARC_WDT_FEED_INTERVAL
 
 config TT_BH_ARC_SCRATCHPAD_SIZE
 	hex "Size of scratchpad memory in bytes"
-	default 0x10000
+	default 0x9000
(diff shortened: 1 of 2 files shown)


cfriedt:
At some point, when we Zephyrize ERISC and MRISC drivers (which should be soon), it might be preferable to have a common / generic "load data from SPI flash via TLB" algorithm somewhere, that we could use to load FW one chunk at a time.

danieldegrasse (approved):

PR #403, opened 2025-07-10
drivers: serial: uart_tt_virt: updates to support tracing on blackhole

danieldegrasse:
Add support for tracing on blackhole. Tracing uses the VUART, and thus adds the following to the tt-console program:
- support for "binary pipe" mode. When tt-console detects output is being redirected to a file (or anything that is not a tty) it will read data directly from the VUART, and not swap carriage returns for newlines. It also uses a more efficient copy routine, to improve UART read performance. This enables the console program to keep pace with tracing data.
- support for multiple channels. This is needed so that we can have console output on one channel while piping tracing data across another channel

> [!NOTE]
Cards do not write to SPI reliably when tracing is active. Avoid flashing a build with tracing enabled on a card you cannot recover (IE one without an STM32 debug connection)

Tracing is enabled via tt-console- the console output can be written directly to a file, and input piped in can be sent to the ARC over vuart to enable or disable tracing.

Example of tracing usage:

```
$ west build -p -b tt_blackhole@p100a/tt_blackhole/smc --sysbuild app/smc/ -DEXTRA_CONF_FILE=tracing.conf -DEXTRA_DTC_OVERLAY_FILE=tracing.overlay
$ west flash -r tt_flash --force
$ mkdir data
$ scripts/tooling/tt-tracing data/trace
$ cp ../zephyr/subsys/tracing/ctf/tsdl/metadata data/
$ babeltrace2 data # or run python3 ./scripts/ctf_to_chrome.py to convert to chrome format
```

~This PR depends on #407, and will be in draft until that merges~


scripts/tt-console/console.c
```diff
@@ -758,22 +779,25 @@ static void usage(const char *progname)
...
-	  __func__, progname, UART_TT_VIRT_DISCOVERY_ADDR, TT_DEVICE, BH_SCRAPPY_PCI_DEVICE_ID,
+	  __func__, progname, UART_TT_VIRT_DISCOVERY_ADDR, UART_CHANNEL,
+	  TT_DEVICE, BH_SCRAPPY_PCI_DEVICE_ID,
 	  UART_TT_VIRT_MAGIC);
 }
 
 static int parse_args(struct console *cons, int argc, char **argv)
 {
 	int c;
+	unsigned long channel = UART_CHANNEL;
```
cfriedt, line 798: This should probably go into the `cons` structure.

scripts/tt-console/console.c
```diff
@@ -787,6 +811,20 @@ static int parse_args(struct console *cons, int argc, char **argv)
 			}
 			cons->addr = addr;
 		} break;
+		case 'c': {
+			errno = 0;
+			channel = strtol(optarg, NULL, 0);
```
cfriedt, line 816:
```suggestion
			cons->channel = strtol(optarg, NULL, 0);
```

scripts/tt-console/console.c
```diff
@@ -860,14 +898,17 @@ static int parse_args(struct console *cons, int argc, char **argv)
 		}
 	}
 
+	/* Offset address based on channel selection */
+	cons->addr += channel * sizeof(uint32_t);
```
cfriedt, line 902:
```suggestion
	cons->addr += cons->channel * sizeof(uint32_t);
```

app/smc/vuart.overlay
```diff
@@ -10,10 +10,15 @@
 		zephyr,shell-uart = &vuart0;
 	};
 
-	vuart0: uart_tt_virt {
-		compatible = "tenstorrent,vuart";
-		version = <0x00000000>;
-		status = "disabled";
+	vuarts {
```
cfriedt, line 13: I guess this is necessary for non-contiguous / starting from 0 instances, right?

scripts/tt-console/console.c
```diff
@@ -664,7 +671,17 @@ static int loop(struct console *const cons)
...
+	 * Check if STDOUT is a tty. If not, we should simply stream binary
+	 * data the file without any processing.
+	 */
+	if (isatty(STDOUT_FILENO)) {
+		using_tty = true;
+		I("Press Ctrl-a,x to quit");
+	} else {
+		fprintf(stderr, "Not a tty, streaming binary data to STDOUT. "
+				"Press Ctrl-C to quit.\n");
+	}
```
cfriedt, line 684: I would undo this bit - you can definitely stream binary data even if it is an interactive console.

Additionally, console interactivity is usually determined by whether or not the input is a tty.

cfriedt:
@danieldegrasse - are there existing zephyr shell commands to turn tracing on and off?

danieldegrasse:
> @danieldegrasse - are there existing zephyr shell commands to turn tracing on and off?

No- the expectation is that the backend (UART, usb, custom) will call `tracing_cmd_handle` with the text sent from the host side.

I think this architecture makes sense, because in order to get a valid trace in CTF you need to:
- enable tracing
- read all tracing data the system outputs (without dropping any)
- disable tracing
- read any remaining tracing data

So the enable/disable functionality is essentially synced with the tracing collection

drivers/serial/uart_tt_virt.c
```diff
@@ -165,33 +163,9 @@ static void uart_tt_virt_irq_callback_set(const struct device *dev,
...
-
-	K_SPINLOCK(&data->err_lock) {
-		data->err_irq_en = true;
-		submit = !!data->err_flags;
-	}
-
-	if (submit) {
-		(void)k_work_submit(&data->irq_work);
-	}
-}
```
cfriedt, line 190: What's the reason for removing `err_disable()` and `err_enable()`?

I think they should probably be reworked. The idea was that this would be used against upstream test suites, and I think one of them required this functionality.

scripts/tooling/logging.h
```diff
@@ -0,0 +1,24 @@
+/**
+ * Copyright (c) 2025 Tenstorrent AI ULC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef __LOGGING_H__
+#define __LOGGING_H__
```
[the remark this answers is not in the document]
danieldegrasse: Good catch, fixed

drivers/serial/uart_tt_virt.c
danieldegrasse: I didn't love the requirement to nest two spinlocks to check if the interrupt could be disabled, and I didn't see these APIs being used. I've added them back though and reworked them

scripts/tooling/logging.h
```diff
@@ -0,0 +1,24 @@
+/**
+ * Copyright (c) 2025 Tenstorrent AI ULC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef LOGGING_H_
+#define LOGGING_H_
```
cfriedt, line 7: Is it possible to use more unique header guards such as `TENSTORRENT_TOOLING_LOGGING_H_`? There is no namespacing at all here.

Same with `vuart.h`

drivers/serial/uart_tt_virt.c
```diff
@@ -214,27 +219,30 @@ static int uart_tt_virt_irq_is_pending(const struct device *dev)
...
 		data->rx_irq_en = false;
 	}
+	tx_key = k_spin_lock(&data->tx_lock);
+	err_key = k_spin_lock(&data->err_lock);
+	if (!(data->tx_irq_en || data->err_irq_en)) {
+		/* If other interrupts are disabled, stop timer */
+		k_timer_stop(&data->irq_timer);
+	}
+	k_spin_unlock(&data->err_lock, err_key);
+	k_spin_unlock(&data->tx_lock, tx_key);
```
cfriedt, line 234:
```suggestion
```

drivers/serial/uart_tt_virt.c
```diff
@@ -272,27 +280,30 @@ static int uart_tt_virt_irq_tx_complete(const struct device *dev)
...
 		data->tx_irq_en = false;
 	}
+	rx_key = k_spin_lock(&data->rx_lock);
+	err_key = k_spin_lock(&data->err_lock);
+	if (!(data->rx_irq_en || data->err_irq_en)) {
+		/* If other interrupts are disabled, stop timer */
+		k_timer_stop(&data->irq_timer);
+	}
+	k_spin_unlock(&data->err_lock, err_key);
+	k_spin_unlock(&data->rx_lock, rx_key);
```
cfriedt, line 295:
```suggestion
```

cfriedt (changes requested):
As soon as we're acquiring multiple locks, it becomes a bit of a risk for deadlock, but it's also much more expensive than necessary.

Can you switch to a single `lock` that covers all internal state of `data` and change the locking strategies in `uart_tt_virt_irq_{err,rx,tx}_{en,dis}able()`?

Also, please include path components in headers for header guards.

danieldegrasse:
> As soon as we're acquiring multiple locks, it becomes a bit of a risk for deadlock, but it's also much more expensive than necessary.
> 
> Can you switch to a single `lock` that covers all internal state of `data` and change the locking strategies in `uart_tt_virt_irq_{err,rx,tx}_{en,dis}able()`?
> 
> Also, please include path components in headers for header guards.

Reworked to use one lock, and header guards have been namespaced

cfriedt:
@danieldegrasse - can you make a follow-up PR to update the README?

PR #405, opened 2025-07-11
scripts: tt_boot_fs.py

alexapostolu:
Generate a boot fs binary from the yaml file, perform fsck on the binary, and use that in the build

Follow up to #371


alexapostolu:
I still have to make the DTS compatible with all board (only works for p150a), and I have to remove the old bootfs yaml files.

cfriedt (approved):

PR #407, opened 2025-07-12
scripts: tt-console: properly support vuart input

danieldegrasse:
Fix a few issues prevent vuart input from functioning correctly. With these changes the shell works on the SMC:

image

Note that the SMC ram size (in the LD file) needs to be increased to get the shell to fit without overflowing the SRAM region, but it does work

### DIFF
--- a/scripts/tt-console/console.c
+++ b/scripts/tt-console/console.c
@@ -604,7 +604,7 @@ static inline size_t vuart_space(struct console *cons)
 		return 0;
 	}
 
-	return tt_vuart_buf_size(vuart->rx_head, vuart->rx_tail);
+	return tt_vuart_buf_space(vuart->rx_head, vuart->rx_tail, vuart->rx_cap);
 }
 
 static inline void vuart_putc(struct console *cons, int ch)
@@ -615,13 +615,13 @@ static inline void vuart_putc(struct console *cons, int ch)
 		return;
 	}
 
-	if (tt_vuart_buf_space(vuart->rx_head, vuart->rx_tail, vuart->rx_cap) > 0) {
-		++vuart->rx_tail;
-	}
-
 	volatile char *const rx_buf = (volatile char *)&vuart->buf[vuart->tx_cap];
 
 	rx_buf[vuart->rx_tail % vuart->rx_cap] = ch;
+
+	if (tt_vuart_buf_space(vuart->rx_head, vuart->rx_tail, vuart->rx_cap) > 0) {
+		++vuart->rx_tail;
+	}
 }
 
 static inline int vuart_getc(struct console *cons)
@@ -697,6 +697,8 @@ static int loop(struct console *const cons)
 				putchar('\r');
 			}
 			(void)putchar(ch);
+			/* Flush to STDOUT */
+			(void)fflush(stdout);
 		}
 
 		fd_set fds;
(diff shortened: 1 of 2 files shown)


scripts/tt-console/console.c
```diff
@@ -707,7 +709,7 @@ static int loop(struct console *const cons)
 		FD_ZERO(&fds);
 		FD_SET(STDIN_FILENO, &fds);
 
-		ret = select(STDIN_FILENO + 1, &fds, NULL, NULL, &tv);
```
cfriedt, line 710: This should be the highest-numbered file descriptor in the set + 1. 

Of course, we all know that `STDIN_FILENO` is 0, but it's good to illustrate the relationship. I don't think there is any damage keeping it in.

include/tenstorrent/uart_tt_virt.h
```diff
@@ -185,7 +185,7 @@ static inline int tt_vuart_poll_in(volatile struct tt_vuart *vuart, unsigned cha
 
 		if (atomic_compare_exchange_strong(headp, &head, head + 1)) {
 			*p_char = vuart->buf[offs + (head % cap)];
-			return *p_char;
```
cfriedt, line 188: this change seems to break the api

> @return The incoming byte, or -1 (EOF) if the buffer is empty

scripts/tt-console/console.c
```diff
@@ -697,6 +697,8 @@ static int loop(struct console *const cons)
 				putchar('\r');
 			}
 			(void)putchar(ch);
+			/* Flush to STDOUT */
```
cfriedt, line 700: Is this needed? Typically stdout buffered file descriptors flush automatically when their internal buffers are full (or if they are configured for line-buffering, when a newline is encountered). That reduces the number of system calls needed.

I would guess the main fix needed here was replacing `tt_vuart_buf_size()` with `tt_vuart_buf_space()`.

scripts/tt-console/console.c
```diff
@@ -604,7 +604,7 @@ static inline size_t vuart_space(struct console *cons)
 		return 0;
 	}
 
-	return tt_vuart_buf_size(vuart->rx_head, vuart->rx_tail);
+	return tt_vuart_buf_space(vuart->rx_head, vuart->rx_tail, vuart->rx_cap);
```
cfriedt, line 607: good catch 👍

scripts/tt-console/console.c
```diff
@@ -615,13 +615,13 @@ static inline void vuart_putc(struct console *cons, int ch)
...
-		++vuart->rx_tail;
-	}
-
 	volatile char *const rx_buf = (volatile char *)&vuart->buf[vuart->tx_cap];
 
 	rx_buf[vuart->rx_tail % vuart->rx_cap] = ch;
+
+	if (tt_vuart_buf_space(vuart->rx_head, vuart->rx_tail, vuart->rx_cap) > 0) {
+		++vuart->rx_tail;
+	}
```
cfriedt, line 624: interesting - i can sort of see how this might have been like an off-by-one error.

what kind of error was happening before?

include/tenstorrent/uart_tt_virt.h
danieldegrasse: Fixed, I'll just check the return value in `uart_tt_virt_poll_in`

scripts/tt-console/console.c
danieldegrasse: Fixed

scripts/tt-console/console.c
cfriedt: OK - it's possibly a side-effect of termios mucking.

PR #409, opened 2025-07-15
boards: tt_blackhole: default sysclk to 800 MHz

cfriedt:
Set sysclk to 800 MHz instead of 500, or 540.

### DIFF
--- a/soc/tenstorrent/tt_blackhole/Kconfig.defconfig.tt_blackhole_smc
+++ b/soc/tenstorrent/tt_blackhole/Kconfig.defconfig.tt_blackhole_smc
@@ -25,7 +25,7 @@ config RGF_NUM_BANKS
 	default 1
 
 config SYS_CLOCK_HW_CYCLES_PER_SEC
-	default 540000000
+	default 800000000
 
 config CODE_DENSITY
 	default y
--- a/app/smc/boards/tt_blackhole_tt_blackhole_smc.overlay
+++ b/app/smc/boards/tt_blackhole_tt_blackhole_smc.overlay
@@ -11,14 +11,6 @@
 	status = "disabled";
 };
 
-/*
- * ARC clock will be raised to 800MHz in the application
- */
-&sysclk {
-	status = "okay";
-	clock-frequency = <800000000>;
-};
-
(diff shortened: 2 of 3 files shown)


soc/tenstorrent/tt_blackhole/Kconfig.defconfig.tt_blackhole_smc
```diff
@@ -25,7 +25,7 @@ config RGF_NUM_BANKS
 	default 1
 
 config SYS_CLOCK_HW_CYCLES_PER_SEC
```
danieldegrasse, line 27: We may want to set this to 540 if BH_ARC is disabled for now, because that is what indicates if the PLL will actually be reclocked

cfriedt: Hmm.... @alexapostolu - how close are we to getting the PLL driver working?

It would be great to calibrate the ARC clock against the host to make sure that it's not too far off.

afongTT: Not sure why calibration is needed? We set the PLL's ratio vs the refclk, which is 50 MHz.

We also do have clock counters on the chip which count PLL clocks against refclk cycles.

alexapostolu: @cfriedt @lzhengTT was working on integrating the PLL driver with the code

lzhengTT: @alexapostolu Yep, @afongTT if you could let me know the de-dupping and required order of the PLLInit()'s, that should be the last bit left to complete for the integration.

cfriedt: @lzhengTT - do you have a PR ready? It would be much easier to comment on that.

cfriedt:
I think this isn't necessary any longer, so it can probably be closed.

PR #411, opened 2025-07-16
Automate preflash image creation

nxuTT:
added script to create preflash image for Quanta assembly

### DIFF
--- a/scripts/build-preflash.sh
+++ b/scripts/build-preflash.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+# Copyright (c) 2025 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+# Helper script to build preflash bootfs image for assembly
+set -e
+
+# handle temp directory for the build
+if [ -z "$BUNDLE_TEMP_PREFIX" ]; then
+  TEMP_DIR="$(mktemp -d)/"
+
+  # Sanity check
+  if [ "$TEMP_DIR" = "/" ]; then
+    echo "ERROR: Failed to create temp dir or invalid TEMP_DIR ('$TEMP_DIR')"
+    exit 1
+  fi
+
+  echo "Created TEMP_DIR"
+else
+  echo "Using existing BUNDLE_TEMP_PREFIX: $BUNDLE_TEMP_PREFIX"
+  TEMP_DIR="$BUNDLE_TEMP_PREFIX"
+fi
+
+cleanup() {
+    if [ -z "$BUNDLE_TEMP_PREFIX" ] && [ -d "$TEMP_DIR" ]; then
+        rm -rf "$TEMP_DIR"
+        echo "Removed TEMP_DIR"
+    fi
+}
+trap cleanup EXIT
+
+# Get the script directory and base paths
+SCRIPT_DIR="$(dirname "$0")"
+TTZP_BASE="$(dirname "$SCRIPT_DIR")"
+
+# board revision is agnostic, we use p100 for preflash
+BOARD_REV="p100"
+BUILD_DIR="$TEMP_DIR"
+
+# firmware version info
+MAJOR="$(grep "^VERSION_MAJOR" "$TTZP_BASE/VERSION" | awk '{print $3}')"
+MINOR="$(grep "^VERSION_MINOR" "$TTZP_BASE/VERSION" | awk '{print $3}')"
+PATCH="$(grep "^PATCHLEVEL" "$TTZP_BASE/VERSION" | awk '{print $3}')"
+EXTRAVERSION="$(grep "^EXTRAVERSION" "$TTZP_BASE/VERSION" | awk '{print $3}')"
+
+if [ -z "$MAJOR" ] || [ -z "$MINOR" ] || [ -z "$PATCH" ]; then
+  echo "required version info missing: MAJOR=$MAJOR MINOR=$MINOR PATCH=$PATCH"
+  exit 1
+fi
+
+if [ -n "$EXTRAVERSION" ]; then
+  EXTRAVERSION="-$EXTRAVERSION"
+fi
+
+RELEASE="$MAJOR.$MINOR.$PATCH$EXTRAVERSION"
+
+echo "Running sysbuild for $BOARD_REV..."
+BOARD="$("$SCRIPT_DIR/rev2board.sh" "$BOARD_REV")"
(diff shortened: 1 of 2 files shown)


scripts/build-preflash.sh
```diff
@@ -0,0 +1,60 @@
...
+MAJOR="$(grep "^VERSION_MAJOR" "$TTZP_BASE/VERSION" | awk '{print $3}')"
+MINOR="$(grep "^VERSION_MINOR" "$TTZP_BASE/VERSION" | awk '{print $3}')"
+PATCH="$(grep "^PATCHLEVEL" "$TTZP_BASE/VERSION" | awk '{print $3}')"
+EXTRAVERSION="$(grep "^EXTRAVERSION" "$TTZP_BASE/VERSION" | awk '{print $3}')"
+
+if [ -z "$MAJOR" ] || [ -z "$MINOR" ] || [ -z "$PATCH" ]; then
+  echo "required version info missing: MAJOR=$MAJOR MINOR=$MINOR PATCH=$PATCH"
+  exit 1
+fi
+
```
cfriedt, line 45: Without this, you'll get e.g. `preflash-18.6.0rc3.ihex` instead of `preflash-18.6.0-rc3.ihex`

```suggestion

if [ -n "$EXTRAVERSION" ]; then
  EXTRAVERSION="-$EXTRAVERSION"
  EXTRAVERSION_NUMBER="$(echo "$EXTRAVERSION" | sed 's/[^0-9]*//g')"
fi
```

cfriedt (dismissed):
Let's keep unrelated changes in separate PRs.

nxuTT:
> Let's keep unrelated changes in separate PRs.

https://github.com/tenstorrent/tt-zephyr-platforms/pull/416

scripts/build-preflash.sh
```diff
@@ -0,0 +1,64 @@
...
+  TEMP_DIR="$(mktemp -d)/"
+  echo "Created TEMP_DIR"
+else
+  echo "Using existing BUNDLE_TEMP_PREFIX: $BUNDLE_TEMP_PREFIX"
+  TEMP_DIR="$BUNDLE_TEMP_PREFIX"
+fi
+
+cleanup() {
+    if [ -z "$BUNDLE_TEMP_PREFIX" ] && [ -d "$TEMP_DIR" ]; then
+        rm -Rf "$TEMP_DIR"
```
alewycky-tenstorrent, line 20: lower-case -r is canonical

scripts/build-preflash.sh
```diff
@@ -0,0 +1,64 @@
...
+trap cleanup EXIT
+
+# Get the script directory and base paths
+SCRIPT_DIR="$(dirname "$0")"
+TTZP_BASE="$(dirname "$SCRIPT_DIR")"
+PATH="$SCRIPT_DIR:$PATH"
+
+# board revision is agnostic, we use p100 for preflash
+BOARD_REV="p100"
+BUILD_DIR="${TEMP_DIR}"
```
alewycky-tenstorrent, line 33: Why `{}` here but not in any other variable expansion? They aren't necessary here.

scripts/build-preflash.sh
```diff
@@ -0,0 +1,64 @@
...
+  EXTRAVERSION="-$EXTRAVERSION"
+fi
+
+RELEASE="$MAJOR.$MINOR.$PATCH$EXTRAVERSION"
+
+echo "Running sysbuild for $BOARD_REV..."
+BOARD="$("$SCRIPT_DIR/rev2board.sh" "$BOARD_REV")"
+west build -d "$BUILD_DIR" --sysbuild -p -b "$BOARD" app/smc \
+    -DEXTRA_CONF_FILE=vuart.conf \
+    -DEXTRA_DTC_OVERLAY_FILE=vuart.overlay \
```
alewycky-tenstorrent, line 56: Probably leave vuart out of the recovery firmware. People are just going to replace it with the real firmware, not interact with it.

scripts/build-preflash.sh
```diff
@@ -0,0 +1,64 @@
...
+    >/dev/null 2>&1
+echo "Build completed successfully"
+
+# Create preflash.ihex using preflash-bootfs.yaml
+echo "Generating preflash.ihex..."
+YAML_PATH="./boards/tenstorrent/tt_blackhole/bootfs/preflash-bootfs.yaml"
```
alewycky-tenstorrent, line 62: Earlier the script works out its directory (and the tt-z-p directory), so I assume it can run from anywhere, but then there's this relative path. Does tt_boot_fs want a path relative to tt-z-p or does this break if the script is run from a place where this path isn't valid?

scripts/build-preflash.sh
alewycky-tenstorrent: This is kind of dangerous because if somehow mktemp -d outputs an empty string then TEMP_DIR=/ and this does `rm -rf /` which has well-known downsides. Maybe sanity-check TEMP_DIR when initializing it.

scripts/build-preflash.sh
```diff
@@ -0,0 +1,64 @@
...
+        rm -Rf "$TEMP_DIR"
+        echo "Removed TEMP_DIR"
+    fi
+}
+trap cleanup EXIT
+
+# Get the script directory and base paths
+SCRIPT_DIR="$(dirname "$0")"
+TTZP_BASE="$(dirname "$SCRIPT_DIR")"
+PATH="$SCRIPT_DIR:$PATH"
```
alewycky-tenstorrent, line 29: Why actually are we setting PATH here? The only thing you run is tt_boot_fs.py, you could just specify the path to it. Or does tt_boot_fs.py need SCRIPT_DIR on the PATH?

scripts/build-preflash.sh
```diff
@@ -0,0 +1,67 @@
...
+
+if [ -n "$EXTRAVERSION" ]; then
+  EXTRAVERSION="-$EXTRAVERSION"
+fi
+
+RELEASE="$MAJOR.$MINOR.$PATCH$EXTRAVERSION"
+
+echo "Running sysbuild for $BOARD_REV..."
+BOARD="$("$SCRIPT_DIR/rev2board.sh" "$BOARD_REV")"
+west build -d "$BUILD_DIR" --sysbuild -p -b "$BOARD" $TTZP_BASE/app/smc >/dev/null 2>&1
```
alewycky-tenstorrent, line 60: Should have quotes around $TTZP_BASE/app/smc

alewycky-tenstorrent (approved):

PR #413, opened 2025-07-16
Switch DMC SMBus controller to HW I2C

alewycky-tenstorrent:
This is draft because it needs testing and need to decide if losing PEC is acceptable.


cfriedt:
Is it possible to have a testsuite to exercise the i2c cancellation mechanism?

E.g. as in https://github.com/zephyrproject-rtos/zephyr/compare/main...cfriedt:zephyr:i2c-api-settable-timeout-with-emul-impl-and-testsuite ?

cfriedt:
> > Is it possible to have a testsuite to exercise the i2c cancellation mechanism?
> > E.g. as in [zephyrproject-rtos/zephyr@main...cfriedt:zephyr:i2c-api-settable-timeout-with-emul-impl-and-testsuite](https://github.com/zephyrproject-rtos/zephyr/compare/main...cfriedt:zephyr:i2c-api-settable-timeout-with-emul-impl-and-testsuite) ?
>
> I don't think so. There's no interesting HW-independent code in the cancel implementation. And the test environment presumably can't emulate STM32 I2C HW.

It is possible. I was just trying to politely ask to include test coverage in this PR. Without test coverage, we have to consider this code non-functional.

The approach I suggested above already has the testsuite that you would include in an upstream-bound patch, and you would just need to implement the API for setting the i2c timeout for stm32. From what I gathered a while ago when I looked into it, the stm32 timeout does not apply to the entirety of the transaction but only for portions of a transaction, so that is the bit that would need to be changed in the stm32 driver.

cfriedt:
> What I would expect to see in this PR, in terms of commits, are:
> 
> 1. commit to add a configurable timeout and architecture-agnostic test to the i2c api
> 2. commit to implement the new API for the stm32 driver (as a patch for upstream)
> 3. code commit for tt-z-p that adjusts our code to use the new cancellation api
> 4. test commit for tt-z-p that activates a test (likely the testsuite provided [here](https://github.com/zephyrproject-rtos/zephyr/compare/main...cfriedt:zephyr:i2c-api-settable-timeout-with-emul-impl-and-testsuite)) for DMC

Woops - I forgot to include the commits you added for disabling / removing the forked stm32 driver.

It's looking promising though.

cfriedt:
should be merged in #478

PR #414, opened 2025-07-17
drivers: mspi: mspi_dw: add support for polling mode

danieldegrasse:
After some time with tracing, I am convinced Synopsis has won. The SPI DW peripheral simply has too small of a TX FIFO (and interrupt latency is too high) to support interrupt driven transfers.

Add support for a polling mode for the MSPI DW driver, which performs transfers with interrupts locked. This should minimize risk of TX FIFO underflows, provided that the SPI frequency is slow enough the CPU can keep pace with the FIFO.

### DIFF
--- a/drivers/mspi/mspi_dw.c
+++ b/drivers/mspi/mspi_dw.c
@@ -269,6 +269,10 @@ static bool mspi_dw_transfer(const struct device *dev, uint32_t int_status)
 		 * shifting out the last frame (the last interrupt occurs when
 		 * the TX FIFO is empty). Wait if it signals that it is busy.
 		 */
+
+		while ((read_sr(dev) & SR_TFE_BIT) == 0) {
+			/* Wait until TX FIFO is empty */
+		}
 		while (read_sr(dev) & SR_BUSY_BIT) {
 		}
 
--- a/drivers/mspi/mspi_dw.c
+++ b/drivers/mspi/mspi_dw.c
@@ -1232,6 +1232,8 @@ static int api_timing_config(const struct device *dev,
 			     const uint32_t param_mask, void *cfg)
 {
 	if (param_mask & MSPI_DW_RX_TIMING_CFG) {
+		/* Disable controller before configuring */
+		write_ssienr(dev, 0);
 		write_rx_sample_dly(dev, (uint32_t)(uintptr_t)cfg);
 		return 0;
 	}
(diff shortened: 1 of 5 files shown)


drivers/mspi/mspi_dw.c
```diff
@@ -107,6 +107,7 @@ DEFINE_MM_REG_RD(rxflr,		0x24)
 DEFINE_MM_REG_RD(sr,		0x28)
 DEFINE_MM_REG_WR(imr,		0x2c)
 DEFINE_MM_REG_RD(isr,		0x30)
+DEFINE_MM_REG_RD(risr, 0x34)
```
cfriedt, line 110: It looks like this is maybe misaligned.

cfriedt (dismissed):

drivers/mspi/mspi_dw.c
```diff
@@ -956,6 +967,21 @@ static int start_next_packet(const struct device *dev, k_timeout_t timeout)
...
+	/*
+	 * The SPI DW peripheral has a nasty implementation of CS. If
+	 * the TX FIFO empties, than the peripheral will deassert CS.
+	 * to avoid interrupts preempting us, we lock interrupts and poll out
+	 * the TX data while reading RX
+	 */
+	key = irq_lock();
+	/* Write SER to start the transfer */
+	write_ser(dev, BIT(dev_data->dev_id->dev_idx));
+	while (!mspi_dw_transfer(dev, read_risr(dev) & imr)) {
```
afongTT, line 980: Polling with IRQ locked seems very unfortunate.

afongTT:
Do interrupt priorities + preemption not help with this issue?

Is it just the latency of the SPI ISR itself? Maybe we could configure the SPI clock to be slower on write?

Alternatively we could mess with the page size in the flash driver to make each write fit completely in our TX FIFO, but that seems to be propagating the issue across different drivers.

danieldegrasse:
> Do interrupt priorities + preemption not help with this issue?
> 
> Is it just the latency of the SPI ISR itself? Maybe we could configure the SPI clock to be slower on write?
> 
> Alternatively we could mess with the page size in the flash driver to make each write fit completely in our TX FIFO, but that seems to be propagating the issue across different drivers.

Interrupt priorities can help, yeah. If we have tracing off things still seem reliable as well. The specific case I saw was that with tracing on, the additional latency from logging ISR events would cause the TX FIFO to drain

Lowering the SPI clock may help as well- I just updated #243 to just have the SPI read/write test- if we can get that passing using interrupts we can probably avoid using polling mode

PR #416, opened 2025-07-17
fw_table: prevent NULL pointer access

nxuTT:
Return fake table if FW table loading fails to prevent NULL pointer access


alewycky-tenstorrent:
Please flatten the nesting, e.g.
```
if (!data->initialized
    && tt_load_table_whatever() != 0
    && !CONFIG(RECOVERY)) {
    return NULL;
}

alewycky-tenstorrent (approved):

PR #417, opened 2025-07-17
tt-boot-fs

mshimozatoTT:
tt-boot-fs is never initialized at the beginning of the boot process which prevents tables (fwtable, flash_only table, read-only table) from being loaded at boot-up. 
This PR designs the new tt-boot-fs stateless API. By passing the device (const struct device *dev) directly to ls, read and write, any part of the code can read from the flash at any time, as long as it has a pointer to the device. Eliminates the need for a cache by calling tt_boot_fs_ls which reads the entire fd table directly from the SPI flash device. 

Done:
- create a test suite for the new tt-boot-fs API 
- design and write the new tt-boot-fs API

To be done:
- integrate the new API
- refactoring will allow the bh_fwtable driver to load the firmware table, read-only table and flash-info table at init time


include/tenstorrent/tt_boot_fs.h
```diff
@@ -74,21 +69,44 @@ typedef enum {
...
+ * @brief Find file descriptor by image tag in provided array
+ *
+ * @param tag Image tag to search for (null-terminated string)
+ * @param fds Array of file descriptors to search
+ * @param count Number of file descriptors in array
+ *
+ * @return Pointer to matching file descriptor, or NULL if not found
+ * @retval NULL Tag not found or invalid parameters
+ */
+const tt_boot_fs_fd *tt_bootfs_ng_find_fd_by_tag(const uint8_t *tag, tt_boot_fs_fd *fds, int count);
```
cfriedt, line 109: I'm thinking that this function would make more sense if it were to look up a single file descriptor based on the tag provided. Does that make sense?

In that scenario, a flash device would still need to be passed in, `count` would not be necessary, and `fd` would be an output parameter.

```suggestion
/**
 * @brief Find a tt-boot-fs file descriptor by name on a given flash device.
 *
 * @param flash_dev flash device containing the tt_boot_fs to search through
 * @param tag name of the image to search for
 * @param[out] fd pointer to memory where the file descriptor will be written, if found
 *
 * @retval 0 on success
 * @retval -EIO if an I/O error occurs
 * @retval -ENOENT if no file was found matching the specified tag
 */
int tt_bootfs_ng_find_fd_by_tag(const struct device *flash_dev, const char *tag, tt_boot_fs_fd *fd);
```

The implementation of this function could be as straightforward as
```cpp
int tt_bootfs_ng_find_fd_by_tag(const struct device *flash_dev, const char *tag, tt_boot_fs_fd *fd)
{
	int ret;
	tt_boot_fs_fd fds[CONFIG_TT_BOOT_FS_IMAGE_COUNT_MAX];

	ret = tt_boot_fs_ng_ls(flash_dev, fds, ARRAY_SIZE(fds), 0);
	if (ret < 0) {
		return ret;
	}
	
	ARRAY_FOR_EACH_PTR(fds, it) {
		if (strncmp(tag, fd->image_tag, sizeof(fd->image_tag)) == 0) {
			*fd = *it;
			return 0;
		}
	}
	
	return -ENOENT;
}
```

include/tenstorrent/tt_boot_fs.h
```diff
@@ -74,21 +69,44 @@ typedef enum {
...
 
-uint32_t tt_boot_fs_cksum(uint32_t cksum, const uint8_t *data, size_t size);
+/**
+ * @brief List all file descriptors in boot filesystem
+ *
+ * @param dev Flash device containing the boot filesystem
+ * @param fds Output array to store file descriptors
+ * @param nfds Maximum number of file descriptors to read
+ */
+int tt_bootfs_ls(const struct device *dev, tt_boot_fs_fd *fds, size_t nfds);
```
cfriedt, line 97:
```suggestion
/**
 * @brief List all file descriptors in boot filesystem
 *
 * Read up to @p nfds file descriptors from a boot filesystem on flash device @p flash_dev starting from index @p offset. If reading from @p flash_dev causes an error, then this function will return `-EIO`. If @p flash_dev does not contain a valid boot fs, this function returns `-ENXIO`. On success, the number of file descriptors is returned.
 *
 * This function may also be used to count the number of files that exist on a boot filesystem if @p fds is `NULL`. In that case, the `nfds` and `offset` parameters are ignored.
 *
 * @param dev Flash device containing the boot filesystem
 * @param fds Output array to store file descriptors, or `NULL` to count files
 * @param nfds Maximum number of file descriptors to read
 * @param offset File index from which to begin reading file descriptors
 *
 * @return the number of file descriptors successfully read or a negative error code on failure.
 */
int tt_bootfs_ls(const struct device *dev, tt_boot_fs_fd *fds, size_t nfds, size_t offset);
```

include/tenstorrent/tt_boot_fs.h
```diff
@@ -74,21 +69,44 @@ typedef enum {
...
+ * @brief Read data from the boot filesystem at a specified address
+ *
+ * @param dev Flash device to read from
+ * @param addr Flash address to read from
+ * @param buffer Output buffer to store read data
+ * @param size Number of bytes to read
+ */
+int tt_bootfs_ng_read(const struct device *dev, uint32_t addr, uint8_t *buffer, size_t size);
+
+int tt_bootfs_ng_write(const struct device *dev, uint32_t addr, uint8_t *buffer, size_t size);
```
cfriedt, line 86: I'm not entirely sure if these are useful currently, since it looks like they just duplicate the flash read and flash write apis.

tests/lib/tenstorrent/tt_boot_fs/src/main.c
```diff
@@ -5,8 +5,83 @@
...
+	struct {
+		uint32_t spi_addr;
+		const char *image_tag;
+		uint32_t size;
+		uint32_t copy_dest;
+		uint32_t data_crc;
+		uint32_t flags;
+		uint32_t fd_crc;
+	} expected_fds[] = {
+		{81920, "cmfwcfg", 56, 0, 2158370831, 56, 4168430605},
```
cfriedt, line 42: These seem kind of like magic numbers - where do they come from?

tests/lib/tenstorrent/tt_boot_fs/src/main.c
```diff
@@ -5,8 +5,83 @@
...
+		{176128, "ethfwcfg", 512, 0, 2352493, 512, 3455414089},
+		{180224, "ethfw", 34304, 0, 433295191, 34304, 2151631411},
+		{217088, "memfwcfg", 256, 0, 15943, 256, 3453442091},
+		{221184, "memfw", 10032, 0, 3642299916, 10032, 1066009376},
+		{233472, "ethsdreg", 1152, 0, 897437643, 1152, 273632020},
+		{237568, "ethsdfw", 19508, 0, 3168980852, 19508, 818321009},
+		{258048, "bmfw", 35744, 0, 2928587200, 35744, 637115074},
+		{294912, "flshinfo", 4, 0, 50462976, 4, 3672136659},
+		{299008, "failover", 65828, 268435456, 2239637331, 33620260, 1985122380},
+		{16773120, "boardcfg", 0, 0, 0, 0, 3670524614},
```
cfriedt, line 53: These all seem kind of like magic numbers. Where do they come from?

cfriedt: 😆 It seems like forever ago that I wrote that in python

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+	}
+
+	size_t count = 0;
+	size_t found = 0;
+
+	/* Iterate through the fds now in RAM, starting from offset */
+	for (size_t i = 0; i < CONFIG_TT_BOOT_FS_IMAGE_COUNT_MAX; i++) {
+		tt_boot_fs_fd *current_fd = &temp_fds[i];
+
+		if (current_fd->flags.f.invalid) {
```
cfriedt, line 204:
```suggestion
		/* start reading after offset */
		if (i < offset) {
			continue;
		}

		if (current_fd->flags.f.invalid) {
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+			(uint8_t *)current_fd, sizeof(tt_boot_fs_fd) - sizeof(uint32_t),
+			current_fd->fd_crc, false);
+
+		if (chk_res == TT_BOOT_FS_CHK_OK) {
+			found++;
+		}
+	}
+
+	/* If no valid descriptors found at all, treat as not a valid boot fs */
+	if (found == 0 && offset == 0) {
```
cfriedt, line 218:
```suggestion
	if (found == 0) {
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+			break;
+		}
+
+		tt_checksum_res_t chk_res = calculate_and_compare_checksum(
+			(uint8_t *)current_fd, sizeof(tt_boot_fs_fd) - sizeof(uint32_t),
+			current_fd->fd_crc, false);
+
+		if (chk_res == TT_BOOT_FS_CHK_OK) {
+			found++;
+		}
```
cfriedt, line 214:
```suggestion
		if (chk_res == TT_BOOT_FS_CHK_OK) {
			found++;
		}
		
		/* a checksum is invalid - not a valid tt boot fs */
		return -ENXIO;
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+
+	tt_boot_fs_fd temp_fds[CONFIG_TT_BOOT_FS_IMAGE_COUNT_MAX];
+
+	if (flash_read(dev, TT_BOOT_FS_FD_HEAD_ADDR, (uint8_t *)temp_fds, sizeof(temp_fds)) != 0) {
+		return -EIO;
+	}
+
+	if (offset >= CONFIG_TT_BOOT_FS_IMAGE_COUNT_MAX) {
+		return -EINVAL;
+	}
```
cfriedt, line 195:
```suggestion
	if (nfds == 0) {
		return 0;
	}
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+
+		if (chk_res == TT_BOOT_FS_CHK_OK) {
+			found++;
+		}
+	}
+
+	/* If no valid descriptors found at all, treat as not a valid boot fs */
+	if (found == 0 && offset == 0) {
+		return -ENXIO;
+	}
```
cfriedt, line 220: If any invalid file descriptor is found (aside from a possible terminating FD?) then we should probably return an error.

```suggestion
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+
+	if (offset >= found) {
+		return -EINVAL;
+	}
+
+	for (size_t i = offset; i < found && count < nfds; i++) {
+		fds[count++] = temp_fds[i];
+	}
+
+	return count;
```
cfriedt, line 238:
```suggestion
	return found;
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+{
+	if (!dev || !device_is_ready(dev)) {
+		return -ENXIO;
+	}
+
+	tt_boot_fs_fd temp_fds[CONFIG_TT_BOOT_FS_IMAGE_COUNT_MAX];
+
+	if (flash_read(dev, TT_BOOT_FS_FD_HEAD_ADDR, (uint8_t *)temp_fds, sizeof(temp_fds)) != 0) {
+		return -EIO;
+	}
```
cfriedt, line 191:
```suggestion
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+	if (offset >= CONFIG_TT_BOOT_FS_IMAGE_COUNT_MAX) {
+		return -EINVAL;
+	}
+
+	size_t count = 0;
+	size_t found = 0;
+
+	/* Iterate through the fds now in RAM, starting from offset */
+	for (size_t i = 0; i < CONFIG_TT_BOOT_FS_IMAGE_COUNT_MAX; i++) {
+		tt_boot_fs_fd *current_fd = &temp_fds[i];
```
cfriedt, line 202:
```suggestion
		tt_boot_fs_fd *current_fd = (fds == NULL) ? &temp_fd : &fds[found];
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+
+		if (current_fd->flags.f.invalid) {
+			break;
+		}
+
+		tt_checksum_res_t chk_res = calculate_and_compare_checksum(
```
cfriedt, line 208:
```suggestion
		ret = calculate_and_compare_checksum(
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+	if (flash_read(dev, TT_BOOT_FS_FD_HEAD_ADDR, (uint8_t *)temp_fds, sizeof(temp_fds)) != 0) {
+		return -EIO;
+	}
+
+	if (offset >= CONFIG_TT_BOOT_FS_IMAGE_COUNT_MAX) {
+		return -EINVAL;
+	}
+
+	size_t count = 0;
+	size_t found = 0;
```
cfriedt, line 198:
```suggestion
```

lib/tenstorrent/boot_fs/tt_boot_fs.c
```diff
@@ -175,3 +177,90 @@ int tt_boot_fs_get_file(const tt_boot_fs *tt_boot_fs, const uint8_t *tag, uint8_
...
+
+	/* Iterate through the fds now in RAM, starting from offset */
+	for (size_t i = 0; i < CONFIG_TT_BOOT_FS_IMAGE_COUNT_MAX; i++) {
```
cfriedt, line 201:
```suggestion
	/* Skipping to offset, iterate through the whole fs, when fds is NULL, or at most nfds */
	for (size_t i = 0, found = 0; (fds == NULL) || (found < nfds); i++, found++) {
```

tests/lib/tenstorrent/tt_boot_fs/src/main.c
```diff
@@ -66,4 +146,96 @@ ZTEST(tt_boot_fs, test_tt_boot_fs_cksum)
...
+	ztest_test_skip();
+	return;
+#endif
+
+	static tt_boot_fs_fd fds[MAX_FDS];
+
+	/* Test 1: NULL device */
+	int result1 = tt_boot_fs_ls(NULL, fds, MAX_FDS, 0);
+
+	zassert_equal(result1, -ENXIO, "Should return -ENXIO for NULL device");
```
cfriedt, line 161: Comment is optional in many places, since tests should be pretty self-descriptive. I typically like to add a comment separating "degenerate" test cases, where they are expected to fail, and legitimate test cases, where they are expected to pass.

```suggestion
	zexpect_equal(tt_boot_fs_ls(NULL, fds, MAX_FDS, 0), -ENXIO);
```

cfriedt: Also, "permuting" through the range of function inputs is _incredibly powerful_.

dev:
- `NULL`, or valid flash device

fds:
- NULL, or valid array of `tt_boot_fs`

nfds:
- 0, 1, ... MAX_FDS - 1, MAX_FDS (limit to boundary conditions)

offset:
- 0, 1, ... MAX_FDS - 1, MAX_FDS, SIZE_MAX

2 * 2 * 4 * 5 = 80 possible combinations! 😮 

It's a little bit crazy to think that there might be that many combinations of input to test, but if you do it, then you can basically guarantee that there is 100% test coverage for that function.

That means, no hidden surprises in the future, basically always works as advertised, and never needs to be revisited again 😎

Of course, what is the best way to go through all of those test cases? A for loop 😁 

```cpp
static const struct test_spec {
	struct device *dev;
	tt_boot_fs_fd *fds;
	size_t nfds;
	size_t offset;
	int expect;
} specs[] = {
	/* build a truth table out of the input variables */
	DECL_TEST_SPEC(NULL, NULL, 0, 0, -ENXIO),
	/* ... */
};

ARRAY_FOR_EACH(specs, i) {
	const struct test_spec *spec = &specs[i];

	int actual = tt_boot_fs_ls(spec->dev, spec->fds, spec->nfds, spec->offset);
	zexpect_equal(actual, spec->expect, "%zu: actual: %d expected: %d", i, actual, spec->expect);
}
```

Technically speaking for pointer inputs, there are other kinds of tests that we could have where we try to feed the API bogus pointers. This is where we get into e.g. `CONFIG_USERSPACE`, `__syscall`, and other things, where the kernel does pointer validation to ensure that the current thread has permission to write and / or read to and / or from the specified memory location.

That's probably beyond the scope of this PR though, so we don't need to get into it.

tests/lib/tenstorrent/tt_boot_fs/src/main.c
```diff
@@ -5,8 +5,102 @@
...
+	printk("FLASH_DEVICE: %p\n", FLASH_DEVICE);
+	printk("Flash device name: %s\n", FLASH_DEVICE->name);
+	printk("Flash device ready: %s\n", device_is_ready(FLASH_DEVICE) ? "YES" : "NO");
+	zassert_not_null(FLASH_DEVICE, "FLASH_DEVICE is NULL!");
+
+	static tt_boot_fs_fd fds[MAX_FDS];
+
+	memset(fds, 0, sizeof(fds));
+
+	struct {
```
cfriedt, line 52: If you make this static const, then it won't take memory on the stack.

tests/lib/tenstorrent/tt_boot_fs/src/main.c
```diff
@@ -7,9 +7,89 @@
 #include <zephyr/ztest.h>
 #include <zephyr/device.h>
 #include <zephyr/drivers/flash.h>
+#include <zephyr/devicetree.h>
 #include <string.h>
 #include <tenstorrent/tt_boot_fs.h>
 
+#define SPI_FLASH_NODE DT_NODELABEL(spi_flash)
+
+#if DT_NODE_EXISTS(SPI_FLASH_NODE) && DT_NODE_HAS_STATUS(SPI_FLASH_NODE, okay)
```
cfriedt, line 16: we should always have a flash device for this test.

PR #418, opened 2025-07-17
lib: tenstorrent: bh_arc: add flash-aware spi flash data loading algorithm

alayander:
Add generic flash-aware algorithm to load Mrisc, Ethernet, and Serdes Ethernet firmware from spi-flash into a buffer in stack memory (rather than statically allocated `large_sram_buffer`) in chunks, which then transfers to the desired tile via TLB.

Follow up to: https://github.com/tenstorrent/tt-zephyr-platforms/pull/402

### DIFF
--- a/doc/release/release-notes-19.0.md
+++ b/doc/release/release-notes-19.0.md
@@ -7,7 +7,10 @@
 
 Major enhancements with this release include:
 
-[comment]: <> (H3 Performance Improvements, if applicable)
+### Performance Improvements
+
+* Load spi flash images using temporary buffer rather than large statically allocated SRAM buffer
+
 [comment]: <> (H3 New and Experimental Features, if applicable)
 [comment]: <> (H3 External Project Collaboration Efforts, if applicable)
 


cfriedt (dismissed):
@alayander - can you restructure your commits like this?

1. code commit (add functions)
2. test commit (add ztests for new functions) + enable in CI
3. refactor existing code commit

I think commit 2 should probably add a test in `tests/lib/tenstorrent/tt_boot_fs/src/main.c`.

In `tests/lib/tenstorrent/tt_boot_fs/testcase.yaml`, you will need to move the `platform_allow` bits to the existing test configurations and then add a new test configuration with platform_allow set to the various blackhole cards.
`lib.tenstorrent.boot_fs: {}`.

Also, if CI is ever failing for an unexplained reason (such as the weird bugs we have which seem to manifest in some incoherent state), please just add a blob like this at the bottom of your PR description.

Testing Done:
```shell
west or twister command + output showing test case was successful.
```

test-conf/tests/drivers/flash/common/tt_blackhole_tt_blackhole_smc.conf
```diff
@@ -2,3 +2,4 @@ CONFIG_FLASH=y
 CONFIG_MSPI=y
 CONFIG_SPI=n
 CONFIG_FLASH_MSPI_NOR_RUNTIME_PROBE=y
+CONFIG_HEAP_MEM_POOL_SIZE=8192
```
cfriedt, line 5: You could probably just add this to the board configuration file. It's easier than adding it in so many other places.

lib/tenstorrent/bh_arc/eth.c
```diff
@@ -226,10 +234,31 @@ int LoadEthFw(uint32_t eth_inst, uint32_t ring, uint8_t *fw_image, uint32_t fw_s
...
+int LoadEthFwCfg(uint32_t eth_inst, uint32_t ring, uint32_t eth_enabled, const uint8_t *tag)
 {
-	uint32_t *fw_cfg_32b = (uint32_t *)fw_cfg_image;
+	int rc;
+	tt_boot_fs_fd fd_data;
+
+	if (tt_boot_fs_find_fd_by_tag(flash, tag, &fd_data) != TT_BOOT_FS_OK) {
+		LOG_ERR("%s (%s) failed: %d", "tt_boot_fs_find_fd_by_tag", tag, -EIO);
+		return -EIO;
+	}
```
cfriedt, line 245: It should be possible to re-use the return value from `tt_boot_fs_find_fd_by_tag()` on error.

```suggestion
	rc = tt_boot_fs_find_fd_by_tag(flash, tag, &fd_data);
	if (rc < 0) {
		LOG_ERR("%s (%s) failed: %d", "tt_boot_fs_find_fd_by_tag", tag, rc);
		return rc;
	}
```

lib/tenstorrent/bh_arc/eth.c
```diff
@@ -226,10 +234,31 @@ int LoadEthFw(uint32_t eth_inst, uint32_t ring, uint8_t *fw_image, uint32_t fw_s
...
+	uint32_t spi_address = fd_data.spi_addr;
+	uint32_t image_size = fd_data.flags.f.image_size;
+
+	uint8_t *buf = (uint8_t *)k_aligned_alloc(4, image_size);
+
+	if (buf == NULL) {
+		return -EIO;
```
cfriedt, line 252:
```suggestion
		return -ENOMEM;
```

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -248,37 +268,54 @@ static int InitMrisc(void)
...
-				SCRATCHPAD_SIZE, &fw_size) != TT_BOOT_FS_OK) {
-		LOG_ERR("%s(%s) failed: %d", "tt_boot_fs_get_file", MRISC_FW_TAG, -EIO);
-		return -EIO;
-	}
 	uint32_t dram_mask = GetDramMask();
 
 	for (uint8_t gddr_inst = 0; gddr_inst < NUM_GDDR; gddr_inst++) {
 		if (IS_BIT_SET(dram_mask, gddr_inst)) {
-			if (LoadMriscFw(gddr_inst, (uint8_t *)large_sram_buffer, fw_size)) {
+			if (LoadMriscFw(gddr_inst, MRISC_FW_TAG)) {
```
cfriedt, line 281: Possibly even more useful than refactoring and calling `LoadMriscFW()` would be to just call `spi_arc_dma_transfer_to_tile()` from this function, and to remove `LoadMriscFW()`. 

```cpp
    tt_boot_fs_fd fd; /* initialize this */
	uint8_t *buf = (uint8_t *)k_aligned_alloc(4, TEMP_SPI_BUFFER_SIZE);

	if (buf == NULL) {
		return -ENOMEM;
	}

	for (uint8_t gddr_inst = 0; gddr_inst < NUM_GDDR; gddr_inst++) {
		if (IS_BIT_SET(dram_mask, gddr_inst)) {
			/* let's look up fd offset and size from tt-boot-fs (above) and then simply re-use what we already know here */
			rc = spi_arc_dma_transfer_to_tile(flash, fd.offset, fd.size, buf, TEMP_SPI_BUFFER_SIZE,
					  SetupMriscL1Tlb(gddr_inst));
			/* not 100% sure if SetupMriscL1Tlb(gddr_inst) can ever fail, but it might be important to check for failures */
			if (rc < 0) {
				LOG_ERR("%s() failed: %d", "spi_arc_dma_transfer_to_tile", rc);
				break;
			}
		}
	}

	k_free(buf);

	return rc;
}
```

include/tenstorrent/spi_flash_buf.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include <stdint.h>
+#include <stddef.h>
+
+#include <zephyr/device.h>
+
+#define TEMP_SPI_BUFFER_SIZE 4096
+
+int spi_arc_dma_transfer_to_tile(const struct device *dev, const char *tag, uint8_t *buf,
+				 size_t buf_size, uint8_t *tlb_dst);
```
cfriedt, line 18: It should be possible to perform a `tt_boot_fs_find_fd_by_tag()` once, outside of this function, and then to pass the "file" offset and size in spi flash, so I would suggest minor changes to the signature of this function.

This will help performance-wise especially in the case where the same firmware image is loaded over and over to multiple e.g. memory controllers.  

```suggestion
int spi_arc_dma_transfer_to_tile(const struct device *dev, size_t spi_image_offset, size_t spi_image_size, uint8_t *buf, size_t buf_size, uint8_t *tlb_dst);
```

alayander:
Pushed to address requested changes.
- Moved buffer allocation and tt-boot-fs lookup out of various init functions, to save on lookup calls and allocation/deallocation.

Still scoping out whether `LoadSerdesEthRegs()` can be modified to use `spi_arc_dma_transfer_to_tile()`, since it uses `NOC2AXIWrite32()` instead of `ArcDmaTransfer()` to transfer data to the tile

alayander:
> Still scoping out whether `LoadSerdesEthRegs()` can be modified to use `spi_arc_dma_transfer_to_tile()`, since it uses `NOC2AXIWrite32()` instead of `ArcDmaTransfer()` to transfer data to the tile

I don't think there's a very clean/straightforward way of doing this, since the data from spi flash is configured to be register-value pairs (address data) that would need to be casted to SerdesRegData struct and passed through `NOC2AXIWrite32()`

lib/tenstorrent/bh_arc/eth.c
```diff
@@ -286,76 +297,108 @@ static void SerdesEthInit(void)
...
 	}
+	image_size = tag_fd.flags.f.image_size;
+	spi_address = tag_fd.spi_addr;
 
+	buf = (uint8_t *)k_aligned_alloc(4, image_size);
+
+	if (buf == NULL) {
+		LOG_ERR("%s(%d) failed: %d", "k_aligned_alloc", image_size, -ENOMEM);
+		return;
+	}
```
cfriedt, line 389: Can we drop this allocation and just use the previous one?

The main point of this change is to use a buffer that is significantly smaller than the whole image, so that we can DMA to ERISC, MRISC, etc, incrementally rather than all at once.

If necessary, add e.g.

```cpp
	__ASSERT(TEMP_SPI_BUFFER_SIZE >= image_size, "spi buffer size %zu must be larger than image size %zu", TEMP_SPI_BUFFER_SIZE, image_size);
```

lib/tenstorrent/bh_arc/eth.c
```diff
@@ -286,76 +297,108 @@ static void SerdesEthInit(void)
...
-			LoadEthFwCfg(eth_inst, ring, tile_enable.eth_enabled, large_sram_buffer,
-				     fw_size);
+			LoadEthFwCfg(eth_inst, ring, buf, tile_enable.eth_enabled, spi_address,
+				     image_size);
 			ReleaseEthReset(eth_inst, ring);
 		}
 	}
+
+	k_free(buf);
+	buf = NULL;
```
cfriedt, line 401:
```suggestion

free_buf:
	k_free(buf);
```

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -248,37 +249,70 @@ static int InitMrisc(void)
...
 			}
 		}
 	}
 
-	if (tt_boot_fs_get_file(&boot_fs_data, MRISC_FW_CFG_TAG, (uint8_t *)large_sram_buffer,
-				SCRATCHPAD_SIZE, &fw_size) != TT_BOOT_FS_OK) {
-		LOG_ERR("%s(%s) failed: %d", "tt_boot_fs_get_file", MRISC_FW_CFG_TAG, -EIO);
-		return -EIO;
+	k_free(buf);
+	buf = NULL;
```
cfriedt, line 290: Similarly here, let's use a single allocation.

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -248,37 +249,70 @@ static int InitMrisc(void)
...
+
+	/* Allocate a temp buffer to read gddr speed from cfg */
+
+	rc = tt_boot_fs_find_fd_by_tag(flash, MRISC_FW_CFG_TAG, &tag_fd);
+	if (rc < 0) {
+		LOG_ERR("%s (%s) failed: %d", "tt_boot_fs_find_fd_by_tag", MRISC_FW_CFG_TAG, rc);
+		return rc;
```
cfriedt, line 297:
```suggestion
		goto free_buf;
```

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -303,6 +337,9 @@ static int InitMrisc(void)
 		}
 	}
 
+	k_free(buf);
+	buf = NULL;
+
 	return 0;
```
cfriedt, line 343:
```suggestion
```suggestion
	rc = 0;

free_buff:
	k_free(buf);

	return rc;
``````

File: lib/tenstorrent/bh_arc/gddr.c [hunk 11]
```diff
@@ -294,7 +328,7 @@ static int InitMrisc(void)
 
 	for (uint8_t gddr_inst = 0; gddr_inst < NUM_GDDR; gddr_inst++) {
 		if (IS_BIT_SET(dram_mask, gddr_inst)) {
-			if (LoadMriscFwCfg(gddr_inst, (uint8_t *)large_sram_buffer, fw_size)) {
+			if (LoadMriscFwCfg(gddr_inst, buf, spi_address, image_size)) {
 				LOG_ERR("%s(%d) failed: %d", "LoadMriscFwCfg", gddr_inst, -EIO);
 				return -EIO;
```
cfriedt, line 333:
```suggestion
				rc = -EIO;
				goto free_buf;
```

jgrowdenTT:
Why is dynamic memory allocation better than the static array? 

I see that the size is reduced... the heap is 8KB and the scratchpad is 36KB or whatever. Can we not just make the scratchpad size 8KB?

cfriedt:
Dynamic allocation is definitely not better, it's just less permanent.

My main concern was that any static memory will be used once and never again, so it's effectively wasted space.

I actually think it might even be safe to use stack memory.

For example, why not make the buffer size 256 bytes? That should fit nicely on the stack and does not require static or dynamic allocation.

alayander:
That makes sense - I will look into modifying the implementation to use stack memory

lib/tenstorrent/bh_arc/wipe_l1.c
cfriedt, line 1: This file should never even have been merged. It's garbage.

cfriedt:
> As mentioned, I'm also completely satisfied if we make the static buffer smaller (same size as the heap in this PR), which I think has all the benefits without all the mallocing and checking around that. Although, if we start sharing that global resource between threads, we would have to eventually think about locking. Stack buffers are better in that regard.

The main reason I would advise against a global buffer is that it works against the process that we're working toward, which is converting bringup code to individual zephyr drivers (e.g. gddr.c -> memory controller, eth.c -> ethernet controller, serdes_eth.c -> probably also just part of the ethernet controller).

So as long as each "init" function allocated enough stack space (e.g. 256 or 512 bytes, 32-bit aligned) before calling the transfer function, then they would maintain independence.

cfriedt (approved):

PR #421, opened 2025-07-21
dts: arc: tenstorrent: tt_blackhole_fixed_partitions: increased padto

alexapostolu:
Increased the padto values for each image in flash partition to keep up with the growing firmware image size.

### DIFF
--- a/dts/arc/tenstorrent/tt_blackhole_fixed_partitions.dtsi
+++ b/dts/arc/tenstorrent/tt_blackhole_fixed_partitions.dtsi
@@ -15,55 +15,55 @@
 
 		cmfwcfg@14000 {
 			label = "cmfwcfg";
-			reg = <0x014000 0x50>;
+			reg = <0x014000 0x100>;
 			binary-path = "$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/fw_table.bin";
 			read-only;
 		};
 
 		origcfg@15000 {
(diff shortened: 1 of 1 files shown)


dts/arc/tenstorrent/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -15,55 +15,55 @@
...
-			reg = <0x015000 0x50>;
+			reg = <0x015000 0x100>;
 			binary-path = "$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/fw_table.bin";
 			read-only;
 		};
 
 		cmfw@16000 {
 			label = "cmfw";
-			reg = <0x016000 0x19bb0>;
+			reg = <0x016000 0x19fff>;
```
cfriedt, line 32: Why not round this to 0x1a000?

alexapostolu: I'm not sure if the continuous memory region is causing a problem, so I want to leave some gap

cfriedt: What were the failures?

alexapostolu: It was the same ones I'm getting now

PR #426, opened 2025-07-22
readme: additional fixes for readme based on user feedback

cfriedt:
Based on user feedback, fix up the readme to

1. output tt-console in a consistent location
2. use a board other than p100 (p100a is a safer base configuration)
3. note that users should change their board for different board revisions
4. add a note / link about using a patched version of openocd

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,9 @@ west blobs fetch
 # Apply local patches
 west patch apply
 
+# Install Python packages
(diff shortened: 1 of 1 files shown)


jgrowdenTT (approved):
LGTM. Thank you!

danieldegrasse (approved):

PR #427, opened 2025-07-22
ci: further uplift metal tests to v0.61

tt-rkim:
Just another good thing: move up tests to be more updated

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -15,64 +15,6 @@ jobs:
     uses: ./.github/workflows/build-fw.yml
     secrets:
       SIGNATURE_KEY: ${{ secrets.SIGNATURE_KEY }}
-  hardware-metal-test:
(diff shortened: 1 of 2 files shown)


tt-rkim:
Hey guys, thanks for comments

- removed from long CI
- signed off my commit

danieldegrasse:
Fine to remove the long test, but is there another metal test we should run overnight in the future? The idea of the long tests is that they run nightly and may take 30min-1hr to execute

cfriedt (approved):

cfriedt:
Just need CI to be green before this gets merged

danieldegrasse (approved):

tt-rkim:
> another metal test we should run overnight in the future?

I think a good candidate could be some stress tests we have. We have model-level stress tests, although I think we should see if there are other candidates from portions of the SW stack that are a bit closer to the hardware.

For now, looks like it's green.

PR #431, opened 2025-07-23
readme: remove convenience variables, make a separate test section

cfriedt:
* remove comments re convenience variables and convenience variables
* move testing to its own section

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,8 @@ all Python dependencies, and the Zephyr SDK are already installed and activated.
 
 ```shell
 # Create a west workspace
-MODULE=tt-zephyr-platforms
-west init -m https://github.com/tenstorrent/$MODULE ~/$MODULE-work
-cd ~/$MODULE-work
+west init -m https://github.com/tenstorrent/tt-zephyr-platforms ~/tt-zephyr-platforms-work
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,7 +58,7 @@ cd $MODULE
 gcc -Iinclude -std=gnu11 -o /tmp/tt-console scripts/tt-console/console.c
 
 # Build and flash firmware
-west build --sysbuild -p -b tt_blackhole@p100a/tt_blackhole/smc app/smc \
+west build --sysbuild -b tt_blackhole@p100a/tt_blackhole/smc app/smc \
```
danieldegrasse, line 61: I'd keep the `-p`- if people have to retry this step I think that is helpful. Just my two cents

cfriedt: Sure.
It shouldn't be necessary, but if it is, that probably means we aren't tracking build dependencies correctly. If that's the case, build dependencies should be fixed.

danieldegrasse: It likely isn't necessary if they get the command right first try- what I've seen in the past is that people get into weird states (ie they type the board wrong on their first attempt or something)

cfriedt: No, it isn't necessary on the first try, but it shouldn't be necessary on subsequent tries either.

PR #436, opened 2025-07-24
drivers: clock_control: integrating new PLL driver

lzhengTT:
integrating new PLL driver

Minor addition: Adding set clock controller to initial rate option for re-initializing after bypassing. Note that `clock_control_set_rate` to initial rate should only be used before initAiclkPPM() and InitMrisc().

Minor fix: with fixing k_busy_wait usage in driver

Minor patch: k_busy_wait_ns() for nano second resolution waiting

### DIFF
--- a/lib/tenstorrent/bh_arc/reset.c
+++ b/lib/tenstorrent/bh_arc/reset.c
@@ -13,7 +13,6 @@
 #include "noc.h"
 #include "noc_init.h"
 #include "noc2axi.h"
-#include "pll.h"
 #include "reg.h"
 #include "status_reg.h"
 #include "tensix_cg.h"
@@ -30,6 +29,15 @@
 #include <zephyr/logging/log.h>
 #include <zephyr/sys/util.h>
 #include <zephyr/toolchain.h>
+#include <zephyr/device.h>
+#include <zephyr/devicetree.h>
+#include <zephyr/drivers/clock_control/clock_control_tt_bh.h>
+#include <zephyr/drivers/clock_control.h>
+
+#define DT_DRV_COMPAT         tenstorrent_bh_clock_control
+#define PLL_DEVICE_INIT(inst) DEVICE_DT_INST_GET(inst),
+
+static const struct device *const pll_devs[] = {DT_INST_FOREACH_STATUS_OKAY(PLL_DEVICE_INIT)};
 
 LOG_MODULE_REGISTER(InitHW, CONFIG_TT_APP_LOG_LEVEL);
 
@@ -101,8 +109,10 @@ static int DeassertRiscvResets(void)
 	}
 
 	/* Go back to PLL bypass, since RISCV resets need to be deasserted at low speed */
-	PLLAllBypass();
-
+	ARRAY_FOR_EACH(pll_devs, i) {
+		clock_control_configure(pll_devs[i], NULL,
+					(void *)CLOCK_CONTROL_TT_BH_CONFIG_BYPASS);
+	}
 	/* Deassert RISC reset from reset_unit */
 
 	for (uint32_t i = 0; i < 8; i++) {
@@ -121,7 +131,11 @@ static int DeassertRiscvResets(void)
 	ddr_reset.f.ddr_risc_reset_n = 0xffffff;
 	WriteReg(RESET_UNIT_DDR_RESET_REG_ADDR, ddr_reset.val);
 
-	PLLInit();
+	ARRAY_FOR_EACH(pll_devs, i) {
+		clock_control_set_rate(pll_devs[i],
+				       (clock_control_subsys_t)CLOCK_CONTROL_TT_BH_INIT_STATE,
+				       (clock_control_subsys_rate_t)-1);
+	};
 
 	return 0;
 }
@@ -185,7 +199,10 @@ static int DeassertTileResets(void)
 	}
 
 	/* Put all PLLs back into bypass, since tile resets need to be deasserted at low speed */
-	PLLAllBypass();
+	ARRAY_FOR_EACH(pll_devs, i) {
+		clock_control_configure(pll_devs[i], NULL,
+					(void *)CLOCK_CONTROL_TT_BH_CONFIG_BYPASS);
+	}
 
 	RESET_UNIT_GLOBAL_RESET_reg_u global_reset = {.val = RESET_UNIT_GLOBAL_RESET_REG_DEFAULT};
 
--- a/lib/tenstorrent/bh_arc/telemetry.c
+++ b/lib/tenstorrent/bh_arc/telemetry.c
@@ -8,7 +8,6 @@
 #include "fan_ctrl.h"
 #include "functional_efuse.h"
 #include "harvesting.h"
-#include "pll.h"
 #include "pvt.h"
 #include "reg.h"
 #include "regulator.h"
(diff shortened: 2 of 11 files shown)


app/smc/prj.conf
```diff
@@ -35,3 +35,5 @@ CONFIG_BOOT_BANNER=n
 CONFIG_TT_BOOT_BANNER=y
 CONFIG_TT_BOOT_BANNER_GIT_VERSION=y
 CONFIG_TT_BOOT_BANNER_SDK_VERSION=y
+
+CONFIG_CLOCK_CONTROL=y
```
cfriedt, line 39: Add a newline after this line

cfriedt, line 39: The blank line looks to be in the wrong place here

drivers/clock_control/clock_control_tt_bh.c
```diff
@@ -344,13 +344,13 @@ static void clock_control_tt_bh_update(const struct clock_control_tt_bh_config *
 	/* Setup external postdivs */
 	clock_control_tt_bh_config_ext_postdivs(config, settings);
 
-	k_busy_wait(300);
+	k_busy_wait(0.3);
```
cfriedt, line 347: I'm surprised that this compiles without a warning, since it's an implicit cast from `float` to `uint32_t`.

Bringup code does include a `WaitNs()` function. If we can avoid using it, that would be nice.

Zephyr lacks a `k_busy_wait_ns()` function currently though, likely because few  Zephyr devices would need it.

Can you add this zephyr patch file as a workaround until `k_busy_wait_ns()` has a proper implementation?
```diff
diff --git a/include/zephyr/kernel.h b/include/zephyr/kernel.h
index 614b2eab1b9..e9752bb0fdb 100644
--- a/include/zephyr/kernel.h
+++ b/include/zephyr/kernel.h
@@ -617,6 +617,11 @@ __syscall int32_t k_usleep(int32_t us);
  */
 __syscall void k_busy_wait(uint32_t usec_to_wait);
 
+static inline void k_busy_wait_ns(uint32_t nsec_to_wait)
+{
+       k_busy_wait(k_cyc_to_us_ceil32(k_ns_to_cyc_floor32(nsec_to_wait)));
+}
+
 /**
  * @brief Check whether it is possible to yield in the current context.
  *
```

```suggestion
	k_busy_wait_ns(300);
```

cfriedt (changes requested):

lib/tenstorrent/bh_arc/telemetry.c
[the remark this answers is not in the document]
lzhengTT: Okay, the auto-formatting on save seems to go against the compliance check though

app/smc/prj.conf
cfriedt, line 39: This should probably be moved to the board `tt_blackhole_tt_blackhole_smc_defconfig`

lib/tenstorrent/bh_arc/reset.c
```diff
@@ -121,7 +131,16 @@ static int DeassertRiscvResets(void)
 	ddr_reset.f.ddr_risc_reset_n = 0xffffff;
 	WriteReg(RESET_UNIT_DDR_RESET_REG_ADDR, ddr_reset.val);
 
-	PLLInit();
+	/* PLLInit */
+	SetPostCode(POST_CODE_SRC_CMFW, POST_CODE_ARC_INIT_STEP4);
+	if (!IS_ENABLED(CONFIG_ARC)) {
+		return 0;
+	}
```
cfriedt, line 138:
```suggestion
```

lzhengTT: This logic is in the current` PLLInit()`, how come it is not necessary for the refactor?

cfriedt (dismissed):
LGTM - I'm kind of in the middle of bringing my p100 back to life, so if either @danieldegrasse or @afongTT can double check it, that would be good.

danieldegrasse:
> LGTM - I'm kind of in the middle of bringing my p100 back to life, so if either @danieldegrasse or @afongTT can double check it, that would be good.

Looking now

zephyr/patches/zephyr/k-busy-wait-ns.patch
```diff
@@ -0,0 +1,16 @@
...
+index 614b2eab1b9..e9752bb0fdb 100644
+--- a/include/zephyr/kernel.h
++++ b/include/zephyr/kernel.h
+@@ -617,6 +617,11 @@ __syscall int32_t k_usleep(int32_t us);
+  */
+ __syscall void k_busy_wait(uint32_t usec_to_wait);
+
++static inline void k_busy_wait_ns(uint32_t nsec_to_wait)
++{
++       k_busy_wait(k_cyc_to_us_ceil32(k_ns_to_cyc_floor32(nsec_to_wait)));
```
danieldegrasse, line 11: I'm concerned about the amount of rounding we do here. For us (if my math is right), the evaluation would be something like this:

```
k_busy_wait_ns(300)
= k_busy_wait(k_cyc_to_us_ceil32(k_ns_to_cyc_floor32(300)))
= k_busy_wait(k_cyc_to_us_ceil32(240))
= k_busy_wait(k_cyc_to_us_ceil32(240))
= k_busy_wait(1)
```

I think if we want a delay function at this accuracy, we need to implement `k_busy_wait_ns` similar to how `k_busy_wait` is implemented, where we get the cycle count and then delay that many NOPs. We could even move `k_busy_wait` and `k_busy_wait_ns` to use the same underlying function (that delays for a certain cycle or loop count)

https://github.com/zephyrproject-rtos/zephyr/blob/main/kernel/busy_wait.c#L12

drivers/clock_control/clock_control_tt_bh.c
```diff
@@ -344,13 +344,13 @@ static void clock_control_tt_bh_update(const struct clock_control_tt_bh_config *
 	/* Setup external postdivs */
 	clock_control_tt_bh_config_ext_postdivs(config, settings);
 
-	k_busy_wait(300);
+	k_busy_wait_ns(300);
```
danieldegrasse, line 347: See my comment about the implementation of `k_busy_wait_ns`. Since this effectively works out to `k_busy_wait(1)`, should we just use that here? Or is my timer math bad here?

drivers/clock_control/clock_control_tt_bh.c
```diff
@@ -475,6 +475,11 @@ static int clock_control_tt_bh_set_rate(const struct device *dev, clock_control_
 			clock_control_tt_bh_write_reg(config, PLL_CNTL_1_OFFSET, pll_cntl_1.val);
 			k_busy_wait_ns(100);
 		}
+	} else if (clock == CLOCK_CONTROL_TT_BH_INIT_STATE) {
```
danieldegrasse, line 478: This seems like a good use case for deferred init: https://docs.zephyrproject.org/latest/kernel/drivers/index.html#deferred-initialization

If we used this here, could we just defer the init until we need to put the PLL into initial state? Or do we need to do the device init separately from setting the initial PLL settings?

tests/drivers/flash/performance/testcase.yaml
danieldegrasse, line 1: See https://github.com/tenstorrent/tt-zephyr-platforms/pull/448- I think the reclocking we are now doing by default (ARCCLK is 800MHz not 540 MHz) is what is causing this, so we don't actually want to increase these deadlines

drivers/clock_control/clock_control_tt_bh.c
cfriedt: It rounds up to 1 us in the half-assed implementation which was my suggestion, but it will work for the time being and be API compatible with a proper solution that eventually gets upstreamed.

cfriedt:
> Only real blocking comment is the `k_busy_wait_ns` patch

That shouldn't be a blocking issue. It was my suggestion

https://github.com/tenstorrent/tt-zephyr-platforms/pull/436#discussion_r2231392755

zephyr/patches/zephyr/k-busy-wait-ns.patch
danieldegrasse: Ok, makes sense. I would really prefer we just use `k_busy_wait(1)` then. IMO it doesn't make sense to create a patch we can't upstream just to use a function we might implement later. Let's either:
- do the full implementation now and upstream
- use `k_busy_wait(1)` and add a comment that the true delay required is 300ns. We can move back to a NS delay function once one exists

lzhengTT:
okay I'll look into creating `k_busy_wait_ns` in a manner we can upstream.

tests/drivers/flash/performance/testcase.yaml
cfriedt: I rebased this on #448 and dropped the last commit.

cfriedt:
> okay I'll look into creating `k_busy_wait_ns` in a manner we can upstream.

@lzhengTT - I think that might be a bit of a deep rabbit hole. I can take care of the upstream side of that.

zephyr/patches/zephyr/k-busy-wait-ns.patch
cfriedt: @danieldegrasse - the reason I suggested to make a workaround here was because I would be doing the upstream work to add support. I think it's fine.

danieldegrasse: Ah ok, if we are going to upstream it that's fine with me then, just wasn't clear from your initial comment.

PR #439, opened 2025-07-25
blobs: update memory init parameters for galaxy

afongTT:
Reduce GDDR memory speed to 14G for Blackhole Galaxy.


afongTT:
FYI @cfriedt @danieldegrasse, the current BH Galaxy setting is 16G GDDR6 and needs to be reduced to 14G due to known issues.
@psivarajahTT and @amangatTT want this to get into 18.7.0 and I agree with them since BH Galaxy is unstable at 16G.

danieldegrasse (approved):

PR #440, opened 2025-07-25
boards: tenstorrent: improve reliablity when programming P100 flash

danieldegrasse:
Add changes to improve the reliability of programming the P100 flash. These changes were tested in #243, but this PR removes the commit adding the test and only includes the fixes.

### DIFF
--- a/drivers/mspi/mspi_dw.c
+++ b/drivers/mspi/mspi_dw.c
@@ -1197,9 +1197,30 @@ static int api_timing_config(const struct device *dev,
 			     const struct mspi_dev_id *dev_id,
 			     const uint32_t param_mask, void *cfg)
 {
+	struct mspi_dw_data *dev_data = dev->data;
+	int rc, rc2;
+
 	if (param_mask & MSPI_DW_RX_TIMING_CFG) {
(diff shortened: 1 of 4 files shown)


drivers/mspi/mspi_dw.c
```diff
@@ -1198,6 +1198,8 @@ static int api_timing_config(const struct device *dev,
 			     const uint32_t param_mask, void *cfg)
 {
 	if (param_mask & MSPI_DW_RX_TIMING_CFG) {
+		/* Disable controller before configuring */
+		write_ssienr(dev, 0);
```
afongTT, line 1202: Shouldn't this already be disabled when this timing_config API gets called?

cfriedt: I think it's fine for the driver to disable / enable interrupts inside fo `api_timing_config()`, but what I would recommend doing is backing up the previous `ssienr` bit and then restoring the previous bit when the function exits. Doing it inside this function would at least reduce the number of system calls needed to configure the device.

Of course, there should be adequate locking as well, to make sure that interrupts aren't disabled while some other transaction is mid-flight.

cfriedt (approved):

PR #448, opened 2025-07-30
tests: drivers: flash_performance: make test use average times

danieldegrasse:
Take average of 10 iterations when programming or reading flash. This should improve CI stability, since one long programming duration won't cause test failure.

Also, update the system clock cycle settings, since enabling BH_ARC by default makes the ARC clock run at 800MHz rather than 540.

### DIFF
--- a/boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc.dts
+++ b/boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc.dts
@@ -39,6 +39,11 @@
 	};
 };
 
+/* BH_ARC PLL library will raise system clock to 800MHz */
+&sysclk {
+	clock-frequency = <800000000>;
+};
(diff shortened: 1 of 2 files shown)


danieldegrasse:
This PR now extends the programming deadline for the MT25- I am not clear why this deadline has changed, because it does not appear to be violated locally (programming requires about 362ms)

boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc.dts
```diff
@@ -39,6 +39,11 @@
 	};
 };
 
+/* BH_ARC PLL library will raise system clock to 800MHz */
+&sysclk {
+	clock-frequency = <800000000>;
+};
+
```
cfriedt, line 46: I think we could probably put this in the soc dtsi

PR #450, opened 2025-07-30
ci: hardware: remove use of hardware-map definition

danieldegrasse:
We don't strictly *need* the hardware map file currently, as we only
have one DUT per test system. While the file represents a convenient way
to encode multiple flash options, it resides outside of version control
and has caused CI headaches in the past.

Encode twister run options directly into commands used in CI and use
tt-console built from source, so that all portions of the test are version
controlled.


cfriedt (approved):
Not entirely sure how we want this to work for P300 or other systems with e.g. 2xP150 or e.g. 2xP100.

I guess we can cross that bridge when we get to it though.

danieldegrasse:
> Not entirely sure how we want this to work for P300 or other systems with e.g. 2xP150 or e.g. 2xP100.
> 
> I guess we can cross that bridge when we get to it though.

Kinda my thought as well- I like the idea of generation that you mentioned.

PR #452, opened 2025-07-30
scripts: tooling: tt-console: support rootless rescan via IOCTL

danieldegrasse:
Support rootless rescan of the PCIe bus using an IOCTL- if this rescan fails, we will fall back to a standard rescan using the PCIe sysfs (which does require root access)


scripts/tooling/rescan.c
```diff
@@ -0,0 +1,276 @@
...
+		/* No need for full PCIe rescan */
+		goto out;
+	}
+
+	/* Now try PCIe rescan */
+	(void)pcie_remove();
+	ret = pcie_rescan();
+out:
+	close(fd);
+	return ret;
```
cfriedt, line 275: This should go into a separate function, maybe `pcie_rescan_sysfs()`

danieldegrasse: Are you asking for two functions to be exposed in `rescan.h`?

scripts/tooling/rescan.c
```diff
@@ -0,0 +1,276 @@
...
+}
+
+/**
+ * Rescan the PCIe bus for Tenstorrent devices.
+ *
+ * @param tt_dev_name Name of the Tenstorrent device to rescan.
+ * @return 0 on success, or a negative error code on failure.
+ */
+int rescan_pcie(const char *tt_dev_name)
```
cfriedt, line 239: something like this would work better, because currently, it aborts the rescan if either open or ioctl fail
```cpp
return pcie_rescan_ioctl() || pcie_rescan_sysfs();
```

cfriedt: It might even be fine to simply use the ioctl method exclusively

scripts/tooling/rescan.c
cfriedt: two static functions. I guess if we keep the sysfs one around, then we can at least rescan if the device isn't already visible.

cfriedt (approved):

PR #453, opened 2025-07-30
ci: do not retry if twister jobs fail

cfriedt:
Historically, `twister` has required the `--retry-failed` argument and friends because of extremely-difficult-to-guarantee synchronization between the host clock and the "hardware clock" in Qemu guests.

This is most prevalently seen when running kernel benchmarks, SMP locking, and SMP thread stress tests when running multiple SMP Qemu guests in parallel.

It reduces to the fact that one Qemu instance will make assumptions about host CPU availability that break down under heavy load. That causes jitter in the guest "hardware clock", typically on the order of a few ms or sometimes tens of ms.

We don't (yet) run those kinds of stress tests involving multiple Qemu targets in our CI, so retrying is really just exacerbating a problem, which is that if our hardware platform fails on the first try it typically fails on subsequent retries of the same test.

Retrying is therefore just a waste of cycles and clogs-up CI.

Let's eliminate retries until they are absolutely necessary.

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -64,8 +64,7 @@ jobs:
           # flash DMC and SMC firmware, but since each chip uses a separate
           # debug adapter this doesn't work. For now, just flash DMC
           # then run twister with SMC firmware
-          ./scripts/twister -i --retry-failed 3 \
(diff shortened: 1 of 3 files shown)


danieldegrasse (approved):
Faster ci please

PR #454, opened 2025-07-30
scripts: tooling: tt_console: fix rescan within container

danieldegrasse:
Within a container, failing to open the device /dev/tenstorrent/0 will fail with error code ENXIO because we have indicated to docker that /dev/tenstorrent/0 is a "special device" Add this errno code as a special case that triggers a rescan.

This should resolve lingering CI issues with smoke tests.


danieldegrasse:
All CI systems have been updated with the new executable built from this PR

PR #458, opened 2025-07-31
fan_ctrl: enable p300a fan control

kikkiakki:
Making this PR on Sherry's request as she is on vacation without access to her laptop.

Depends on https://github.com/tenstorrent/tt-zephyr-platforms/pull/413

### DIFF
--- a/app/dmc/src/main.c
+++ b/app/dmc/src/main.c
@@ -30,8 +30,9 @@
 #include <tenstorrent/event.h>
 #include <tenstorrent/jtag_bootrom.h>
 #include <tenstorrent/tt_smbus_regs.h>
+#include <tenstorrent/read_only_table.h>
 
-#define RESET_UNIT_ARC_PC_CORE_0                0x80030C00
+#define RESET_UNIT_ARC_PC_CORE_0 0x80030C00
 
 LOG_MODULE_REGISTER(main, CONFIG_TT_APP_LOG_LEVEL);
 
@@ -48,6 +49,9 @@ static const struct device *const max6639_pwm_dev =
 	DEVICE_DT_GET_OR_NULL(DT_NODELABEL(max6639_pwm));
 static const struct device *const max6639_sensor_dev =
 	DEVICE_DT_GET_OR_NULL(DT_NODELABEL(max6639_sensor));
+
+static uint8_t auto_fan_speed[BH_CHIP_COUNT] = { 35 }; /* per–chip */
+static uint8_t forced_fan_speed;                     /* 0-100 % */
 
 int update_fw(void)
 {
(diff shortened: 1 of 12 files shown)


cfriedt (commented):
Is there any way that we can focus on having fan control on the DMC where fan control actually resides and leave the SMC side as basically a message passing layer?

This looks overcomplicated.

What is the "echo" that the timer solves? Is it reading back a different telemetry value for fan speed than the one expected?

kikkiakki:
@cfriedt What would you move to the DMC?

SMC could just send temperatures, and the fan curve calculations can happen on the DMC - but we also have the "force fan speed" mode which we trigger from the host. We should probably keep this in place because we use this for thermal testing/characterization. So that might be the one reason why it makes sense for a target fan speed to be calculated on the SMC side instead.

For me, the major thing that overcomplicates this is the broadcast back to all chips after one chip forces the fan speed (sending `kCm2DmMsgIdForcedFanSpeedUpdate`). It stops the timer for fan speed updates across all chips until receiving notification that the fan is unforced, but we don't necessarily need to do that because the DMC can ignore all regular fan speed updates (`kCm2DmMsgIdFanSpeedUpdate` msgs). I think the problem that Sherry was trying to solve with the broadcast back (`CMFW_SMBUS_FAN_SPEED` msg) was the FAN_SPEED telemetry value, which was supposed to be target fan percentage on single-chip cards, going out of sync. What I think makes more sense for that telemetry value is to calculate it as FAN_SPEED = FAN_RPM / MAX_RPM, since FAN_RPM is already broadcast back from the DMC separately.

cfriedt:
@kikkiakki 

> SMC could just send temperatures, and the fan curve calculations can happen on the DMC - but we also have the "force fan speed" mode which we trigger from the host. We should probably keep this in place because we use this for thermal testing/characterization. So that might be the one reason why it makes sense for a target fan speed to be calculated on the SMC side instead.

Couldn't SMC just send a desired fan speed to DMC?

From a 30k foot perspective, DMC is really acting as a proxy to the fan controller in many ways, so is basically repeating control directives (on non-p100 cards), but it's also not an i2c "peripheral" of the SMC, the SMC is an i2c peripheral of the DMC, which makes this very messy.

> For me, the major thing that overcomplicates this is the broadcast back to all chips (`CMFW_SMBUS_FAN_SPEED` msg) after one chip forces the fan speed (sending `kCm2DmMsgIdForcedFanSpeedUpdate`). It stops the timer for fan speed updates across all chips until receiving notification that the fan is unforced, but we don't necessarily need to do that because the DMC can ignore all regular fan speed updates (`kCm2DmMsgIdFanSpeedUpdate` msgs).

Is there a design doc or something that describes the broadcast in a bit more detail? Slack me, if there is, because I would really like to understand it better.

> I think the problem that Sherry was trying to solve with the broadcast back was the FAN_SPEED telemetry value, which was supposed to be target fan percentage on single-chip cards, going out of sync.

That's what I was mainly trying to address. Since it is not a synchronous operation, that's basically unavoidable unless there is an entire round trip where the SMC (and potentially host) get an ACK or a NAK.

There are mainly 3 different ways of propagating a "set fan speed" message from the host through the SMC, to the DMC, to the fan controller.

1. entirely synchronous (host, smc, dmc all block until the fan control operation has been performed, error or success is propogated all the way back)
  - likely not possible, because it looks like we relaying this message to DMC in an interrupt handler or workqueue and cannot block
  - also, I don't believe there is a straight forward way to propogate any error back all the way, but I may be wrong
2. asynchronous with submission and completion queues (requires transaction ids)
  - likely would require a major overhaul, so not realistic
3. polling only (set fan speed, periodically check that fan speed looks reasonable, or simply fire and forget) 
  - seems like the only option at this point
  - but if it's a broadcast to each SMC on the SMBus, then wouldn't each SMC only act on the message once?
  - probably some details in a design doc that I'm missing

kikkiakki:
I've closed this until further notice because I'd rather let Sherry own the PR and drive this when she's back.

PR #460, opened 2025-07-31
ci: enable the rtt log backend to capture logs when we fail to enumerate

cfriedt:
The default mode for the RTT logging backend is to _block_ which means, that if another process on the host is not actively emptying the RTT log buffer, it could prevent all threads on the device from executing, depending on the log mode.

Default the log backend to overwrite, as that would mean we lose older messages in favour of newer ones.

The main thing we would like to use the RTT log backend for is to ensure we capture some logs in CI systems if PCIe ever fails to enumerate. We can then dump those logs in the case that the CI job fails. Once PCIe enumerates, we really don't need to care about the RTT log backend any longer, since the virtual uart will still mostly work even if firmware has hung. Eventually, we will likely enable dynamic log configuration, which will allow us to programatically disable the RTT log backend after PCIe has enumerated (from the application).

Testing Done:
```shell
$ west build -S rtt-logging -p -b tt_blackhole@p100/tt_blackhole/smc app/smc -- -DCONFIG_SHELL=y
...
# ensure that overwrite mode is default for blackhole
$ grep LOG_BACKEND_RTT_MODE build/zephyr/.config 
# CONFIG_LOG_BACKEND_RTT_MODE_DROP is not set
# CONFIG_LOG_BACKEND_RTT_MODE_BLOCK is not set
CONFIG_LOG_BACKEND_RTT_MODE_OVERWRITE=y
```

```shell
# ensure that logs can be fetched via rtt
$ west flash --openocd=/opt/tenstorrent/bin/openocd-rtt
$ west rtt --openocd=/opt/tenstorrent/bin/openocd-rtt --rtt-no-reset
...
      :+=:
          .:.               ::
          .===-:        .-===-
          .=======:. :-======-
          .==================-
          .==================-
           ==================:
            :-==========-:.
                .:====-.

ooting tt_blackhole with Zephyr OS v4.2.0 ***
*** TT_GIT_VERSION v18.7.0-19-gb9819e3306e1 ***
*** SDK_VERSION zephyr sdk 0.17.2 ***
Tenstorrent Blackhole CMFW 0.16.99
[00:00:00.306,000] <inf> throttler: Throttler 0 limit set to 150
[00:00:00.306,000] <inf> throttler: Throttler 1 limit set to 220
[00:00:00.306,000] <inf> throttler: Throttler 2 limit set to 200
[00:00:00.306,000] <inf> throttler: Throttler 3 limit set to 90
[00:00:00.306,000] <inf> throttler: Throttler 4 limit set to 150
[00:00:00.306,000] <inf> throttler: Throttler 5 limit set to 85
[00:00:00.338,000] <inf> throttler: Cable Power Limit: 300
[00:00:00.338,000] <inf> throttler: Throttler 4 limit set to 50
```

### DIFF
--- a/test-conf/samples/hello_world/sample.yaml
+++ b/test-conf/samples/hello_world/sample.yaml
@@ -15,8 +15,6 @@ common:
       - "Hello World! (.*)"
 tests:
   sample.basic.helloworld.rtt:
-    filter: CONFIG_HAS_SEGGER_RTT
-    required_snippets:
-      - rtt-console
+    filter: CONFIG_USE_SEGGER_RTT
     tags:
       - rtt


jgrowdenTT (approved):

zephyr/patches.yml
```diff
@@ -132,3 +132,17 @@ patches:
     upstreamable: false
     comments: |
       Add k_busy_wait_ns() helper for nanosecond busy wait support.
+  - path: zephyr/rtt-logging-snippet.patch
```
danieldegrasse, line 135: Can we make this a downstream snippet? The upstream PR has a change request, so I imagine we will be carrying this patch for a bit

cfriedt: We might even just be able to make them permanent Kconfigs.

cfriedt:
This probably won't make it into rc1, but I think it's safe to add anytime when the rando build failure is resolved.

danieldegrasse:
Try pulling in https://github.com/tenstorrent/tt-zephyr-platforms/pull/362/commits/faf7bbf3dd6bd9ecb2b191e99fd8f8b490b6a892. I think this is the same issue I ran into there, maybe it will fix this for you as well.

danieldegrasse (approved):

PR #462, opened 2025-08-01
drivers: clock_control: clock_control_tt_bh: fix set_aiclk handler

danieldegrasse:
Set AICLK handler needs to account for the postdiv on the PLL clock
(defaults to 4) or the AICLK frequency will be set to a quarter of the
expected value.

Update the AICLK setter to handle the external postdiv.

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -36,6 +36,7 @@ ARC_START_TIME_REG_ADDR = 0x80030440
 # ARC messages
 ARC_MSG_TYPE_REINIT_TENSIX = 0x20
 ARC_MSG_TYPE_FORCE_AICLK = 0x33
+ARC_MSG_TYPE_GET_AICLK = 0x34
 ARC_MSG_TYPE_TEST = 0x90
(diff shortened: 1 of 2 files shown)


danieldegrasse:
Provided this works correctly, the metal test execution should be ~11 minutes in this PR, versus the ~30 minute timeline we are currently seeing on mainline

alexapostolu:
Oh that's interesting, metal did speed up for this change. Was it directly because AICLK was a quarter of it's expected value this entire time?

cfriedt:
> Oh that's interesting, metal did speed up for this change. Was it directly because AICLK was a quarter of it's expected value this entire time?
> 
> I'm curious why this bug wasn't caught in the ztest,

Same - it might be worth it to try out a range of frequencies and see if the driver holds up for several of them.

@lzhengTT @alexapostolu - would either of you be able to make a follow-up PR to expand on the ZTest a bit?

danieldegrasse:
> I'm curious why this bug wasn't caught in the ztest

Is it possible the postdiv for PLL0 wasn't programmed by init code in that test? If postdiv was 1 the existing implementation would work

cfriedt (approved):

PR #470, opened 2025-08-05
tests: lib: tenstorrent: Add native_sim tests for smbus_target

jgrowdenTT:
Add native_sim testing for smbus_target. To do this, we need to fake out the reg.h interface for native sim, and provide a few different callbacks to placate the emulated target controller.


app/smc/prj.conf
```diff
@@ -26,3 +26,6 @@ CONFIG_BOOT_BANNER=n
 CONFIG_TT_BOOT_BANNER=y
 CONFIG_TT_BOOT_BANNER_GIT_VERSION=y
 CONFIG_TT_BOOT_BANNER_SDK_VERSION=y
+
+#CRC
+CONFIG_CRC=y
```
cfriedt, line 31: Can probably drop these here and just make `bh_arc` `select CRC` here
https://github.com/tenstorrent/tt-zephyr-platforms/blob/4c4c6ce2111c2080181e3b33223d9e7762ecef3f/lib/tenstorrent/bh_arc/Kconfig#L9

tests/drivers/flash/performance/prj.conf
```diff
@@ -4,3 +4,4 @@ CONFIG_FLASH=y
 CONFIG_FLASH_PAGE_LAYOUT=y
 CONFIG_ZTEST_STACK_SIZE=4096
 CONFIG_STD_C11=y
+CONFIG_CRC=y
```
cfriedt, line 7: This can probably be dropped here too with by adding `select CRC` to `bh_arc`

tests/lib/tenstorrent/bh_arc/prj.conf
```diff
@@ -3,3 +3,9 @@ CONFIG_ZTEST=y
 CONFIG_TT_BH_ARC=y
 CONFIG_TT_BOOT_FS=y
 CONFIG_NANOPB=y
+
+CONFIG_EMUL=y
+CONFIG_I2C_EMUL=y
+CONFIG_I2C_TARGET=y
+CONFIG_I2C=y
+CONFIG_CRC=y
```
cfriedt, line 11: CONFIG_CRC can be dropped here too

cfriedt:
Looks good to me - just a couple of minor nits with `CONFIG_CRC`.

cfriedt:
CI issue is probably transient.

PR #471, opened 2025-08-05
scripts: tooling: blackhole-recovery: add open source FLM

danieldegrasse:
Add source files for flash loader module used with the STM32 MCU present on blackhole cards. This module is loaded to RAM, and used by pyocd to program the EEPROM present on blackhole cards when the blackhole ASIC is hung or otherwise unresponsive.


danieldegrasse:
Added a CI job to build the flash loader module in latest push

PR #473, opened 2025-08-06
scripts: dmc_reset: fix Failed to reset DMC None error

cfriedt:
Previously, if I ran `./scripts/dmc_reset.py`, I would see this on a p100 machine. Not sure if it's the same everywhere.

```shell
./scripts/dmc_reset.py
Failed to reset DMC: None
```

Expanding the verbosity (with `-d`), i was able to determine if i took out `-c 'adapter serial auto'`, then the command would work as expected.

It seems like there was a minor inconsistency with using the default value "auto". Removing it seems to make the script work reliably again.

### DIFF
--- a/scripts/dmc_reset.py
+++ b/scripts/dmc_reset.py
@@ -69,7 +69,6 @@ def parse_args():
     parser.add_argument(
         "-i",
         "--jtag-id",
-        default="auto",
         help="Specify the JTAG ID / serial number for the debug adapter",
         metavar="ID",
     )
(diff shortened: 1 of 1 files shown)


danieldegrasse:
Do you have a hardware map file present in /opt on your system? I think this is the only way this code path would be taken

scripts/dmc_reset.py
```diff
@@ -262,7 +261,7 @@ def main():
         logger.info("No hardware map provided, using first ST-Link")
```
danieldegrasse, line 261: I think we can drop this `if` block with these lines, since jtag_id should already be set to `None` by default.

cfriedt: Fixed!

cfriedt:
> Do you have a hardware map file present in /opt on your system? I think this is the only way this code path would be taken

I do, yeah

danieldegrasse (approved):

PR #481, opened 2025-08-07
ci: move metal tests to nightly temporarily

cfriedt:
Metal tests can sometimes cause single PRs to queue for 40 minutes while waiting for a self-hosted runner to become available.

This is the case currently for both p100a and p150a, but it is much more pronounced for p150a since we only have one self-hosted runner with that hardware. There are 2 self-hosted runners with p100a hardware.

Move metal tests to nightly, since the wait time is impacting developer productivity.

This change should make our CI take closer to 5 minutes again as opposed to 20 or 40 minutes (depending on the phase of the moon).


danieldegrasse:
This seems reasonable. Can we also talk with the metal team about producing a shorter (3-4 minute) metal test? All we care about on a per-PR basis is that metal *runs*, not that a full set of workloads completes

cfriedt:
> This seems reasonable. Can we also talk with the metal team about producing a shorter (3-4 minute) metal test? All we care about on a per-PR basis is that metal _runs_, not that a full set of workloads completes

Yep - I just want to unclog the pipes for now.

PR #482, opened 2025-08-07
lib: Tenstorrent: bh_arc: Add SMBUS message for telem read and write

jgrowdenTT:
Add SMBUS messages for telem read and write which mirror the WH implementations of the same SMBUS commands.

Telemetry read is mostly the same. Telemetry write, according to the ticket, is mostly not required to implement at this point and instead we only need to relay the control data word indicating if we want a reset or not. However, the format of the message on the wire should match so 3rd party controllers expecting the prior interface should "work" for the functionality we need them to.

On that note, protocol wise, I'm attempting to match the behaviour seen in WH. Hence the not-quite-SMBUS-compliant-PEC implementation, the Block-write-block-read commands, etc. But this does leave me with some testing concerns because of the non-standard additions, and that I can only test based on my own expectations and what's in the WH documentation.

### DIFF
--- a/lib/tenstorrent/bh_arc/smbus_target.c
+++ b/lib/tenstorrent/bh_arc/smbus_target.c
@@ -494,6 +494,9 @@ static int I2CReadHandler(struct i2c_target_config *config, uint8_t *val)
 		pec = Crc8(pec, I2C_TARGET_ADDR << 1 | I2C_WRITE_BIT); /* start address byte */
 		pec = Crc8(pec, smbus_data.command);
 
+		if (curr_cmd->trans_type == kSmbusTransBlockWriteBlockRead) {
+			pec = Crc8(pec, curr_cmd->expected_blocksize_w);
+		}
 		/* any received data */
 		for (int i = 0; i < smbus_data.rcv_index; i++) {
 			pec = Crc8(pec, smbus_data.received_data[i]);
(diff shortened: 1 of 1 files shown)


app/dmc/src/main.c
```diff
@@ -230,7 +230,7 @@ static int bh_chip_run_smbus_tests(struct bh_chip *chip)
 	int pass_val = 0xFEEDFACE;
 	uint8_t count;
 	uint8_t data[32]; /* Max size of SMBUS block read */
-
+	uint32_t app_version;
```
cfriedt, line 233: Sort of surprised that this wasn't caught by the linter, but can you keep the newline under this?

jgrowdenTT:
After looking at the P300 PR I think the PEC here for block write block read is incorrect (Should take into account the block write size too) 

I'm going to take a moment to read the SMBUS spec to verify.

jgrowdenTT:
> After looking at the P300 PR I think the PEC here for block write block read is incorrect (Should take into account the block write size too)
> 
> I'm going to take a moment to read the SMBUS spec to verify.

Done.

lib/tenstorrent/bh_arc/cm2dm_msg.c
```diff
@@ -348,3 +354,47 @@ int32_t Dm2CmSendThermTripCountHandler(const uint8_t *data, uint8_t size)
 	UpdateTelemetryThermTripCount(therm_trip_count);
 	return 0;
 }
+
+int32_t Dm2CmWriteTelemetry(const uint8_t *data, uint8_t size)
+{
+	if (size != 33) {
+		return -1;
```
cfriedt, line 361: Just a general rule of thumb, if our FW is able to return an int on error, we should use a negative error value if an error occurs or zero on success.

So e.g. `-EINVAL` or something along those lines.

PR #485, opened 2025-08-07
fan_ctrl: enabling forced fan speed msg for both chips

ShLiTT:
Depends on https://github.com/tenstorrent/tt-zephyr-platforms/pull/413

This branch includes changes to senable fan control on the P300A.

**Fan controller test output:**

Initial read from the fan controller:
```
Target fan speed: 36%
Actual fan RPM: 2274 RPM
```

After forcing the fan speed to 50% and reading back:
```
(my-env) xiaoruli@yyz-syseng-36:~/work/syseng/src/t6ifc/t6py$ python packages/tenstorrent/scripts/tensix_sm/fan/fan_control.py --read --interface pci:1
Target fan speed: 50%
Actual fan RPM: 3018 RPM
```

### DIFF
--- a/app/dmc/src/main.c
+++ b/app/dmc/src/main.c
@@ -31,7 +31,7 @@
 #include <tenstorrent/jtag_bootrom.h>
 #include <tenstorrent/tt_smbus_regs.h>
 
-#define RESET_UNIT_ARC_PC_CORE_0                0x80030C00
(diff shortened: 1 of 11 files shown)


cfriedt:
AFAIK, this is already merged, so I'll close it. If that's not the case, please reopen.

PR #486, opened 2025-08-07
Updated the p150a runner to use a runner group

DRvader:


cfriedt (commented):
`p150a-jtag` is the existing runner group, like `p100a-jtag` and `p100-jtag`.

PR #489, opened 2025-08-08
zephyr: blobs: Update ERISC FW to v1.5.0

bingliTT:
- Updated some eth training sequencing to help with Galaxy UBB products
- Added CALL_ACK postcode in eth msg mailboxes to show message has been read and is being processed
- Added fence inst in eth msg mailboxes to invalidate L1 cache when polling
- Changed default training mode to AUTO mode from ANLT, will result in manual eq on Galaxy QSFP ports
- ETH msg FEATURE_ENABLE: allows for enable/disablement of eth fw features

PR #501, opened 2025-08-12
doc: release: rename 18.9 to 19.0 and adjust versions

cfriedt:
Our next release, v19.0.0 will include a major version bump from 18 to 19 to account for the fact that the SPI flash and internal DMC flash layouts will be changed.


cfriedt:
Should be merged before any PRs that add release notes.

E.g. https://github.com/tenstorrent/tt-zephyr-platforms/pull/489

PR #503, opened 2025-08-12
lib: bh_arc: spi_eeprom: drop call to tt_boot_fs_mount()

cfriedt:
* lib: bh_arc: spi_eeprom: drop call to `tt_boot_fs_mount()`
* lib: bh_arc: move spi_eeprom message handling to the application


cfriedt:
It looks like this is blocked by https://github.com/tenstorrent/tt-zephyr-platforms/pull/418

PR #504, opened 2025-08-12
tests: lib: jtag_bootrom: malloc correct length

danieldegrasse:
JTAG emul driver expects the buffer length to be provided as a count of 32 bit words, so malloc() a buffer large enough to hold the entire bootcode region


danieldegrasse:
Fixes CI errors like the following: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/16922135266/job/47952006620?pr=496

cfriedt (approved):

PR #505, opened 2025-08-12
boards: tt_blackhole: ensure fixed partitions fall under flash node

cfriedt:
Previously, fixed partitions were included as a top-level node under the root node (`/`) instead of under the `spi_flash` node (SMC) or `flash1` node (DMC).

Move `tt_blackhole_fixed_partitions.dtsi` to the board directory, since it is fairly specific to the spi flash device which is defined on a per-board basis, and ensure that it is pulled into devicetree under the correct spi flash node.


alexapostolu (approved):

jgrowdenTT (approved):

tests/drivers/flash/performance/boards/tt_blackhole_tt_blackhole_smc.overlay
```diff
@@ -8,13 +8,11 @@
...
 	partitions {
-		compatible = "fixed-partitions";
-		#address-cells = <1>;
-		#size-cells = <1>;
-
 		storage_partition: partition@2008000 {
 			label = "storage";
 			reg = <0x2008000 DT_SIZE_K(32)>;
+			/* binary-path is a required property, but we don't care what it contains */
+			binary-path = "$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/read_only.bin";
```
danieldegrasse, line 15: Could we make this property optional, and only generate a tt-boot-fs entry if the property is populated? Another option here might be to change where we set the `"tenstorrent,tt-boot-fs"` compatible to be on each child node.

danieldegrasse (commented):
Is the intention to make this the finalized partition layout, or is this PR just cleanup?

tests/drivers/flash/performance/boards/tt_blackhole_tt_blackhole_smc.overlay
cfriedt: We can, but I would rather defer that to a separate PR, since this is really fixing a bug and that would be more of a feature (it would also involve a lot of hacks regarding `tt_boot_fs.py` for generating YAML from DT).

cfriedt: @alexapostolu - can you create a ticket to make the `binary-path` property optional in the DT bindings and adjust `tt_boot_fs.py` accordingly?

PR #506, opened 2025-08-13
libs: tenstorrent: bh_arc: Smbus target compression

jgrowdenTT:
Currently SMBUS target uses a sparsely populated array of entries to store the SMBUS command handler metadata for that specific SMBUS command; the SMBUS command ID is the array index parameter, so we need 256 entries to cover.

Instead of storing the whole structure, we can store pointers to these structures then just define the handlers as needed. This costs 1KB for the pointers, but saves (256-N)*sizeof(Cmd), where N is your number of commands... Or currently, approaching 4KB. I think it also prepares us for a registration type model so that the SMBUS target processing and the actual handlers can be separate code blocks.

A SLL/DLL model would also work but would turn CMD lookup to be O(N). I discussed it with Chris and he wasn't against it, and keep in mind we can tweak it so we only hit it once per SMBUS cmd instead of on every byte as is currently. So feedback on that is welcome.

Behaviour before/after should be identical so the existing native_sim SMBUS tests should be enough to cover.


cfriedt:
LGTM - I think this is mostly already covered by the existing `native_sim` test suite, right @jgrowdenTT ?

jgrowdenTT:
> LGTM - I think this is mostly already covered by the existing `native_sim` test suite, right @jgrowdenTT ?

Yeah, exactly, and this is exactly why I wanted those tests in place.

PR #507, opened 2025-08-13
boards: tt_blackhole: correct partition labels

cfriedt:
Previously, it was not possible to use macros such as `FIXED_PARTITION_SIZE(ethfw)` because labels were slightly malformed compared to what is conventional syntax for fixed partitions.

So change things like:
```
ethfw@1234abcd {
...
};
```
to
```
ethfw : partition@1234abcd {
...
};
```

This allows us to use the `fixed-partition` Devicetree accessors available via the `<zephyr/storage/flash_map.h>` API, which converts determining tt-boot-fs file offsets and lengths a compile-time-const operation rather than a runtime operation. With that, we can perform e.g. `BUILD_ASSERT()` assertions at compile-time and avoid hitting runtime errors, we can use compile-time-const conditionals, etc.


alexapostolu (approved):
For the warning, I think it'll be solved when we transition to the new partition layout

cfriedt:
Created issue since CI is otherwise successful but fails on recovery phase after all other jobs have passed https://github.com/tenstorrent/tt-zephyr-platforms/issues/509

PR #508, opened 2025-08-13
drivers: smbus_target driver

jgrowdenTT:
Implement an SMBUS target Zephyr driver to replace the bh_arc lib implementation.

This is advantageous for a few reasons:

- We can now create multiple instances on the same bus if we wanted to 
- We can test at the driver interface layer with FFFed functions eventually
- We can theoretically prepare the smbus_target implementation for upstreaming. 
- Hardcoding around `i2c0` is resolved since it stems from the devicetree. 

The existing SMBUS native sim tests pass without too much alteration.


drivers/smbus_target/smbus_target.c
```diff
@@ -0,0 +1,389 @@
...
+ *Incoming data streams. The i2c_emul controller requires this CB be implemented,
+ *otherwise it will trip on a NULL ptr dereference.
+ */
+#if CONFIG_BOARD_NATIVE_SIM
+static int32_t smbus_write_requested(struct i2c_target_config *config)
+{
+	(void)config;
+	return 0;
+}
+#endif
```
cfriedt, line 332: It would be better to define an 'emul' variant of this that has backend hooks than to make native_sim specific ifdefs here

tests/lib/tenstorrent/bh_arc/app.overlay
```diff
@@ -7,3 +7,11 @@ fwtable: fwtable
 		flash-dev = <&flashcontroller0>;
 	};
 };
+
+&i2c0 {
```
cfriedt, line 11: This is interesting - i2c0 being the i2c endpoint that must be configured in target mode. We should add a `BUILD_ASSERT()` to check a devicetree property or something along those lines for consistency.

drivers/smbus_target/smbus_target.c
```diff
@@ -0,0 +1,389 @@
...
+#include <zephyr/drivers/i2c.h>
+#include <zephyr/sys/crc.h>
+#include <tenstorrent/smbus_target.h>
+
+#define LOG_LEVEL CONFIG_I2C_LOG_LEVEL
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(i2c_target);
+
+#define I2C_WRITE_BIT 0
+#define I2C_READ_BIT  1
```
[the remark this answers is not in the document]
jgrowdenTT: Good catch. We define these inside bh_arc lib and I didn't wan this driver source to include that so I redeclared them here. Looked around, I am pretty sure these are replaceable by `I2C_MSG_WRITE` /`I2C_MSG_READ` inside zephyr/drivers/i2c.h

Thanks!

drivers/smbus_target/smbus_target.c
jgrowdenTT: The `smbus_write_requested` can probably always go in. I can make this change and test to ensure this. In fact, the tests become a lot cleaner if we move the state setup here vs in `stop` but I'll leave that aside.

The `read` is my biggest concern. Different targets need different read implementations because some (like `i2c_emul`) say "first byte is handled by `read_requested`, the rest by `read_processed`. `i2c_dw` says "every byte is handled by `read_requested` then again by `read_processed` after the byte is sent on air." If that was aligned I have no implementation specific code, and I would have no need for this deffed out code.

Without looking, that makes me think this problem goes beyond emul. If other drivers have different interpretation of the callback contract then they're also not going to work with this supposed "driver invariant implementation" shown in this PR.

I'lll go looking. I think either way the NATIVE_SIM stuff here is hacks and needs to go.

tests/lib/tenstorrent/bh_arc/app.overlay
jgrowdenTT: I didn't actually see anywhere where we actually say "i2c0 is in target mode". Sure, `CONFIG_I2C_TARGET` needs to be defined, but I didn't need to tweak anything in the overlay for i2c0 regarding "this i2c device is a target" other than the registration done runtime... 

The "design" of having it being a child of the i2c node was lifted from the `eeprom_target` in upstream. Maybe phandles would allow us to have the same smbus_target device handle the same set of commands from multiple i2c busses(?) I think either is good.

drivers/smbus_target/smbus_target.c
jgrowdenTT: @cfriedt  I took a look through quite a few controllers (`i2c_emul`, `i2c_andes`, `i2c_gecko`, `i2c_ite_its1xxxx`) and concluded `i2c_dw` is bugged. I would like to propose a fix for it in the following way, i.e. 

```
				if (!dw->read_in_progress) {
					/* First byte of the read transfer */
					if (slave_cb->read_requested) {
						slave_cb->read_requested(dw->slave_cfg, &data);
						i2c_dw_write_byte_non_blocking(port, data);
					}
					dw->read_in_progress = true;
				} else {
					/* Subsequent bytes */
					if (slave_cb->read_processed) {
						slave_cb->read_processed(dw->slave_cfg, &data);
						i2c_dw_write_byte_non_blocking(port, data);
					}
				}
```

Where we off `read_in_progress` on the stop condition. There might be a way to do this without the SW flag, and instead use DW int signals, but I'm not sure without going thru synopsis spec.

Anyways. If I change this, I can avoid all the NATIVE_SIM stuff here and the controller should work for any i2c HW driver (so long as it adheres to the proper callback contract..)

cfriedt:
@jgrowdenTT - did you get any feedback from `tomb` (Tom Burdick) on the Zephyr discord about anything?

Ideally this would go upstream, but it would be smart to make sure we are on the right path for that.

drivers/smbus_target/smbus_target.c
cfriedt: @jgrowdenTT - the patch looks good to me. It's a good idea to get the PR open ASAP. Ideally, we can get it merged before 4.3.0 is released so we have fewer patches to carry in-tree.

tests/lib/tenstorrent/bh_arc/app.overlay
cfriedt: Yeah, either way works for me.

cfriedt:
Upstream likes changes in small pieces, so it would probably make sense to make a separate PR for the i2c_dw bug fix (might require an issue with details about how to reproduce the bug).

I think the smbus target device API is a great idea though, so that would be great to send upstream as well.

PR #516, opened 2025-08-15
app: smc: improve code reuse in E2E stress test

danieldegrasse:
Improve code reuse in E2E stress test. This is in response to a few
duplicated code sections flagged by pylint when updating the E2E stress
test for the ARC watchdog.

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -212,11 +212,10 @@ def get_int_version_from_file(filename) -> int:
     return version_int
 
 
-@pytest.mark.flash
-def test_arc_watchdog(arc_chip):
+def arc_watchdog_test(arc_chip):
     """
-    Validates that the DMC firmware watchdog for the ARC will correctly
-    reset the chip
+    Helper to run ARC watchdog test. Returns True on pass, or False on
+    failure
     """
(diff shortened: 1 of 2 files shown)


app/smc/pytest/e2e_smoke.py
```diff
@@ -267,7 +289,20 @@ def test_pcie_fw_load_time(arc_chip):
         f"PCIe init completion timestamp: {pcie_init_cpl_time}."
     )
 
-    assert duration_in_ms < 40, f"duration {duration_in_ms:.4f}ms, exceeds 40ms"
+    if duration_in_ms > 40:
+        logger.warning(f"PCIe firmware load time exceeded 40ms: {duration_in_ms:.4f}ms")
+        return False
+
+    return True
```
alexapostolu, line 296: Would it be better to have the durations as variables at the top of the file for readability?

alexapostolu (dismissed):

app/smc/pytest/e2e_smoke.py
danieldegrasse: The pattern we have right now is that variables are set at the top of tests- I'll update this code block to follow that pattern

alexapostolu (approved):
Looks good, and I think CI failures are unrelated to the PR

PR #518, opened 2025-08-15
Avoid overflow in DIV_ROUND_UP

alewycky-tenstorrent:
DIV_ROUND_UP(UINT32_MAX, 2) expands to (UINT32_MAX + 2 - 1) / 2, which due to overflow/wrap is 0/2.

Replace it with an open-coded expression that computes the desired number.

### DIFF
--- a/lib/tenstorrent/bh_arc/noc_dma.c
+++ b/lib/tenstorrent/bh_arc/noc_dma.c
@@ -106,9 +106,20 @@ static uint32_t get_expected_acks(uint32_t noc_cmd, uint64_t size)
 /* wrap around aware comparison for half-range rule */
 static inline bool is_behind(uint32_t current, uint32_t target)
 {
-	uint32_t diff = current - target;
-
-	return diff > DIV_ROUND_UP(UINT32_MAX, 2);
+	/*
+	 * "target" and "current" are NOC transaction counters, and may wrap around, so we must
+	 * consider the case where target and current have wrapped a different number of times.
+	 * There's no way to know how many times they have wrapped, instead we assume that they are
+	 * within 2**31 of each other as that gives an unambiguous ordering.
+	 *
+	 * We deal with this by considering
+	 * target - 2**31 <  current < target         MOD 2**32 as before target and
+	 * target         <= current < target + 2**31 MOD 2**32 as after target.
+	 *
+	 * We can't just check target == current because just one spurious NOC transaction could
+	 * result in the loop hanging with current = target+1.
+	 */
+	return (int32_t)(current - target) < 0;
 }
 
 static bool wait_noc_dma_done(uint32_t noc_cmd, uint32_t expected_acks)


nxuTT (approved):

lib/tenstorrent/bh_arc/noc_dma.c
```diff
@@ -102,8 +102,9 @@ static uint32_t get_expected_acks(uint32_t noc_cmd, uint64_t size)
 static inline bool is_behind(uint32_t current, uint32_t target)
 {
 	uint32_t diff = current - target;
+	uint32_t uint32_middle = UINT32_MAX / 2 + 1;
 
-	return diff > DIV_ROUND_UP(UINT32_MAX, 2);
+	return diff > uint32_middle;
```
cfriedt, line 107: what about `CLAMP()`? There are other static inline functions in Zephyr that will return `true` if an overflow occurs.

lib/tenstorrent/bh_arc/noc_dma.c
```diff
@@ -102,8 +102,9 @@ static uint32_t get_expected_acks(uint32_t noc_cmd, uint64_t size)
 static inline bool is_behind(uint32_t current, uint32_t target)
 {
 	uint32_t diff = current - target;
+	uint32_t uint32_middle = UINT32_MAX / 2 + 1;
```
danieldegrasse, line 105: If I'm understanding this code right we are trying to handle the case where `target` has wrapped around and is less than `current`, right?

If so, the C standard defines that subtraction for unsigned integers will always wrap around, so we could just check if `target - current` > 0, right?

For example, try running this C program:

```
#include <stdio.h>
#include <stdint.h>
#include <limits.h> // For UCHAR_MAX

int main() {
    uint32_t current = UINT32_MAX - 10;
    uint32_t target = current + 8;

    for (int i = 0; i < 16; i++) {
        printf("current: %u, target %u\n", current, target);
        if (current > target) {
            printf("Unsigned underflow will occur\n");
        }
        printf("target - current: %u\n", (target - current));
        target++; current++;
    }

    return 0;
}
```

I see the following output:
```
current: 4294967285, target 4294967293
target - current: 8
current: 4294967286, target 4294967294
target - current: 8
current: 4294967287, target 4294967295
target - current: 8
current: 4294967288, target 0
Unsigned underflow will occur
target - current: 8
current: 4294967289, target 1
Unsigned underflow will occur
target - current: 8
current: 4294967290, target 2
Unsigned underflow will occur
target - current: 8
current: 4294967291, target 3
Unsigned underflow will occur
target - current: 8
current: 4294967292, target 4
Unsigned underflow will occur
target - current: 8
current: 4294967293, target 5
Unsigned underflow will occur
target - current: 8
current: 4294967294, target 6
Unsigned underflow will occur
target - current: 8
current: 4294967295, target 7
Unsigned underflow will occur
target - current: 8
current: 0, target 8
target - current: 8
current: 1, target 9
target - current: 8
current: 2, target 10
target - current: 8
current: 3, target 11
target - current: 8
current: 4, target 12
target - current: 8
```

danieldegrasse: That's fine with me- mostly just trying to understand what this code does. Do we always expect `target` to be greater than `current`, except when `target` has wrapped ? IE are we just waiting until `current` catches up to `target`?

PR #519, opened 2025-08-15
lib: tenstorrent: bh_arc: support running NOC code in native_sim

danieldegrasse:
Commit d3cb9055 (security: wipe l1s, 2025-08-06) caused a regression by adding the NOC code in bh_arc into the bh_arc test built in simulation, since we now perform NOC access during system init.

Rather than reverting this commit, add support in the noc code for running within emulation. This support is *very* limited- we essentially create RAM buffers to fake out the NOC TLB and configuration spaces, and fake NOC completion where required. However, it is sufficient to fix the bh_arc tests.

Note that we need something like this, or need to revert d3cb9055 to fix CI

### DIFF
--- a/lib/tenstorrent/bh_arc/noc2axi.c
+++ b/lib/tenstorrent/bh_arc/noc2axi.c
@@ -7,8 +7,6 @@
 #include "noc.h"
 #include "noc2axi.h"
 
-#define NIU_0_A_REG_MAP_BASE_ADDR 0x80050000
-
 typedef struct {
 	uint32_t passthrough_bits: 24;
 	uint32_t lower_addr_bits: 8;
@@ -62,6 +60,15 @@ typedef union {
 #define NOC2AXI_NUM_TLB_PER_RING 16
 #define RING0_TLB_REG_OFFSET     0x1000
 #define AXI2NOC_RING_SEL_BIT     15
+
+#ifdef CONFIG_BOARD_NATIVE_SIM
+#define NIU0_A_REG_SPACE_SIZE 0x10000
+/* Running within simulation. Fake out TLB register space */
+static uint8_t fake_niu_reg_space[NIU0_A_REG_SPACE_SIZE];
+#define NIU_0_A_REG_MAP_BASE_ADDR ((uintptr_t)fake_niu_reg_space)
+#else
+#define NIU_0_A_REG_MAP_BASE_ADDR 0x80050000
+#endif
 
 static inline uint32_t volatile *GetTlbRegStartAddr(const uint8_t ring)
 {
--- a/lib/tenstorrent/bh_arc/noc2axi.h
+++ b/lib/tenstorrent/bh_arc/noc2axi.h
@@ -7,9 +7,18 @@
 #define NOC2AXI_H
 
 #include <stdint.h>
+#include <zephyr/sys/util.h>
 
+#ifdef CONFIG_BOARD_NATIVE_SIM
+/* Fake out both NOC address spaces (16 MB each) */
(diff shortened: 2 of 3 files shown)


lib/tenstorrent/bh_arc/noc2axi.c
```diff
@@ -63,6 +61,15 @@ typedef union {
 #define RING0_TLB_REG_OFFSET     0x1000
 #define AXI2NOC_RING_SEL_BIT     15
 
+#ifdef CONFIG_EMUL
+#define NIU0_A_REG_SPACE_SIZE 0x10000
+/* Running within simulation. Fake out TLB register space */
+static uint8_t fake_niu_reg_space[NIU0_A_REG_SPACE_SIZE];
```
danieldegrasse, line 67: @nxuTT can you check to make sure the address space for NIU is only `0x10000` bytes? That is my understanding based on the NOC architecture documents

lib/tenstorrent/bh_arc/noc2axi.h
```diff
@@ -7,9 +7,18 @@
 #define NOC2AXI_H
 
 #include <stdint.h>
+#include <zephyr/sys/util.h>
 
+#ifdef CONFIG_EMUL
+/* Fake out both NOC address spaces (16 MB each) */
+static uint8_t fake_noc0_window[MB(16)];
```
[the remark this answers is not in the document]
alewycky-tenstorrent: This is enough to fake 1 TLB window for each NOC. That's enough for this code.

(The full HW capability is 16 16MB windows per NOC.)

lib/tenstorrent/bh_arc/noc2axi.c
alewycky-tenstorrent: Yes. (0x8005_0000 - 0x8006_0000)

lib/tenstorrent/bh_arc/noc2axi.c
```diff
@@ -63,6 +61,15 @@ typedef union {
 #define RING0_TLB_REG_OFFSET     0x1000
 #define AXI2NOC_RING_SEL_BIT     15
 
+#ifdef CONFIG_EMUL
```
cfriedt, line 64: Usually this implies that there is an emulation driver being used.

> [NOTE]
> There is a [dma_emul driver](https://github.com/zephyrproject-rtos/zephyr/blob/main/drivers/dma/Kconfig.emul) upstream. If `noc_dma.c` simply implemented [the DMA API](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/drivers/dma.h) (i.e. was a Zephyr driver), then we wouldn't need to specifically have a fake for NOC.

danieldegrasse: Which Kconfig would you rather I use?

cfriedt: Other bringup code uses `#if CONFIG_BOARD_NATIVE_SIM`, which is still a bit dirty.

I think the reason that `CONFIG_EMUL` was used in the test was because the `i2c_emul` driver is used.

cfriedt: Not saying you should implement the DMA API in this PR, which is just a workaround for a regression, but it would have been good if the code was added that way in the first place.

danieldegrasse: Switched to use `CONFIG_BOARD_NATIVE_SIM`

alewycky-tenstorrent (approved):

PR #522, opened 2025-08-15
boards: tenstorrent: blackhole: define a fixed partition layout

danieldegrasse:
Define a fixed partition layout for the tt-boot-fs within blackhole boards. This shouldn't be merged as is, we need to get tt-boot-fs to support hex formatted images first. Mostly opening this for review of the partition layout.

### DIFF
--- a/boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
+++ b/boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
@@ -18,69 +18,63 @@ partitions {
 		label = "cmfwcfg";
 		reg = <0x14000 DT_SIZE_K(4)>;
 		binary-path = "$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/fw_table.bin";
-		read-only;
 	};
 
 	origcfg: partition@15000 {
 		label = "origcfg";
 		reg = <0x15000 DT_SIZE_K(4)>;
 		binary-path = "$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/fw_table.bin";
-		read-only;
 	};
 
 	cmfw: partition@10000000 {
 		label = "cmfw";
 		reg = <0x10000000 DT_SIZE_K(106)>;
+		offset = <0x10000000>;
 		binary-path = "$BUILD_DIR/smc/zephyr/zephyr.bin";
+		executable;
 	};
 
 	ethfwcfg: partition@30000 {
 		label = "ethfwcfg";
 		reg = <0x30000 DT_SIZE_K(4)>;
 		binary-path = "$BLOBS_DIR/tt_blackhole_erisc_params.bin";
-		read-only;
 	};
 
 	ethfw: partition@31000 {
 		label = "ethfw";
 		reg = <0x31000 DT_SIZE_K(36)>;
 		binary-path = "$BLOBS_DIR/tt_blackhole_erisc.bin";
-		read-only;
 	};
 
 	memfwcfg: partition@3a000 {
 		label = "memfwcfg";
 		reg = <0x3a000 DT_SIZE_K(4)>;
-		binary-path = "$BLOBS_DIR/tt_blackhole_gddr_params_$BOARD_REV.bin";
-		read-only;
+		binary-path = "$BLOBS_DIR/tt_blackhole_gddr_params_$PROD_NAME.bin";
 	};
 
 	memfw: partition@3b000 {
 		label = "memfw";
 		reg = <0x3b000 DT_SIZE_K(16)>;
 		binary-path = "$BLOBS_DIR/tt_blackhole_gddr_init.bin";
-		read-only;
 	};
 
 	ethsdreg: partition@3f000 {
 		label = "ethsdreg";
 		reg = <0x3f000 DT_SIZE_K(4)>;
 		binary-path = "$BLOBS_DIR/tt_blackhole_serdes_eth_fwreg.bin";
-		read-only;
 	};
 
 	ethsdfw: partition@40000 {
 		label = "ethsdfw";
 		reg = <0x40000 DT_SIZE_K(20)>;
 		binary-path = "$BLOBS_DIR/tt_blackhole_serdes_eth_fw.bin";
-		read-only;
 	};
 
 	bmfw: partition@45000 {
 		label = "bmfw";
 		reg = <0x45000 DT_SIZE_K(52)>;
+		padto = <DT_SIZE_K(52)>;
 		binary-path = "$BUILD_DIR/dmc/zephyr/zephyr.signed.bin";
-		read-only;
 	};
 
 	flshinfo: partition@52000 {
@@ -88,13 +82,14 @@ partitions {
 		reg = <0x52000 DT_SIZE_K(4)>;
 		binary-path =
 			"$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/flash_info.bin";
-		read-only;
 	};
 
 	boardcfg: partition@53000 {
 		label = "boardcfg";
 		reg = <0x53000 DT_SIZE_K(4)>;
 		binary-path = "$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/read_only.bin";
+		source = "$END - 0x1000";
+		provisioning-only;
 		read-only;
 	};
 
@@ -108,7 +103,7 @@ partitions {
 	failover: failover_partition@10000000 {
 		label = "failover";
 		reg = <0x10000000 DT_SIZE_K(76)>;
+		offset = <0x10000000>;
 		binary-path = "$BUILD_DIR/recovery/zephyr/zephyr.bin";
-		read-only;
 	};
 };
--- a/boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
+++ b/boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
@@ -26,84 +26,78 @@ partitions {
 		binary-path = "$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/fw_table.bin";
 	};
 
-	cmfw: partition@10000000 {
+	ethfwcfg: partition@16000 {
+		label = "ethfwcfg";
+		reg = <0x16000 DT_SIZE_K(4)>;
+		binary-path = "$BLOBS_DIR/tt_blackhole_erisc_params.bin";
+	};
+
+	ethfw: partition@17000 {
+		label = "ethfw";
+		reg = <0x17000 DT_SIZE_K(36)>;
+		binary-path = "$BLOBS_DIR/tt_blackhole_erisc.bin";
+	};
+
+	memfwcfg: partition@20000 {
+		label = "memfwcfg";
+		reg = <0x20000 DT_SIZE_K(4)>;
+		binary-path = "$BLOBS_DIR/tt_blackhole_gddr_params_$PROD_NAME.bin";
+	};
+
+	memfw: partition@21000 {
+		label = "memfw";
+		reg = <0x21000 DT_SIZE_K(16)>;
+		binary-path = "$BLOBS_DIR/tt_blackhole_gddr_init.bin";
+	};
+
+	ethsdreg: partition@25000 {
+		label = "ethsdreg";
+		reg = <0x25000 DT_SIZE_K(4)>;
+		binary-path = "$BLOBS_DIR/tt_blackhole_serdes_eth_fwreg.bin";
+	};
+
+	ethsdfw: partition@26000 {
+		label = "ethsdfw";
+		reg = <0x26000 DT_SIZE_K(20)>;
+		binary-path = "$BLOBS_DIR/tt_blackhole_serdes_eth_fw.bin";
+	};
+
+	flshinfo: partition@2b000 {
+		label = "flshinfo";
+		reg = <0x2b000 DT_SIZE_K(4)>;
+		binary-path =
+			"$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/flash_info.bin";
+	};
+
+	cmfw: partition@2c000 {
 		label = "cmfw";
-		reg = <0x10000000 DT_SIZE_K(106)>;
+		reg = <0x2c000 DT_SIZE_K(512)>;
 		offset = <0x10000000>;
 		binary-path = "$BUILD_DIR/smc/zephyr/zephyr.bin";
 		executable;
 	};
 
-	ethfwcfg: partition@30000 {
-		label = "ethfwcfg";
-		reg = <0x30000 DT_SIZE_K(4)>;
-		binary-path = "$BLOBS_DIR/tt_blackhole_erisc_params.bin";
+	failover: failover_partition@ac000 {
+		label = "failover";
+		reg = <0xac000 DT_SIZE_K(512)>;
+		offset = <0x10000000>;
+		binary-path = "$BUILD_DIR/recovery/zephyr/zephyr.bin";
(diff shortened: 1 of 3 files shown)


boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -18,97 +18,86 @@ partitions {
...
-		reg = <0x10000000 DT_SIZE_K(76)>;
+		reg = <0xac000 DT_SIZE_K(512)>;
+		offset = <0x10000000>;
 		binary-path = "$BUILD_DIR/recovery/zephyr/zephyr.bin";
+	};
+
+	bmfw: partition@12c000 {
+		label = "bmfw";
+		reg = <0x12c000 DT_SIZE_K(512)>;
+		padto = <DT_SIZE_K(52)>;
```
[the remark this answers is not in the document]
alexapostolu: I thought padto was the final size an image should be, so the padto is the same or greater than the image size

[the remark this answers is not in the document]
alexapostolu: Ah okay that makes sense

dts/bindings/mtd/tenstorrent,tt-boot-fs.yaml
cfriedt, line 1: IMHO, it would be best to avoid code changes at all in this PR, so that we're mainly only talking about "what" partition sizes will be, and then the "how" of how we achieve those things would be in subsequent PRs.

That will allow us to focus more and execute in a more top-down approach.

scripts/tt_boot_fs.py
cfriedt, line 1: Same - let's avoid code changes in this PR, so that we focus on the "what" here, and then we can focus on the "how" later.

cfriedt (commented):
Some thoughts:

Please drop code changes from this PR and only focus on defining the partition layout. Let's focus exclusively on "what" we are defining here w.r.t. the partition table, and focus on the "how" in subsequent PRs. That was the intent of having a separate JIRA ticket.

We can add comments to clarify what existing partitions might be renamed to, or what we might need aliases for.

In order to have a consistent state between current (confirmed) and upgrade (test) images, we need to have slot0 and slot1 for basically everything though. E.g. in case an update is only for testing, we expect to boot stable FW on the next run.

We still need to define slot1_partitions for

* cmfw
* bmfw
* erisc fw / cfg
* ethsd fw / cfg
* mrisc fw / cfg

Additionally, tables are also effectively firmware. We already have origcfg, which is effectively slot0. Previous tables should be used after a test fw has been flashed.

The one exception is the read-only table.

MCUBoot, the read-only table, and other things we plan to never change should be moved to the lowest possible region of flash, which would simplify physically protecting the "boot sector" if we ever choose to do that.

I'm on the fence about RISC-V fw / cfg, but we might as well add slot0 / 1 for those as well.

Then, the remaining flash space should just be a filesystem.

The nice thing about this is that once it's done, it's done, and we never have to change it again.

The fwbundle file is compressed, so we won't see the effect of padding bytes there. It's going to come into play though when we try to flash firmware. We definitely need to support sparse flashing as well, as Daniel mentioned.

What is tricky is that cmfw is ram loaded while bmfw is XIP. bmfw updates are _always_ written to slot1 and loaded from slot0. It would be great if we could use the same convention for cmfw, but I don't know if that's possible. In any case, we still need 2 slots, so let's worry about the "how" later.

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -18,97 +18,86 @@ partitions {
...
-		read-only;
+		reg = <0x20000 DT_SIZE_K(4)>;
+		binary-path = "$BLOBS_DIR/tt_blackhole_gddr_params_$PROD_NAME.bin";
 	};
 
-	memfw: partition@3b000 {
+	memfw: partition@21000 {
 		label = "memfw";
-		reg = <0x3b000 DT_SIZE_K(16)>;
+		reg = <0x21000 DT_SIZE_K(16)>;
```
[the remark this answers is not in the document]
danieldegrasse: Not sure there- we can definitely increase these

danieldegrasse:
> Please drop code changes from this PR and only focus on defining the partition layout. Let's focus exclusively on "what" we are defining here w.r.t. the partition table, and focus on the "how" in subsequent PRs. That was the intent of having a separate JIRA ticket.

I can minimize code changes but I don't think I can eliminate them. Currently we have "partial" usage of the `reg` value within the tt_boot_fs code- for example the bmfw tag uses the size field to set the "padto" value.

Right now I have put the partition layout into one commit to help with this, but we could minimize code changes to just pulling `padto` from a devicetree property

> What is tricky is that cmfw is ram loaded while bmfw is XIP. bmfw updates are always written to slot1 and loaded from slot0. It would be great if we could use the same convention for cmfw, but I don't know if that's possible. In any case, we still need 2 slots, so let's worry about the "how" later.

Agree on the two slots- what I'd like to discuss is how we use them for CMFW. As I've mentioned previously, I think one approach we could use for CMFW would be the following:

```
slot0: mission mode firmware. Updated by tt-flash
slot1: recovery firmware. Does not use tt-boot-fs binaries, just enumerates over PCIe so tt-flash can recover card
```

If we use this approach, we don't need A/B versions of all firmware binaries. If we write slot0 and the firmware fails for any reason, mcuboot will fall back to slot1 and we can retry the flash.

Any thoughts on this approach? I personally like it because it minimizes complexity (IE we don't need any A/B logic in the application) but still remains stable

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -18,97 +18,86 @@ partitions {
...
+	ethfwcfg: partition@16000 {
 		label = "ethfwcfg";
-		reg = <0x30000 DT_SIZE_K(4)>;
+		reg = <0x16000 DT_SIZE_K(4)>;
 		binary-path = "$BLOBS_DIR/tt_blackhole_erisc_params.bin";
-		read-only;
 	};
 
-	ethfw: partition@31000 {
+	ethfw: partition@17000 {
```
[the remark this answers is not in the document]
bingliTT: sorry - let's actually push the allocation to 64K, so there's room in the future to grow the FW size if it needs to be increased again

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -14,101 +20,171 @@ partitions {
...
+		reg = <0x169000 DT_SIZE_K(4)>;
+		binary-path = "$BLOBS_DIR/mcuboot_magic_confirmed.bin";
+	};
+
+	/*
+	 * MCUBoot partition- will be loaded by bootrom.
+	 * tagged as fail_over_image in tt-boot-fs,
+	 * but we actually use it as the primary image.
+	 */
+	boot_fallback: partition@1ea000 {
```
danieldegrasse, line 170: Since we tag mcuboot as the CMFW image and failover image, we actually should be fine to continue using the tt-boot-fs in our firmware application. This is because even if we write an invalid tt-boot-fs, the failover image header is in a separate page and thus won't be touched- so the bootrom *should* fall back to the other copy of mcuboot.

We don't have to do this though- it may be safer to instead use a different copy of the tt-boot-fs within our application

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -14,101 +20,171 @@ partitions {
...
+	 * main image should be booted in test mode by mcuboot.
 	 */
-	failover: failover_partition@10000000 {
+	main_test: partition@149000 {
+		label = "maintail";
+		reg = <0x149000 DT_SIZE_K(4)>;
+		binary-path = "$BLOBS_DIR/mcuboot_magic_test.bin";
+	};
+
+	/*
```
cfriedt, line 127: Probably doesn't need to happen right now, but as discussed, we should try to move these bits closer the start of flash in order to potentially facilitate a write-protected area.

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -14,101 +20,171 @@ partitions {
...
 
-	bmfw: partition@4c000 {
-		label = "bmfw";
-		reg = <0x4c000 DT_SIZE_K(68)>;
-		binary-path = "$BUILD_DIR/dmc/zephyr/zephyr.signed.bin";
-		read-only;
+	/* Update firmware image for the DMC */
+	dmfw: partition@4a000 {
+		label = "dmfw";
+		reg = <0x4a000 DT_SIZE_K(512)>;
```
cfriedt, line 92: Not sure if MCUBoot would complain about this size, but I had it as 440 kiB here
https://github.com/tenstorrent/tt-zephyr-platforms/blob/de400d5fb85dd57c57f105700fe5d13a77d17214/boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi#L108

danieldegrasse: Not sure either. I'll make it one section (4KB) larger than the planned internal DMFW flash size (448 KB). So 452KB for this slot

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
danieldegrasse: Done

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
danieldegrasse: Done

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -14,101 +20,173 @@ partitions {
 	flash-device-size = <DT_SIZE_M(64)>;
 	flash-block-size = <DT_SIZE_K(4)>;
 
-	cmfwcfg: partition@14000 {
+	/*
+	 * All partitions below here should eventually be mapped to binaries.
+	 * These binaries will not change, and should provide a safe bootloader
+	 * path in the event of an invalid main image.
+	 */
+
```
cfriedt, line 28: Might as well add a "boot sector" partition at 0 as well, with size 80 kiB

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -14,101 +20,173 @@ partitions {
...
 	};
 
-	origcfg: partition@15000 {
+	origcfg: partition@d6000 {
 		label = "origcfg";
-		reg = <0x15000 DT_SIZE_K(4)>;
+		reg = <0xd6000 DT_SIZE_K(4)>;
 		binary-path = "$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/fw_table.bin";
 		read-only;
 	};
```
cfriedt, line 94: This could probably be moved closer to the beginning too, since it's something we may want to write-protect.

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -14,101 +20,173 @@ partitions {
...
+		reg = <0xb4000 DT_SIZE_K(128)>;
+		binary-path = "$BUILD_DIR/mcuboot/zephyr/zephyr.bin";
+	};
+
+	boardcfg: partition@d4000 {
+		label = "boardcfg";
+		reg = <0xd4000 DT_SIZE_K(4)>;
+		binary-path = "$BUILD_DIR/smc/generated_board_cfg/$BOARD_REV/read_only.bin";
+		read-only;
+	};
```
cfriedt, line 77: I think this should be moved toward the top too, since it's something we likely want to write protect.

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -14,101 +20,173 @@ partitions {
...
+		reg = <0x1fb000 DT_SIZE_K(4)>;
+		binary-path = "$BLOBS_DIR/mcuboot_magic_test.bin";
+	};
+
+	/* Leave some space for future fixed partitions, if needed. */
+
+	/* Free space for storage */
+	storage: partition@300000 {
+		label = "storage";
+		reg = <0x300000 DT_SIZE_M(61)>;
```
cfriedt, line 190:
```shell
echo "$((80 + 128 + 512 + 4 + 128 + 4 + 4 + 4 + 4 + 4 + 4 + 4 + 64 + 64 + 64 + 452 + 4 + 512 + 4))"
2044
```

64*1024 - 2044 = 63492, and 63492 / 1024 is 62, So I guess, technically, the storage partition could be 62 MiB.

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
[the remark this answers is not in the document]
cfriedt: We should be looking at this file mostly not in the context of tt-boot-fs. So padto probably won't be used.

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -14,101 +20,214 @@ partitions {
...
-		read-only;
+	/*
+	 * Image trailer for the DMC firmware update. Indicates to MCUBoot that
+	 * the image is valid and should be booted in test mode.
+	 */
+	dmfw_tail: partition@29d000 {
+		label = "dmfwtail";
+		reg = <0x29d000 DT_SIZE_K(4)>;
+		binary-path = "$BLOBS_DIR/mcuboot_magic_test.bin";
 	};
```
cfriedt, line 203: It's probably worth it to mention here that the trailer is normally part of the slot / mcuboot image, and that this trailer partition is not needed for regular firmware updates.

https://docs.mcuboot.com/design.html#image-trailer

From what I understand, this is only needed for the BL2 update process.

I personally lean on "self-documenting code", rather than heavily-commented code, because the latter can suffer from bit rot, which often ends up being a source of confusion.

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -14,101 +20,214 @@ partitions {
...
-	failover: failover_partition@10000000 {
-		label = "failover";
-		reg = <0x10000000 DT_SIZE_K(76)>;
-		binary-path = "$BUILD_DIR/recovery/zephyr/zephyr.bin";
-		read-only;
+	main_test: partition@31d000{
+		label = "maintail";
+		reg = <0x31d000 DT_SIZE_K(4)>;
+		binary-path = "$BLOBS_DIR/mcuboot_magic_test.bin";
+	};
```
cfriedt, line 224: I feel like the comments here are also misleading, since the trailer is actually part of the slot0 / slot1 partition.

https://docs.mcuboot.com/design.html#image-trailer

Just a word of caution about over-commenting.  Comments are not code, aren't compiled, and tend to suffer from bitrot. They can't really be seen as a source of truth, and can sometimes be inaccurate, which causes some confusion.

PR #523, opened 2025-08-18
drivers: sensor: add pvt driver

alexapostolu:
Add PVT driver based on Zephyr's sensor API.

Zephyr has a new sensor API [submit/decode API](https://docs.zephyrproject.org/latest/hardware/peripherals/sensor/read_and_decode.html#sensor-read-and-decode). It says it's experimental but it's predicted to be the standard in the future. Reading can be done asynchronously with real time IO.

The sensor API allows multiple channels and indices, so we have three channel for temperature, voltage and process, and the index for the devices per channel.

Screenshot 2025-08-18 at 11 33 42 AM

### DIFF
--- a/doc/release/release-notes-18.9.md
+++ b/doc/release/release-notes-18.9.md
@@ -32,6 +32,9 @@ Major enhancements with this release include:
 [comment]: <> (UL Telemetry)
 [comment]: <> (UL Debug / Developer Features)
 [comment]: <> (UL Drivers)
+
+* Added Tenstorrent Blackhole PVT driver based on Zephyr's sensor API.
+
 [comment]: <> (UL Libraries)
 
 [comment]: <> (H2 New Samples, if applicable)


tests/boards/tenstorrent/tt_blackhole/sensor/pvt/prj.conf
```diff
@@ -0,0 +1,7 @@
+CONFIG_ZTEST=y
+
+CONFIG_SENSOR=y
+CONFIG_RTIO=y
+CONFIG_SENSOR_ASYNC_API=y
+CONFIG_NEWLIB_LIBC=y
+CONFIG_NEWLIB_LIBC_FLOAT_PRINTF=y
```
cfriedt, line 7: This seems a little odd - why would we need Newlib specifically?

alexapostolu: That was for float-printing in the tests, but I'm still playing around with the code so I might not need that

alexapostolu:
Current test cases passing

Screenshot 2025-08-29 at 11 22 58 AM

cfriedt (commented):
I think `drivers/sensor/X` are organized by vendor rather than sensor type, so it would be good to rename `pvt` to `tenstorrent` and make related changes.

PR #524, opened 2025-08-18
ci: fixes for e2e-stress tests

danieldegrasse:
Fix a few issues affecting the reliability of the E2E stress test, notably:

> In some CI tests the DMC flash prior to running tt-flash will leave the
    card in a state where a PCIe rescan is required. Add a pcie rescan step to
    the tt-flash runner in order to improve CI reliability, and automate
    this rescan for developers.

> In 3e74f74 (app: smc: improve code reuse in E2E stress test,
    2025-08-15), code that waited for the DMC to set the PCIe init time
    after booting the SMC was mistakenly dropped. This has caused the E2E
    stress test to fail in the PCIe init phase.

> tt-kmd sets a 10000ms watchdog timer on the ARC after reset. In some
cases this overrides our watchdog timer. To work around this, delay an
additional 10 seconds if the SMC initially fails to reset, and log a
warning. Only fail the test if the SMC doesn't reset within 10 seconds.

See individual commits for specific details- this PR essentially amounts to a variety of reliability improvements needed to get the e2e-stress tests functional again.

~This PR also adds a temporary commit to run e2e tests on the PR itself, so we can verify things are passing~

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -122,15 +122,17 @@ jobs:
           # - board: p100
           #  runs-on:
           #    - p100-jtag
+          # disabled until #https://github.com/tenstorrent/tt-metal/issues/19837
+          # is resolved- P300 support for metal is currently WIP
+          # - board: p300a
+          #   runs-on:
+          #     - p300a-jtag
           - board: p100a
             runs-on:
               - p100a-jtag
           - board: p150a
             runs-on:
               - p150a-jtag
-          - board: p300a
-            runs-on:
-              - p300a-jtag
     runs-on: ${{ matrix.config.runs-on }}
     container:
       image: ghcr.io/tenstorrent/tt-metal/upstream-tests-bh:v0.61.0-30-g3a3d9d704e
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -70,20 +70,8 @@ jobs:
       - name: run-e2e-stress-test
         working-directory: zephyr
         run: |
-          # TODO: ideally we would use one twister command to build and
-          # flash DMC and SMC firmware, but since each chip uses a separate
-          # debug adapter this doesn't work. For now, just flash DMC
-          # then run twister with SMC firmware
-          ./scripts/twister -i \
-            --tag e2e \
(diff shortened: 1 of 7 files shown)


danieldegrasse:
run with long tests passing: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/17130516774/job/48593757955?pr=524 (prior to removing commit that makes them run per-pr)

.github/workflows/hardware-long.yml
```diff
@@ -70,26 +70,15 @@ jobs:
...
-          # debug adapter this doesn't work. For now, just flash DMC
-          # then run twister with SMC firmware
-          ./scripts/twister -i \
-            --tag e2e \
-            -p $DMC_BOARD --device-testing \
-            --device-serial-pty ../tt-zephyr-platforms/scripts/dmc_rtt.py \
-            --flash-before \
-            --west-flash \
-            -T ../tt-zephyr-platforms/app \
-            --outdir twister-dmc-e2e
```
alexapostolu, line 84: Does the DMC not need to be flashed before the SMC anymore?

scripts/tt_flash_runner.py
```diff
@@ -4,6 +4,14 @@
...
 from runners.core import RunnerCaps, ZephyrBinaryRunner
 from pathlib import Path
+import sys
+from pcie_utils import rescan_pcie
+
+try:
+    import pyluwen
+except ImportError:
+    print("Error, required modules missing. Please run 'pip install pyluwen'")
+    sys.exit(1)
```
alexapostolu, line 14: Won't this throw the same error even without the try except block?

danieldegrasse: It will, yeah- west runners are expected to gracefully handle these errors and print an error message, see here: https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts/west_commands/runners/__init__.py#L15

danieldegrasse:
Kicked another round of stress tests on this PR. If they pass twice I think the set of fixes can be considered valid 🤞

danieldegrasse:
I believe this is ready for review and if everything looks good ok to merge- should get the overnight HW tests passing again

PR #525, opened 2025-08-19
ci: include built tt-boot-fs binary in uploaded artifacts

danieldegrasse:
Upload tt-boot-fs within built artifacts from CI job


danieldegrasse:
We only needed the tt-boot-fs binary in order to support preflash during manufacturing, but we already have a script to build this. This PR now has a CI job to create preflash firmware

alexapostolu:
It looks like the post run job for `build-preflash` failed, but it doesn't seem to be impacting anything else so is it fine?

Screenshot 2025-08-19 232358

alexapostolu (approved):

danieldegrasse:
> The post run job for `build-preflash` had an error, but it just looks like some minor Rust caching stuff so it should be fine to ignore right?
> Screenshot 2025-08-19 232358

Yeah, I'm not too worried about that because it looks like it's coming from github actions- if the job starts failing to build what it needs I'll worry

PR #526, opened 2025-08-19
lib: bh_chip: add i2c recovery after failed strap configuration

ShLiTT:
add i2c recovery after failed strap configuration for gpio6

### DIFF
--- a/lib/tenstorrent/bh_chip/strapping.c
+++ b/lib/tenstorrent/bh_chip/strapping.c
@@ -9,9 +9,11 @@
 #include <tenstorrent/bh_chip.h>
 
 #include <zephyr/drivers/gpio.h>
+#include <zephyr/drivers/i2c.h>
 
 void bh_chip_set_straps(struct bh_chip *chip)
 {
+	int ret;
 	bharc_enable_i2cbus(&chip->config.arc);
 	const struct gpio_dt_spec straps[] = {
 		chip->config.strapping.gpio6,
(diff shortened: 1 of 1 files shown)


danieldegrasse (approved):

nxuTT:
I'm good with this being a workaround for now, if we want to drop things to Quanta very soon. The problem with this issue is I2C bus cancellation is not done correctly and it causes I2C to be in bad state when MCU needs to access GPIO expander.

I think there should be a better way of doing this, I don't really like this because we don't know if this is the first transfer done on the I2C bus, you're only preventing transfer to GPIO expanders to exceed, but all earlier transfers will fail and even if this is the first transfer, people can potentially switch things around too.

I think we should probably do a I2C recover immediately after the I2C bus cancellation, but we need to measure how long I2C recover takes and make sure it doesn't take too long to affect PCIe enumeration timing.

ShLiTT:
> I'm good with this being a workaround for now, if we want to drop things to Quanta very soon. The problem with this issue if I understand it correctly is I2C bus cancellation is not done correctly and it causes I2C to be in bad state when MCU needs to access GPIO expander.

Yesss
> 
> I think there should be a better way of doing this, I don't really like this because we don't know if this is the first transfer done on the I2C bus, you're only preventing transfer to GPIO expanders to exceed, but all earlier transfers will fail and even if this is the first transfer, people can potentially switch things around too.

Agree — we had a group discussion on improving I²C bus cancellation. The proposed solutions were either using separate I/O operations in different thread contexts or introducing a submission queue for SMBus transactions.
> 
> I think we should probably do a I2C recover immediately after the I2C bus cancellation, but we need to measure how long I2C recover takes and make sure it doesn't take too long to affect PCIe enumeration timing.

Agree, that’s also what @danieldegrasse suggested. I’ll look into timing the I²C recovery.

Thank you for the comments Nora!

PR #534, opened 2025-08-25
ci: workflows: hardware-long: update container, run p300 metal tests

danieldegrasse:
Update container to latest metal revision, and enable metal tests on
P300a now that UMD should be updated to support metal

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -122,11 +122,9 @@ jobs:
           # - board: p100
           #  runs-on:
           #    - p100-jtag
-          # disabled until #https://github.com/tenstorrent/tt-metal/issues/19837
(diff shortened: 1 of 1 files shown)


danieldegrasse:
This PR contains a WIP commit to run metal tests on a per PR basis, which will be dropped before merge

tt-rkim:
latest is best https://github.com/tenstorrent/tt-metal/pkgs/container/tt-metal%2Fupstream-tests-bh/511306417?tag=latest

because that's guaranteed to have passed (well, once at least) on our CI and uploaded as latest release

However, it's up to you if you want to always use `latest`, or check periodically what it is and use the specific tag for it.

By the way, we may want to start using a 2nd container test (so there will be 2, as opposed to the current count of 1)

PR #536, opened 2025-08-25
Switch to fixed partition

mshimozatoTT:
- drivers: bh_fwtable: remove dependency on tt_boot_fs
- lib: bh_arc: remove dependency on tt_boot_fs


mshimozatoTT:
> Just checking- assuming this is dependent on #522?

yes I believe so

PR #543, opened 2025-08-28
Enable hard FP in SMC FW

alewycky-tenstorrent:
Enable hard FP in SMC FW.
- Tell compiler we have SPFP with div/rem and fma
- Enable FPU context switching
- Prefer floats over doubles

ARC does not have dedicated FPU registers, FP data is stored in universal GPRs. The context switch only needs to store 2 additional FPU control words.

### DIFF
--- a/drivers/sensor/tenstorrent/pvt_tt_bh.c
+++ b/drivers/sensor/tenstorrent/pvt_tt_bh.c
@@ -294,7 +294,7 @@ static inline void pvt_tt_bh_clock_config(void)
 	pvt_cntl_clk_synth_reg_u clk_synt;
 
 	clk_synt.val = PVT_CNTL_CLK_SYNTH_REG_DEFAULT;
-	uint32_t synth = (apb_clk * 0.2 - 2) * 0.5;
+	uint32_t synth = (apb_clk * 0.2f - 2.f) * 0.5f;
 
 	clk_synt.f.clk_synth_lo = synth;
 	clk_synt.f.clk_synth_hi = synth;
--- a/drivers/sensor/tenstorrent/pvt_tt_bh_decoder.c
+++ b/drivers/sensor/tenstorrent/pvt_tt_bh_decoder.c
@@ -15,7 +15,7 @@ LOG_MODULE_DECLARE(pvt_tt_bh);
 
 float pvt_tt_bh_raw_to_temp(uint16_t raw)
 {
-	float eqbs = raw / 4096.0 - 0.5;
+	float eqbs = raw / 4096.0f - 0.5f;
 	/* TODO: slope and offset need to be replaced with fused values */
 	return 83.09f + 262.5f * eqbs;
 }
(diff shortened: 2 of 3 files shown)


app/smc/src/main.c
```diff
@@ -145,3 +147,25 @@ static int bh_arc_init_end(void)
...
 }
 SYS_INIT_APP(bh_arc_init_end);
+
+#ifdef CONFIG_FPU_SHARING
+
+static void enable_float_on_current(struct k_work *work)
+{
+	k_float_enable(k_current_get(), 0);
+}
+
```
cfriedt, line 157: Something like this in `main()` would probably work, or at least in an `app_init()` called from `main()`.
```suggestion
if (IS_ENABLED(CONFIG_FPU_SHARING)) {
	k_float_enable(k_current_get(), 0);
}
```

Generally, it's preferable to use `IS_ENABLED()` instead of `#ifdef` since the former ensures that code compiles even when an option is not enabled.

tests/lib/tenstorrent/bh_arc/src/fpu.c
cfriedt, line 1: Thanks for writing tests.

I would guess that there is a fair bit of test coverage for `k_float_enable()` in the [upstream testsuites](https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/kernel/fpu_sharing).

Similarly, I think upstream already has test coverage for `snprintf()` (at least Picolibc does). That's testing the libc and the compiler which should have sufficient coverage.

What about adding a test to ensure that SWQ tasks that exercise floating point produce the desired results?

app/smc/src/main.c
alewycky-tenstorrent: Okay.

tests/lib/tenstorrent/bh_arc/src/fpu.c
cfriedt, line 1: The bh_arc testsuite is built to run on `native_sim` so adding this is causing some CI errors.

https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/tests/lib/tenstorrent/bh_arc/testcase.yaml
https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/17480353201/job/49649468708?pr=543

An alternative is to simply run the upstream `fpu_sharing` testsuite in our CI (rather than adding an additional one in-tree).

https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/kernel/fpu_sharing/generic/testcase.yaml

A way to achieve that is to add something to `test-conf/tests/kernel/fpu_sharing/generic/testcase.yaml`, and annotate the `kernel.fpu_sharing.generic.arc` test configuration with `tags: smoke`.

soc/tenstorrent/tt_blackhole/CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ if(CONFIG_SOC_TT_BLACKHOLE_SMC)
   endif()
 
   zephyr_library()
-  zephyr_library_sources(soc.c)
+  zephyr_library_sources(soc.c fpu_init.c)
```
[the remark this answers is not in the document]
alewycky-tenstorrent: Thanks, I applied this. (Ultimately k_enable_float does nothing if there's no FPU sharing but that's way downstream.)

PR #545, opened 2025-08-29
scripts: recover-blackhole: add recovery bundle scripting

danieldegrasse:
Add scripting to generate and flash a recovery bundle. These bundles are intended to replace the blackhole recovery scripting that is currently deployed to runners as a more stable way to place a known firmware on the card


danieldegrasse:
The recovery image built from this PR can be used like so: `sudo docker run --device /dev/bus/usb --privileged --rm ghcr.io/danieldegrasse/tt-zephyr-platforms/recovery-image:v0.0.1 python3 /tt-zephyr-platforms/scripts/tooling/blackhole_recovery/recover-blackhole.py /recovery.tar.gz p300a --force`

The intention is to use this docker image for recovery on CI systems. If we simply point the CI machines at a tag like "release" for the docker image, we can push a new recovery docker image and all CI machines will transparently update to use it.

scripts/board_metadata.yaml
```diff
@@ -0,0 +1,66 @@
...
+    name: "P100A-1"
+    bootfs-name: "tt_boot_fs"
+    protobuf-name: "P100A"
+    pyocd-config: "pyocd_config_spi1.py"
+p150a:
+  - upi: 0x40
+    name: "P150A-1"
+    bootfs-name: "tt_boot_fs"
+    protobuf-name: "P150A"
+    pyocd-config: "pyocd_config.py"
```
cfriedt, line 18: I think there might be a typo here

scripts/board_metadata.yaml
```diff
@@ -0,0 +1,66 @@
...
+    name: "P300A-1"
+    bootfs-name: "tt_boot_fs_left"
+    protobuf-name: "P300A_L"
+    pyocd-config: "pyocd_config_spi_combo.py"
+p300b:
+  - upi: 0x44
+    name: "P300B-1"
+    bootfs-name: "tt_boot_fs_right"
+    protobuf-name: "P300B_R"
+    pyocd-config: "pyocd_config.py"
```
cfriedt, line 48: possibly also a typo?

cfriedt:
> sudo docker run --device /dev/bus/usb --privileged --rm ghcr.io/danieldegrasse/tt-zephyr-platforms/recovery-image:v0.0.1 python3 /tt-zephyr-platforms/scripts/tooling/blackhole_recovery/recover-blackhole.py /recovery.tar.gz p300a --force

Would be good to add a bash script I think to simplify this - e.g. `blackhole_recovery.sh`?

scripts/board_metadata.yaml
danieldegrasse: Fixed, thanks for catching this

danieldegrasse:
> > sudo docker run --device /dev/bus/usb --privileged --rm ghcr.io/danieldegrasse/tt-zephyr-platforms/recovery-image:v0.0.1 python3 /tt-zephyr-platforms/scripts/tooling/blackhole_recovery/recover-blackhole.py /recovery.tar.gz p300a --force
> 
> Would be good to add a bash script I think to simplify this - e.g. `blackhole_recovery.sh`?

Added a bash script. My expectation is that developers will either:
- build their own recovery bundle with `prepare-recovery-bundle.py` or,
- download a prebuilt recovery bundle (we can produce these with releases if we like)

That being said, no reason not to support the docker method as well. I've added a script- we will likely want to run this script on our CI runners as a recovery method in the future

cfriedt:
✅  Successfully recovered my `p100` system that was not responding to the previous blackhole recovery method.

It might be good to print some additional suggestions here rather than just raising an error.
```shell
Traceback (most recent call last):
  File "/tt-zephyr-platforms/scripts/tooling/blackhole_recovery/recover-blackhole.py", line 218, in <module>
    main()
  File "/tt-zephyr-platforms/scripts/tooling/blackhole_recovery/recover-blackhole.py", line 213, in main
    raise RuntimeError(f"ASIC {idx} did not enumerate after flash")
RuntimeError: ASIC 0 did not enumerate after flash
```

My card did come back after a power cycle.

cfriedt (approved):

PR #548, opened 2025-09-02
zephyr: drivers: sensor: pvt: updated tests

alexapostolu:
Previous PVT sensor driver only read from a single channel. Updated the implementation to allow reading from all channels (8 channels for TS and VM, and 16 for PD)


alexapostolu:
I'm not sure why the twister test suite are ultimately timing out even though all of the individual test cases pass?
Screenshot 2025-09-04 at 5 52 19 PM

alexapostolu:
The issue with the tests timing out seemed to be from too many logs. I removed some and it passes now.

But for the unit tests, I think the new problem now is that since I don't have `CONFIG_SENSOR=y` for unit tests, then it maybe doesn't recognize the helper functions in `pvt_tt_bh.h`?

lib/tenstorrent/bh_arc/pvt.c
```diff
@@ -19,594 +18,114 @@
...
-		pd_read_status = ReadTS(i, &tmon_data);
-		if (pd_read_status != ReadOk) {
-			/* Something went wrong! Return -FLT_MAX to indicate an error */
-			return -FLT_MAX;
-		}
-		ts_sum += DoutToTemp(tmon_data);
-	}
-	return ts_sum / NUM_TS;
-}
+RTIO_DEFINE(pvt_ctx, 16, 16);
```
cfriedt, line 50: Is there a way to get the number of channels (16) from `pd_iodev` programmatically?

lib/tenstorrent/bh_arc/pvt.c
```diff
@@ -19,594 +18,114 @@
...
-
-		/* wait until delay chain takes effect */
-		WaitUs(250);
-		selected_pd_delay_chain = delay_chain;
-	}
-}
+/*
+ * Used for storing raw temperature sensor data when reading.
+ */
+uint8_t buf[sizeof(struct sensor_value) * 8];
```
cfriedt, line 55: Is there a way to get the number of channels from `ts_iodev` programmatically?

alexapostolu: Ah yes there is, I missed that

lib/tenstorrent/bh_arc/pvt.c
```diff
@@ -19,594 +18,114 @@
...
-	uint32_t offset;
-
-	if (type == VM) {
-		offset = VM_OFFSET;
-	} else {
-		offset = TS_PD_OFFSET;
-	}
-	return id * offset + base_addr;
-}
+#ifdef CONFIG_DT_HAS_TENSTORRENT_BH_PVT_ENABLED
```
cfriedt, line 24: Same here - I wonder if it's possible to use like `__maybe_unused` or something like that instead of `#ifdef`s.

lib/tenstorrent/bh_arc/telemetry_internal.c
```diff
@@ -4,16 +4,27 @@
...
 #include "regulator.h"
 
+#include <zephyr/kernel.h>
+#include <zephyr/drivers/sensor.h>
+#include <zephyr/drivers/sensor/tenstorrent/pvt_tt_bh.h>
+
 static int64_t last_update_time;
 static TelemetryInternalData internal_data;
 
+#ifdef CONFIG_DT_HAS_TENSTORRENT_BH_PVT_ENABLED
```
cfriedt, line 18: Same - let's avoid `#ifdef`s if possible.

`__maybe_unused`, `COND_CODE_1()`, and `IS_ENABLED()` could probably be used.

PR #552, opened 2025-09-02
bh_arc: gddr: make local functions static to improve encapsulation

cfriedt:
Keep locally used functions static and private rather than public and in the global namespace.

This helps to improve encapsulation (i.e. keep similar code in the same compilation unit) which is required for creating drivers.


alexapostolu (dismissed):

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -164,6 +171,7 @@ int LoadMriscFwCfg(uint8_t gddr_inst, uint8_t *buf, size_t buf_size, size_t spi_
 	return rc;
 }
 
+/* FIXME: make this static */
```
cfriedt, line 174: I think I'd like to hold off on merging this until after #550 is merged.

cfriedt: Fixed!

alexapostolu (approved):

PR #553, opened 2025-09-03
fw_table: enable aiclk ppm for p300x

afongTT:
Set appropriate limits for p300x boards to enable AICLK PPM. This allows AICLK > 800 MHz.

Board power limits:
- p300a = 450 W
- p300b = 450 W
- p300c = 600 W

Move TDP, and slow TDC throttlers out of the way. Set Fast TDC to 320 A.


cfriedt:
This is pretty close and it's not breaking anything 🤷‍♂️

alexapostolu:
Should this be for 150 and 100 as well?

cfriedt:
> Should this be for p150 and p100 as well?

What specifically? The PR seems like it's mainly intended for p300

PR #554, opened 2025-09-03
lib: tenstorrent: bh_arc: lower scratchpad size

afongTT:
Change TT_BH_ARC_SCRATCHPAD_SIZE Kconfig to int instead of hex type, reducing the scratchpad size from 0x512 (1298) to 512.

This seems to have been causing eth FW loading issues on cold boot.


cfriedt:
Hex is useful because it's unconditionally made unsigned, AFAIR, but that can also be a build assert.

cfriedt (approved):

PR #560, opened 2025-09-05
Fan control fixes

alewycky-tenstorrent:
Fixes:
- Fix pwm_max6639 fan controller channel 2 programming channel 1 instead
- Fix overflow resulting in 100% fan speed programming 0% instead
- Separate DMC->SMC fan speed feedback from SMC->DMC fan speed control
- DMC combines fan speed settings in a consistent way
- DMC always provides fan speed feedback to SMCs for telemetry, not just when speed forced

Cleanups:
- pwm_max6639 fan controller: unify code for channels 1 & 2
- pam_max6639 fan controller: simplify PWM frequency calculation
- Move per-chip fan speed state into bh_chip_data.
- Identify chip sending fan data from the SMBus channel rather than a message flag
- Always keep fan speed timer running. Cross-chip state management was too intricate.
- Initial fan speed from Kconfig rather than hard-coded

### DIFF
--- a/app/dmc/src/main.c
+++ b/app/dmc/src/main.c
@@ -33,6 +33,8 @@
 
 #define RESET_UNIT_ARC_PC_CORE_0 0x80030C00
 
+#define INITIAL_FAN_SPEED 35
+
 LOG_MODULE_REGISTER(main, CONFIG_TT_APP_LOG_LEVEL);
 
 BUILD_ASSERT(FIXED_PARTITION_EXISTS(bmfw), "bmfw fixed-partition does not exist");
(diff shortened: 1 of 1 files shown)


app/dmc/src/main.c
```diff
@@ -431,8 +431,7 @@ int main(void)
 				}
 
 				if (DT_NODE_HAS_STATUS(DT_ALIAS(fan0), okay)) {
-					pwm_set_cycles(max6639_pwm_dev, 0, UINT32_MAX, UINT32_MAX,
-						       0);
+					pwm_set_cycles(max6639_pwm_dev, 0, UINT8_MAX, UINT8_MAX, 0);
```
cfriedt, line 448: This seems specific to the implementation / hardware rather than the API. Is it possible to change how the driver handles `UINT32_MAX` / `UINT32_MAX` rather than in the application?

ShLiTT (approved):
Thank you for cleaning up fan ctrl Andrew!

PR #575, opened 2025-09-10
drivers: sensor: tenstorrent: pvt: added pvt subdirectory

alexapostolu:
Add PVT subdirectory to match with upstream Zephyr sensor directory conventions.

This also fixes the warning,

  No SOURCES given to Zephyr library:
  ..__tt-zephyr-platforms__drivers__sensor__tenstorrent


drivers/sensor/tenstorrent/pvt/Kconfig
```diff
@@ -0,0 +1,14 @@
+#  Copyright (c) 2025 Tenstorrent AI ULC
+#  SPDX-License-Identifier: Apache-2.0
+
+if SENSOR
```
cfriedt, line 4: `drivers/sensor/Kconfig` already has an `if SENSOR` condition, so this conditional can be removed

alexapostolu:
Do all CMake files need a copyright notice? I don't see it in `check-copyright.py` so I wasn't sure

PR #580, opened 2025-09-11
scripts: tooling: blackhole_recovery: always pull latest image

danieldegrasse:
Pull latest image when running blackhole recovery, rather than using a versioned image. This matches the method we will use on CI runners for recovery (and means that CI runners can use this script directly)


cfriedt:
@danieldegrasse - is this one still needed?

PR #583, opened 2025-09-11
ci: build docker containers for ci and recovery

danieldegrasse:
Automate building docker container for ci and recovery


danieldegrasse:
> Well, it looks like this won't work either- we need permissions to publish the container images, which we lack ☹️

After some reading on github token documentation- it seems that jobs running in PRs from forks may not get the same permissions as the main repo: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#changing-the-permissions-in-a-forked-repository-1

So I think provided the docker builds are only failing at the upload step (like they are here): https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/18181233924/job/51757239341?pr=583

we are good to go.

cfriedt (approved):

PR #584, opened 2025-09-11
scripts: define left ASIC first for P300

danieldegrasse:
Recovery scripting flashes ASICs in the order they are described, and
the left bootfs is the only one that contains the DMFW for P300 systems.
Define this asic first in metadata, so the board recovery scripting will
flash the DMFW during the initial recovery phase.


cfriedt (approved):
LGTM - feel free to merge if you don't need any other reviews

PR #589, opened 2025-09-12
app: pytest: e2e_smoke: rework harness setup

danieldegrasse:
Rework the harness setup for the e2e smoke test. This is required because with tt-kmd versions >2.4.0, applications using a "stale FD" referring to the tenstorrent device before a reset will be unable to access it.

We therefore need to modify our harness setup to never hold a reference to the chip detected by pyluwen across a reset, otherwise python won't induce closure of the underlying file descriptors.

### DIFF
--- a/.github/workflows/hardware-smoke.yml
+++ b/.github/workflows/hardware-smoke.yml
@@ -246,6 +246,12 @@ jobs:
             echo "ASIC_ID=0" >> "$GITHUB_ENV"
           fi
 
+      - name: build-tt-smc_console
+        working-directory: tt-zephyr-platforms/scripts/tooling
+        run: |
+          # Build tt-console
+          make
+
       - name: run-e2e-tests
         working-directory: zephyr
         run: |
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -37,6 +37,7 @@ BOOT_STATUS = 0x80030408
 PCIE_INIT_CPL_TIME_REG_ADDR = 0x80030438
 CMFW_START_TIME_REG_ADDR = 0x8003043C
 ARC_START_TIME_REG_ADDR = 0x80030440
+ARC_HANG_PC_REG_ADDR = 0x80030454
 
 # ARC messages
 ARC_MSG_TYPE_REINIT_TENSIX = 0x20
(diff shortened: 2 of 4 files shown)


danieldegrasse:
This PR is now marked as draft- I am running 10 iterations of the smoke e2e test on each card within this PR. If those pass, I'd consider this change reliable enough to merge

danieldegrasse:
The tests that continue to fail here are all related to SPI writes- I think that these errors are the same root cause as what I'm working on fixing on #591, so I think this PR should just depend on that. No reason to merge this without full confidence in the CI results.

app/smc/pytest/e2e_smoke.py
```diff
@@ -281,18 +333,19 @@ def arc_watchdog_test(arc_chip):
     os.environ.get("CONSOLE_DEV") == "/dev/tenstorrent/1",
```
jgrowdenTT, line 333: Any chance this fixes the P300a?

danieldegrasse: It passed once, but it was intermittent, right? I'll make a push running the P300 multiple times

jgrowdenTT: I have a sample from today that failed first try:
https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/18167187230/job/51712427081?pr=676#logs

Before I removed the tests, it certainly felt like it failed basically every time. I unfortunately didn't document this.

This PR just went 6/6 right? I think that's a clear indication that this got a whole lot healthier, and I think we should merge it. Tell the team to be on the lookout for failures, just in case. 

It's your call though; the other option is you merge without enabling the p300 tests and then we push a million p300a runs through in a new PR. 

Thanks for humouring me here, and great job on this!

.github/workflows/hardware-smoke.yml
```diff
@@ -183,16 +183,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        iteration: [1, 2, 3, 4, 5]
```
jgrowdenTT, line 186: Clean this one up before merge

jgrowdenTT (approved):

danieldegrasse:
@cfriedt or @afongTT could you take a look? I'd like a second set of eyes on this, don't want to merge it myself

PR #591, opened 2025-09-13
boards: tenstorrent: limit flash programming size

danieldegrasse:
Flash performance test fails intermittently when the programming frequency is set to 15 MHz, complaining that read data does not match what is programmed. Testing locally the mismatch seems to be that large sections of the flash are 0xff when they should be programmed with sequential bytes, likely indicating the TX FIFO underflowed and CS was deasserted.

To resolve this, limit the flash programming size to never exceed the depth of the TX FIFO. This has performance implications (Flash performance test previously took 383 ms for program, 4s for read and now takes 455 for program and 47 ms for read)


danieldegrasse:
For more details on failures, see runs like these:

https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/17656519915/job/50180489342?pr=581 

https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/17655677993/job/50180829826

cfriedt (approved):
I haven't looked at the schematic in the last hour or so, but are there level-shifters or switches in the signal path?

It's kind of concerning how much bandwidth we've left on the table (2 MHz isn't bad, but I know that a nucleo board is capable of writing at 32 MHz without issue).

Something for our PCB team to look into for sure, and I've already mentioned this to them.

cfriedt:
Oh sorry - just noticed this was still in draft

danieldegrasse:
> Oh sorry - just noticed this was still in draft

Yup, I will remove this from draft once it is reliably passing tests. Want to see if there's a way to automatically repeat them within a job

danieldegrasse:
> I haven't looked at the schematic in the last hour or so, but are there level-shifters or switches in the signal path?
> 
> It's kind of concerning how much bandwidth we've left on the table (2 MHz isn't bad, but I know that a nucleo board is capable of writing at 32 MHz without issue).
> 
> Something for our PCB team to look into for sure, and I've already mentioned this to them.

Just to be clear, the issue here has (AFAIK) nothing to do with board bandwidth. The issue is that the TX FIFO is vanishingly small on the SPI DW peripheral (15 bytes), so it can underrun easily. When it underrruns, the SPI CS is deasserted, so programming operations fail.

A proper (IMO) fix for this would be to just use a GPIO pin for CS on blackhole- then we wouldn't have to worry about the buggy CS implementation and could probably bump SPI programming speed quite a bit.

Another option we could explore (which would be a lot of effort) would be booting a secondary ARC core and using it exclusively as a DMA engine to manage the SPI. This would also probably let us bump frequency, not sure how much though.

danieldegrasse:
> > A proper (IMO) fix for this would be to just use a GPIO pin for CS on blackhole- then we wouldn't have to worry about the buggy CS implementation and could probably bump SPI programming speed quite a bit.
> 
> That might be a possibility - do we know if the CS pin can be configured as a GPIO?

I'm pretty sure it cannot be- the pinmuxing for BH essentially causes the peripheral to take over the relevant pins once it is enabled. If we could use the CS pin as GPIO this would fix the entire issue.

cfriedt:
I would also say performance is a major concern. If the ISR solution works, that would be sweet

danieldegrasse:
I did some tracing using this diff:

```diff
--- a/lib/tenstorrent/bh_arc/spi_eeprom.c
+++ b/lib/tenstorrent/bh_arc/spi_eeprom.c
@@ -65,7 +65,6 @@ static int SpiSmartWrite(uint32_t address, const uint8_t *data, uint32_t num_byt
        int rc;

        __ASSERT(sector_size > sizeof(spi_page_buf), "Sector size is larger than temp buffer");
-       sys_trace_named_event("spiwrite", address, num_bytes);

        /* Phase 1: Write first chunk (may be unaligned)*/
        chunk_size = MIN((ROUND_UP((address + 1), sector_size) - address), num_bytes);
@@ -168,7 +167,10 @@ static uint8_t read_eeprom_handler(uint32_t msg_code, const struct request *requ
                return 1;
        }

-       return SpiBlockRead(spi_address, num_bytes, csm_addr);
+       sys_trace_named_event("spiread_start", spi_address, num_bytes);
+       int rc = SpiBlockRead(spi_address, num_bytes, csm_addr);
+       sys_trace_named_event("spiread_end", spi_address, num_bytes);
+       return rc;
 }


@@ -194,7 +196,10 @@ static uint8_t write_eeprom_handler(uint32_t msg_code, const struct request *req
                return 1;
        }

-       return SpiSmartWrite(spi_address, csm_addr, num_bytes);
+       sys_trace_named_event("spiwrite_start", spi_address, num_bytes);
+       int rc = SpiSmartWrite(spi_address, csm_addr, num_bytes);
+       sys_trace_named_event("spiwrite_end", spi_address, num_bytes);
+       return rc;
 }

 REGISTER_MESSAGE(MSG_TYPE_READ_EEPROM, read_eeprom_handler);
```

The results indicate that our programming time take ~30 ms at the longest, for a 4096 byte transfer (some transfers take longer, but that is likely caused by the SPI flash write-in-progress bit taking time to clear)

The raw data (including the table of programming events) is attached. I generated this table using the following SQL query:

```sql
WITH filtered AS (
  SELECT name AS operation_name,ts-LAG(ts) OVER (ORDER BY ts) AS time_in_ns,
    EXTRACT_ARG(arg_set_id, 'args.arg1') AS operation_size, EXTRACT_ARG(arg_set_id, 'args.arg0') AS address
  FROM slices WHERE name GLOB 'spi*'
)
SELECT *,operation_size * 1000000 /time_in_ns AS bytes_per_ms FROM filtered WHERE operation_name GLOB 'spi*end' ORDER BY time_in_ns DESC
```

[Archive.zip](https://github.com/user-attachments/files/22413481/Archive.zip)

danieldegrasse:
Closing in favor of #606, which makes the SPI interrupt higher priority and has much better performance characteristics

PR #592, opened 2025-09-14
doc: getting_started: add getting started guide

alexapostolu:
Add Getting Started Guide based on Zephyr repository. Screenshot below from local host.

```sh
# remote
sphinx-build -b html doc _build
cd _build; python3 -m http.server 8000

# local
ssh -L 8000:localhost:8000 user@remote

# and now it's at http://localhost:8000/getting-started.html
```

Screenshot 2025-09-14 at 7 36 47 PM


doc/getting-started.rst
```diff
@@ -0,0 +1,236 @@
...
+Prerequisites
+-------------
+
+If you're new to Zephyr, we recommend starting with the official
+`Zephyr Getting Started Guide <https://docs.zephyrproject.org/latest/develop/getting_started/index.html>`_
+to familiarize yourself with the framework.
+
+Before continuing, ensure you have installed the
+`Zephyr system requirements <https://docs.zephyrproject.org/latest/develop/getting_started/index.html#install-dependencies>`_
+for your operating system.
```
alexapostolu, line 15: These instructions were in the README and I moved them over, but would it be clearer if we remove these links and just have all the setup in this doc?

cfriedt:
> @cfriedt Do you want the docs to be exactly matching with the README? Cause I ran it through Claude and it made some grammar corrections and stuff, is that fine?

Seems like a good idea to fix grammatical errors.

doc/getting-started.rst
cfriedt, line 1: To kind of keep a similar template as upstream docs, this should be renamed to `develop/getting_started/index.rst`.

PR #596, opened 2025-09-16
Fine-grained DMC events

alewycky-tenstorrent:
Create a distinct event for each activity in the DMC main loop. Slow activities check the signalled event bitmap. Some activities still have flags in the chip object to know when they've been triggered.

This is in preparation to move ia228_power_update to 1ms frequency. Fan update and CM2DM will stay at 20ms.

### DIFF
--- a/app/dmc/src/main.c
+++ b/app/dmc/src/main.c
@@ -571,6 +571,14 @@ static void send_logs_to_smc(void)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


cfriedt:
I like the general idea here - I just wish it would pass CI.

cfriedt:
@alewycky-tenstorrent - this needs a rebase

PR #598, opened 2025-09-17
boards: tenstorrent: reset DMC prior to flashing SMC

danieldegrasse:
Before running JTAG scan chain on SMC, reset the DMC if an openocd config is provided. This can help kick the SMC out of a bad state, since resetting the DMC will reapply the bootrom workaround.


danieldegrasse:
Closing this, we can reset the DMC in the recovery step for twister tests and it will accomplish the same reliability improvement

PR #603, opened 2025-09-17
improvements for test error reporting

danieldegrasse:
Add two crash handlers:
- a custom version of k_sys_fatal_error_handler
- a custom reset handler

Both crash handlers will log the system BLINK (crash address) into ICCM
at address 0x0000_0004. The reset handler will execute if the ARC
inadvertently triggers a soft reset, and the fatal error handler will
execute for cases that Zephyr considers a fatal exception.

Additionally, add a script to dump SMC state after a crash. This script will be run after a failure in pytest or CI as a whole, and can help us debug a crash on the ARC core.

Finally, exit pytest if the DUT fails to flash or enumerate. This will reduce log verbosity, since if the ARC fails to enumerate there is no point running the pytests and getting a bunch of errors.

Add CI steps to log the machine state, so that we can more easily reproduce errors in CI on local systems.

### DIFF
--- a/soc/tenstorrent/tt_blackhole/soc.c
+++ b/soc/tenstorrent/tt_blackhole/soc.c
@@ -6,6 +6,11 @@
 
 #include <zephyr/devicetree.h>
 #include <stdint.h>
+#include <zephyr/logging/log_ctrl.h>
+#include <zephyr/logging/log.h>
+#include <kernel_arch_func.h>
+
+LOG_MODULE_REGISTER(soc, CONFIG_LOG_DEFAULT_LEVEL);
 
 static inline void delay_spin(uint32_t count)
 {
@@ -16,8 +21,60 @@ static inline void delay_spin(uint32_t count)
 	}
 }
 
+#define PANIC_REASON_MAGIC 0xBADC0DE0
+#define REASON_K_PANIC     0x1
+#define REASON_ARC_RESET   0x2
+
+#define ARC_RESET_ADDR 0x80000000
+
+/*
+ * After panic the arc can't write to scratch registers. Write to ICCM
+ * to indicate we have panicked.
+ */
+#define ARC_PANIC(reason, blink)                                                                   \
+	do {                                                                                       \
+		sys_write32(PANIC_REASON_MAGIC | (reason), 0x0);                                   \
+		sys_write32(blink, 0x4);                                                           \
+	} while (0)
+
+/*
+ * Override the default kernel panic handler. We want to also
+ * dump the BLINK register value.
+ */
+__weak void k_sys_fatal_error_handler(unsigned int reason, const struct arch_esf *esf)
+{
+	ARG_UNUSED(esf);
+
+	ARC_PANIC(REASON_K_PANIC, esf->blink);
+
+	LOG_PANIC();
+	LOG_ERR("Halting system");
+	k_fatal_halt(reason);
+	CODE_UNREACHABLE;
+}
+
+/* This function is a custom hook we run during panic */
+static void arc_panic(void)
+{
+	uint32_t blink;
+
+	/* Get the BLINK register */
+	__asm__ volatile("mov %0, blink" : "=r"(blink));
+
+	ARC_PANIC(REASON_K_PANIC, blink);
+	while (1) {
+		/* Spin */
+	}
+}
+
 void soc_early_init_hook(void)
 {
+	/* Set the reset vector to arc_panic. */
+	sys_write32((uint32_t)&arc_panic, ARC_RESET_ADDR);
+	/* Clear ICCM registers */
+	sys_write32(0, 0x0);
+	sys_write32(0, 0x4);
+
 	if (IS_ENABLED(CONFIG_I2C)) {
 		uint32_t reg;
 		/* Manually toggle I2C reset control bits */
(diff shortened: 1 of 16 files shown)


danieldegrasse:
This PR has grown to quite a few commits- if desired, I can squash them down to a smaller subset

scripts/tooling/version.c
cfriedt, line 1: Sorta thinking `modinfo` could help here

```shell
$ modinfo tenstorrent
filename:       /lib/modules/6.8.0-40-generic/updates/dkms/tenstorrent.ko.zst
version:        2.4.0
description:    Tenstorrent AI kernel driver
license:        GPL
srcversion:     A5C46AEF1632036E5D19BF0
alias:          pci:v00001E52d0000B140sv*sd*bc*sc*i*
alias:          pci:v00001E52d0000401Esv*sd*bc*sc*i*
alias:          pci:v00001E52d0000FACAsv*sd*bc*sc*i*
depends:        
retpoline:      Y
name:           tenstorrent
vermagic:       6.8.0-40-generic SMP preempt mod_unload modversions 
sig_id:         PKCS#7
signer:         ubuntu Secure Boot Module Signature key
sig_key:        21:A2:4F:03:F5:B2:18:6B:52:46:38:2E:EB:61:6E:F3:2B:05:EC:44
sig_hashalgo:   sha512
signature:      31:9B:A8:83:8A:0E:48:A9:66:9C:A2:1F:BD:73:A1:AD:37:69:93:03:
                21:C3:DC:88:CE:A1:15:F9:F0:04:2D:B7:E8:97:A9:21:AF:B0:CF:06:
                63:3E:82:E8:5C:B7:0C:95:DE:E6:F5:59:3A:4C:83:AB:27:0A:B2:FC:
                27:86:58:9F:CB:42:71:4D:EE:37:3F:A4:36:7D:9D:72:8E:64:B9:05:
                4B:7B:7A:84:66:1E:F5:73:C4:75:5E:93:18:91:F6:4E:F6:46:39:0E:
                34:B7:81:49:3F:21:F6:C2:92:42:17:F4:BE:0E:36:47:C3:C5:E2:FA:
                AF:95:BB:2D:D0:74:41:87:F2:DE:3B:57:EB:A0:30:C1:67:4F:02:49:
                8E:61:45:8D:1B:27:B9:91:97:9D:81:F2:0D:FF:8C:2C:E4:90:F5:CA:
                8F:27:BA:32:F5:5A:49:9D:C3:13:88:87:E0:78:01:AD:B7:C1:C3:A4:
                36:C1:05:FE:21:51:11:A1:D3:52:B5:13:69:C4:7E:B7:02:7F:4B:1A:
                BD:1B:6F:81:9A:C1:BC:60:E7:8A:29:DD:00:50:03:D7:72:E2:97:6A:
                8D:E4:25:C3:B2:20:12:C1:EF:85:85:DD:9E:15:51:83:90:E2:11:58:
                42:33:FE:B8:18:FF:57:66:55:23:F8:91:7E:BA:BC:A1
parm:           max_devices:Maximum number of tenstorrent devices (chips) to support. (uint)
parm:           dma_address_bits:DMA address bits, 0 for automatic. (uint)
parm:           reset_limit:Maximum number of times to reset device during boot. (uint)
parm:           auto_reset_timeout:Timeout duration in seconds for M3 auto reset to occur. (byte)
```

soc/tenstorrent/tt_blackhole/soc.c
```diff
@@ -16,8 +21,60 @@ static inline void delay_spin(uint32_t count)
...
+
+/*
+ * Override the default kernel panic handler. We want to also
+ * dump the BLINK register value.
+ */
+__weak void k_sys_fatal_error_handler(unsigned int reason, const struct arch_esf *esf)
+{
+	ARG_UNUSED(esf);
+
+	ARC_PANIC(REASON_K_PANIC, esf->blink);
```
cfriedt, line 48: Upstream, they have the "hwinfo" driver API... but that's usually only useful after a reset and assumes that there is non-volatile memory (or that the SoC can properly reset without a power cycle, which isn't the case for us in this product family).

https://docs.zephyrproject.org/apidoc/latest/group__hwinfo__interface.html

What would also be kind of neat is to add a custom Coredump handler, which I think would allow us to add this hook in without a patch. SYS-1169 in the backlog.

IMHO, that still counts as a stability improvement (although it's kind of also a feature too)
https://docs.zephyrproject.org/latest/services/debugging/coredump.html

Could easily do it in the sprint after next as well.

danieldegrasse: Yeah, I don't think hwinfo will get us much here. This handler *probably* won't ever run in a crash, it only would run if the SOC hit the self-reset bit in the ARC MISC CNTL register.

Coredump looks interesting- I think we probably want to get this enabled with the SPI flash longer term. Probably want multiple backends though- if the system is in a bad enough state we really can't trust that SPI programming will work.

I'll also note this change is probably the lowest value (for debugging) in this PR- when the SMC faults, it rarely does so in a way that actually goes through the Zephyr crash handlers. Usually it just crashes in a way where the program counter reads `0x0` and the system is hung until reset.

scripts/tooling/version.c
cfriedt: I don't think we need to modify the container - we should just bind-mount /lib/modules into it when it's started (assuming that was the only thing modinfo complained about)

danieldegrasse: Ok- the launch command has been modified and I've added a step to install modinfo

danieldegrasse:
@cfriedt I'd like to put this in before the release- do you think that's reasonable?

PR #604, opened 2025-09-17
doc: add documentation for tracing

danieldegrasse:
Add documentation for using tracing support on blackhole.


doc/develop/debugging/tracing.rst
cfriedt, line 1: Looks awesome!

I'm not saying life is _always_ better with animated gifs (ok, maybe life _is always_ better with animated gifs) but we do have your `tracing.gif` demo in `doc/img/tracing.gif` already.

So if you felt so inclined to pull that image into the docs, I wouldn't be opposed to it :-)

cfriedt (approved):

PR #606, opened 2025-09-18
dts: arc: tenstorrent: make SPI interrupt highest priority

danieldegrasse:
Make the SPI interrupt highest priority. This will ensure that the SPI
controller is never preempted while writing to the SPI flash.

Should fix test issues like this one: [tenstorrent/tt-zephyr-platforms/actions/runs/17655677993/job/50180829826](https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/17655677993/job/50180829826)

### DIFF
--- a/dts/arc/tenstorrent/tt_blackhole_smc.dtsi
+++ b/dts/arc/tenstorrent/tt_blackhole_smc.dtsi
@@ -335,7 +335,7 @@
 		compatible = "snps,designware-intc";
 		reg = <0x800c0000 0x1000>;
 		interrupt-parent = <&intc>;
-		interrupts = <33 0>;
+		interrupts = <33 1>;
 		interrupt-controller;
 		#interrupt-cells = <3>;
 		num-irqs = <34>;
@@ -345,7 +345,7 @@
 		compatible = "snps,designware-intc";
 		reg = <0x800d0000 0x1000>;
 		interrupt-parent = <&intc>;
-		interrupts = <34 0>;
+		interrupts = <34 1>;
 		interrupt-controller;
 		#interrupt-cells = <3>;
 		num-irqs = <22>;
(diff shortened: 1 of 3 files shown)


cfriedt:
@danieldegrasse - As much as I hate adding yet another patch 😕, it could also be useful to convert the mspi_dw ISR to a direct one instead of a wrapped one.

https://github.com/zephyrproject-rtos/zephyr/blob/ccaaaabba2611b85f62c9dc6e7dbabd7542bce3b/drivers/mspi/mspi_dw.c#L1473

https://docs.zephyrproject.org/apidoc/latest/group__isr__apis.html#gac6c8746ac28da6ce02b24714f4144ff3

danieldegrasse:
> @danieldegrasse - As much as I hate adding yet another patch 😕, it could also be useful to convert the mspi_dw ISR to a direct one instead of a wrapped one.
> 
> [zephyrproject-rtos/zephyr@`ccaaaab`/drivers/mspi/mspi_dw.c#L1473](https://github.com/zephyrproject-rtos/zephyr/blob/ccaaaabba2611b85f62c9dc6e7dbabd7542bce3b/drivers/mspi/mspi_dw.c#L1473)
> 
> [docs.zephyrproject.org/apidoc/latest/group__isr__apis.html#gac6c8746ac28da6ce02b24714f4144ff3](https://docs.zephyrproject.org/apidoc/latest/group__isr__apis.html#gac6c8746ac28da6ce02b24714f4144ff3)

We unfortunately can't use a direct ISR here as I understand it- the MSPI ISR is a second level interrupt (routed through the DW interrupt controller) so I think we have to accept the standard penalty there, right?

danieldegrasse:
Test run across 10 iterations here: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/17835587474/job/50711823374?pr=606. We see some P300a failures but I believe they are unrelated

cfriedt (approved):

cfriedt:
@danieldegrasse - is this one just about ready?

PR #609, opened 2025-09-18
doc: fix rendering of zephyr-app-commands

danieldegrasse:
Fix rendering of Zephyr app commands, and remove directives we aren't rendering at all

### DIFF
--- a/boards/tenstorrent/tt_blackhole/doc/index.rst
+++ b/boards/tenstorrent/tt_blackhole/doc/index.rst
@@ -91,28 +91,34 @@ Supported Features
    Some components are in the process of being migrated to Zephyr's driver model, and are
    currently implemented in the ``lib/tenstorrent/bh_arc`` library.
 
-.. :external+zephyr:zephyr:board-supported-hw::
-
 Building
 ========
 
 SMC firmware can be built using the standard Zephyr build system. For other apps, tests,
 and samples, simply point the build system to the desired app directory.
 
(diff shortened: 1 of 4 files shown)


danieldegrasse:
So the doc build is failing now because we don't have the Zephyr module cloned in the doc build job (and we need it to use zephyr documentation directives)

@cfriedt personally I think we should just skip the Zephyr documentation directives. They are nice, but I don't think adding them justifies the additional doc build time. Any thoughts?

boards/tenstorrent/tt_blackhole/doc/index.rst
```diff
@@ -91,28 +91,34 @@ Supported Features
    Some components are in the process of being migrated to Zephyr's driver model, and are
    currently implemented in the ``lib/tenstorrent/bh_arc`` library.
 
-.. :external+zephyr:zephyr:board-supported-hw::
```
cfriedt, line 94: Does this necessarily need to be removed? IIRC, it did not cause docs to fail to build, but it wasn't rendering correctly for some reason.

jgrowdenTT:
I also have an immediate use case of using stuff in zephyr (specifically so that I can access their `templates/layout.html`) so would appreciate cloning the zephyr repo during the docs build.

I think, as our documentation improves, we'll have even more use cases where we want to lean on upstream's documentation tooling. I think the 1min is worth the cost of admission.

cfriedt (approved):

PR #619, opened 2025-09-22
wip: working on debugging SMC reset bug

danieldegrasse:
Can't get SMC reset bug to reproduce outside of CI, so run some tests on CI machines with logging.


danieldegrasse:
This CI run: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/17921914208/job/50958771841?pr=619 shows failures on P300a, but I believe those are unrelated. The tt-smi reset error seems to be gone

PR #633, opened 2025-09-23
ci: stop using github.ref

cfriedt:
The default `ref` for `actions/checkout` is

```
The branch, tag or SHA to checkout. When checking out the repository that
triggered a workflow, this defaults to the reference or SHA for that
event. Otherwise, uses the default branch.
```

We generally want our CI to use either the tag or SHA, so the default `ref` for `actions/checkout` is correct.

However, `github.ref` is incorrect, as described below.

```
The fully-formed ref of the branch or tag that triggered the workflow run.
For workflows triggered by push, this is the branch or tag ref that was
pushed. For workflows triggered by pull_request, this is the pull request
merge branch.
```

For push, we do not want the checkout to align with the branch, but the unique `sha`.

For pull requests, we do want the ref to be equal to `github.event.pull_request.number`, since subsequent force-pushes on the same branch should cancel previous jobs with the same PR number.

https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

Also, the tag `build-fw-` was copy-pasta'ed all over the place, so remove it where it is not required.

Also, for reusable workflows, `github.workflow` reflect the name of the top-level workflow being executed rather than any included reusable workflows. That means, `hardware-long` and `hardware-smoke` can also just use `github.workflow` without fear of conflicting with the `build-fw` workflow.

> The github.workflow variable provides context about the top-level workflow initiating the run.

### DIFF
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -25,7 +25,7 @@ on:
       - collab*
 
 concurrency:
-  group: ${{ github.workflow }}-build-fw-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
 # TODO: read boards / board revs from a YAML file (generate a matrix)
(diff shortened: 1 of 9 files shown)


danieldegrasse:
Looking at the documentation for github events: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request it seems that the `GITHUB_SHA` will be the "Last merge commit on the GITHUB_REF branch"- which is going to be the same as the `GITHUB_REF`: "PR merge branch refs/pull/PULL_REQUEST_NUMBER/merge"

So I think this is ok as is

PR #638, opened 2025-09-24
doc: Fix reference prefix

jgrowdenTT:
The current approach doesn't work if you click on the API link in a subpage. Attempt to use absolute pathing to make it work from everywhere.

However, the link will now not work when we look at doc build artifacts.


jgrowdenTT:
Talked with @danieldegrasse; We're not really sure the absolute pathing here is correct, but I'm pretty sure a direct https link will work. This will still be annoying from an artifact DL perspective (because it'll point to online Doxygen, instead of the doxygen link you downloaded from the build) but I'm pretty sure this will work once published.

jgrowdenTT:
Copilot says:

```
Yes, it should work the same! Here's why:

Current URL:
This is an absolute URL that explicitly specifies the full domain.

Your proposed change:
This is a root-relative URL that starts from the domain root.

Result:
If the site is hosted on https://docs.tenstorrent.com, both URLs will resolve to the exact same location:

Current: https://docs.tenstorrent.com/tt-zephyr-platforms/doxygen/index.html
Your change: https://docs.tenstorrent.com/tt-zephyr-platforms/doxygen/index.html
Benefits of the change:
More portable: Works regardless of the actual domain (useful for staging/dev environments)
Cleaner: Shorter and simpler
Protocol agnostic: Works with both HTTP and HTTPS

```

TIL. So, probably we should try to ship  `/tt-zephyr-platforms/doxygen/index.html` . Anyone have any thoughts before I go this way?

alexapostolu (approved):

PR #639, opened 2025-09-24
wip: testing github ref shas

danieldegrasse:
log both github sha refs


danieldegrasse:
Both SHAs look the same, so I think the `github.sha` change is ok

PR #641, opened 2025-09-24
ci: build-fw: fix concurrency group name

danieldegrasse:
build-fw needs a different concurrency group since other jobs depend on it. Otherwise github actions will error out with the message "a deadlock was detected for concurrency group"


cfriedt:
The documentation lies! 😭 
Oh well.

> The github.workflow variable provides context about the top-level workflow initiating the run.

PR #644, opened 2025-09-25
doc: Update doxygen

jgrowdenTT:
Update doxygen to the latest version; I want this because it makes our `@group` go to `Topics` vs `Modules`

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  DOXYGEN_VERSION: 1.9.6
+  DOXYGEN_VERSION: 1.14.0
 
 jobs:
   build:


.github/workflows/docs.yml
```diff
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  DOXYGEN_VERSION: 1.9.6
+  DOXYGEN_VERSION: 1.14.0
```
cfriedt, line 16: Would be good to verify that this won't conflict with anything in the greater Zephyr ecosystem.

Just bring prudent / cautious.

jgrowdenTT: It looks like they use 1.14.0. 
https://docs.zephyrproject.org/latest/doxygen/html/index.html

image

PR #648, opened 2025-09-26
zephyr: patches: use alt stm i2c reload from pr 89501

cfriedt:
Use https://github.com/zephyrproject-rtos/zephyr/pull/89501 to address i2c reload issues.

The timeout logic that was added (in https://github.com/zephyrproject-rtos/zephyr/pull/88631) for `i2c_stm32_msg_read()` and `i2c_stm32_msg_write()` looks sound to me, so the `smbus-cancel.patch` changes are relatively minimal.


danieldegrasse:
Just to make sure I follow, this PR keeps the I2C cancel patch but drops the I2C timeout patch and uses logic from the upstream driver, right?

cfriedt:
> Just to make sure I follow, this PR keeps the I2C cancel patch but drops the I2C timeout patch and uses logic from the upstream driver, right?

Yes, exactly - so it just kind of simplifies the cancellation patch.

alexapostolu (approved):

PR #650, opened 2025-09-26
lib: bh_arc: Rename msg_type to tt_arc_msg_type

jgrowdenTT:
Namespace the msg_type to be under tt to differentiate it from all the other msg types.

Based on PR feedback [here.](https://github.com/tenstorrent/tt-zephyr-platforms/pull/620#discussion_r2377385694)

### DIFF
--- a/include/tenstorrent/msgqueue.h
+++ b/include/tenstorrent/msgqueue.h
@@ -57,7 +57,7 @@ struct message_queue_header {
 
 /** @brief Host request to force the fan speed */
 struct force_fan_speed_rqst {
-	/** @brief The command code corresponding to @ref MSG_TYPE_FORCE_FAN_SPEED*/
+	/** @brief The command code corresponding to @ref TT_SMC_MSG_FORCE_FAN_SPEED*/
 	uint8_t command_code;
 
 	/** @brief Three bytes of padding */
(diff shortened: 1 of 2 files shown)


include/tenstorrent/tt_arc_msg_type.h
cfriedt, line 1: I think because the include path is `<tenstorrent/xxxx>`, we can probably get away with naming this header `<tenstorrent/arc_msg.h>`.

include/tenstorrent/tt_arc_msg_type.h
```diff
@@ -21,8 +21,8 @@ extern "C" {
  * @{
  */
 
-/** @brief Enumeration listing the available host requests IDs the SMC can process*/
-enum msg_type {
+/** @brief Enumeration listing the available host request IDs the SMC can process*/
+enum tt_arc_msg_type {
```
cfriedt, line 25: Maybe even `enum tt_arc_msg_e`. That's just splitting hairs though, so just a suggestion. The important thing is that the enum name is namespaced.

```suggestion
enum tt_arc_msg_type {
```

I guess technically, enum values could also be namespaced `TT_ARC_MSG_xxx`.

include/tenstorrent/tt_arc_msg_type.h
cfriedt: Technically, the message ids really have nothing to do with arc.

tenstorrent/tt_... is kind of redundant though.

cfriedt (dismissed):

danieldegrasse (dismissed):

PR #651, opened 2025-09-26
manifest: switch to zephyr v4.3.0-pre

cfriedt:
Update to Zephry v4.3.0-pre.

This change updates to a relatively stable version of Zephyr that is about 1 month before the feature freeze of v4.3.0-rc1.

Updating to this version allows us to drop 8 patches immediately.

As we approach the v4.3.0-rc1 release candidate (and feature freeze), we should aim to reduce the number of patches to zero (or a more manageable, smaller number).

Some notes:

- the new `LicenseAndCopyrightCheck` compliance checker produces way too many false positives for things like `.patch` files, `.gitignore`, `testcase.yml`, `.in`, and `.txt` files to be useful currently, so it's disabled
- our ci container would need to be automatically updated and deployed with each PR in order to use the `SDK_VERSION` file, but that might happen soon, so there is hope
- not 💯 sure why we need `dotfiles` and `jsonschema` installed manually in some cases
- it's very clear that there needs to be a better way to create an allow-list than `UNDEF_KCONFIG_ALLOWLIST` hard-coded in `check_compliance.py`

Components removed:
  - drivers/sensor/ti/ina228 (driver was added by someone else upstream)

Patches removed:
  - zephyr/k-event-wait-safe.patch
  - zephyr/pr-92709.patch
  - zephyr/tracing-ctf-take-IRQ-lock-before-generating-timestam.patch
  - zephyr/i2c-dw-target-correction.patch
  - zephyr/stm-i2c-reload.patch

Patches added:
  - zephyr/patches/zephyr/drivers-i2c-i2c-stm32-v2-stability-improvements.patch
  
Should be merged after
- [x] https://github.com/tenstorrent/tt-zephyr-platforms/pull/648


cfriedt:
Should be able to remove more patches in the next 4 weeks.

drivers/misc/bh_fwtable/bh_fwtable.c
alewycky-tenstorrent, line 1: Are these changes intended for this merge request?

alewycky-tenstorrent: Why are a log text change and minor code restructure needed when upgrading to this zephyr version?

As for the code change itself, consider unifying the code paths and just switching the array contents for CONFIG_TT_SMC_RECOVERY. The order that the tables are loaded doesn't matter.

alewycky-tenstorrent: I accept that the rest fixes the return value, but isn't it equally valid even before the zephyr update? If so commit it separately, the zephyr version bump is a much riskier change.

cfriedt: I was a little surprised as well, but the newer kernel/init.c result checks asserted on the return value being `1`! I guess before, it would assume success for a negative return value.

alexapostolu (approved):

alexapostolu:
I probably missed it somewhere, but why are we switching to v4.3.0-pre now instead of waiting for the rc1?

drivers/misc/bh_fwtable/bh_fwtable.c
cfriedt: Fixed!

cfriedt:
* pulled in main after https://github.com/zephyrproject-rtos/zephyr/pull/96509 was merged

cfriedt:
> I probably missed it somewhere, but why are we switching to v4.3.0-pre now instead of waiting for the rc1?

The issue with switching to rc1 is that we can't upstream new features anymore. It's after the feature freeze, so bugfixes only.

lib/tenstorrent/event/event.c
```diff
@@ -19,7 +19,7 @@ uint32_t tt_event_wait(uint32_t events, k_timeout_t timeout)
 {
 	uint32_t ret;
 
-	ret = k_event_wait_safe(&tt_event, events, timeout);
+	ret = k_event_wait_safe(&tt_event, false, events, timeout);
```
alexapostolu, line 22: Does the `bool` argument come after the `events` argument?
https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/kernel.h

cfriedt: Fixed!

alexapostolu (approved):

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -201,12 +201,21 @@ static int tt_bh_fwtable_load(const struct device *dev, enum bh_fwtable_e table)
...
+
+	rc = tt_bh_fwtable_load(dev, BH_FWTABLE_BOARDCFG);
 	if (IS_ENABLED(CONFIG_TT_SMC_RECOVERY)) {
-		return tt_bh_fwtable_load(dev, BH_FWTABLE_BOARDCFG);
-	} else {
-		return (tt_bh_fwtable_load(dev, BH_FWTABLE_FLSHINFO) ||
-			tt_bh_fwtable_load(dev, BH_FWTABLE_BOARDCFG) ||
-			tt_bh_fwtable_load(dev, BH_FWTABLE_CMFWCFG));
+		return rc;
+	}
```
cfriedt, line 209: I think I need an extra conditional following this still

cfriedt: Fixed!

alexapostolu (approved):

PR #652, opened 2025-09-26
drivers: sensors: ti: remove ina228 driver in-tree as we are using Zephyr sensor APIs

ShLiTT:
drivers: sensors: ti: remove ina228 driver in-tree as we are using Zephyr sensor APIs


alexapostolu:
Should you also remove the bindings file?
https://github.com/tenstorrent/tt-zephyr-platforms/blob/bb22a80ebe97a97102c72e1faca62cd3d2d1230c/dts/bindings/sensor/ti%2Cina228.yaml#L8

cfriedt:
Closing, since it was removed already.

PR #653, opened 2025-09-28
zephyr: patches: drop stm32-block-read.patch since it is upstream

cfriedt:
* drop stm32-block-read.patch since it landed upstream in https://github.com/zephyrproject-rtos/zephyr/pull/96516


cfriedt:
It looks like subsequent smbus or i2c patches don't apply easily, so I may need to do a minor rework of some of them.

PR #654, opened 2025-09-29
app: smc: pytest: modify verbosity and use failure rate

cfriedt:
Reduce the verbosity on some of the tests that execute e.g. 1000 times (although failures are still always logged).

Additionally, add a `failure_fail_count` as a threashold for group failure rate for repeated tests. For most tests, a single failure is still enough to fail the test but for some 10 / 1000 is probably a good threshold, since several tests fail when 1 or 2 of 1000 iterations fail.

10 / 1000 is a failure rate of 1%, so if we pass with a 0.1% or 0.2% failure rate (1 or 2/1000), it's probably still safe.

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -353,7 +353,6 @@ def smi_reset_test(asic_id):
     smi_reset_result = subprocess.run(
         smi_reset_cmd.split(), capture_output=True, check=False
     )
-    logger.info(f"'tt-smi -r' returncode:{smi_reset_result.returncode}")
     if smi_reset_result.returncode != 0:
         logger.warning(f"'tt-smi -r' failed: {smi_reset_result.stdout.decode()}")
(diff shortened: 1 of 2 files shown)


cfriedt:
> Not sure about the `failure_fail_count` for `test_smi_reset` and `test_dirty_reset`, otherwise lgtm!

Since those use 1000 iterations, `1000 // 100 = 10`, so if iterations fail >= 10 of 1000 times (i.e. has a 1% failure rate) the test will fail overall. `//` is the integer division operator in python.

cfriedt:
> Yes, I understand that—I'm just not sure if 10 is a reliable number for testing.

We don't have any guarantees about these tests, so they're really just whatever we decide they should be.

I did ask for 1000 iterations on these tests when they were added, but I also asked for the ability to decide pass / fail based on a failure rate (number of failures per 1000 iterations). That was never implemented, so this is just adding the thing that I originally requested.

PR #655, opened 2025-09-29
Revert "app: smc: pytest: modify verbosity and use failure rate"

danieldegrasse:
This reverts commit 46dfbc45791a8dee2ad4033226de46db6fdb6e09. In my opinion, we should not be setting thresholds for PCIe enumeration or reset failures- we should be targeting 100% pass rates for these tests.

There are a few reasons I hold this view:
- Strict pass criteria for these tests lets us catch intermittent errors early. If we allow a threshold for errors, then when we do start accumulating enough bugs to hit the threshold we may be chasing several different bugs at once
- Some of these tests (like PCIe enumeration) should *never* fail. Systems should enumerate every boot, full stop. Anything less is a firmware issue in my opinion

I'm open to changing my viewpoint here if others disagree, but I would like to see more discussion than occurred on #654 prior to merge.

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -353,6 +353,7 @@ def smi_reset_test(asic_id):
     smi_reset_result = subprocess.run(
         smi_reset_cmd.split(), capture_output=True, check=False
     )
+    logger.info(f"'tt-smi -r' returncode:{smi_reset_result.returncode}")
     if smi_reset_result.returncode != 0:
         logger.warning(f"'tt-smi -r' failed: {smi_reset_result.stdout.decode()}")
         smc_test_recovery.recover_smc(asic_id)
--- a/app/smc/pytest/e2e_stress.py
+++ b/app/smc/pytest/e2e_stress.py
@@ -66,25 +66,17 @@ def test_arc_watchdog(arc_chip):
     Validates that the DMC firmware watchdog for the ARC will correctly
     reset the chip
     """
-    # todo: find better way to get test name
-    test_name = "ARC watchdog test"
     total_tries = min(MAX_TEST_ITERATIONS, 100)
     fail_count = 0
-    failure_fail_count = 1
-
     for i in range(total_tries):
-        if i % 10 == 0:
-            logger.info(f"{test_name} iteration {i}/{total_tries}")
-
+        logger.info(f"Starting ARC watchdog test iteration {i}/{total_tries}")
         result = arc_watchdog_test(arc_chip)
         if not result:
-            logger.warning(f"{test_name} failed on iteration {i}")
+            logger.warning(f"ARC watchdog test failed on iteration {i}")
             fail_count += 1
 
-    report_results(test_name, fail_count, total_tries)
-    assert fail_count < failure_fail_count, (
-        f"{test_name} failed {fail_count}/{total_tries} times."
-    )
+    report_results("ARC watchdog test", fail_count, total_tries)
+    assert fail_count == 0, "ARC watchdog test failed a non-zero number of times."
 
 
 def test_pcie_fw_load_time(arc_chip):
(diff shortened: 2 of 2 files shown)


cfriedt (changes requested):
Uhmm... no?

danieldegrasse:
> Uhmm... no?

I would really prefer to discuss this more directly, perhaps in a meeting offline. I was in the process of reviewing #654 when you merged it, and I still do not agree we should be setting acceptable failure rates. I don't want to debug 10 different intermittent bugs with our long tests once we do hit the threshold point we set for these failures- I expect that will be much more difficult than narrowing down one issue

cfriedt:
> This reverts commit [46dfbc4](https://github.com/tenstorrent/tt-zephyr-platforms/commit/46dfbc45791a8dee2ad4033226de46db6fdb6e09). In my opinion, we should not be setting thresholds for PCIe enumeration or reset failures- we should be targeting 100% pass rates for these tests.

First, there have been no modifications for PCIe enumeration rates.

Given the following:

1. I asked for the SMC and DMC reset stress tests to have 1000 iterations
2. I asked for the SMC and DMC reset stress tests to have a configurable failure rate (because this is a sane approach for stress testing)
3. The latter was never implemented
4. We regularly need to re-run tests because there is an outlier that causes the entire run to fail

There is no harm in introducing an acceptible failure rate.

The "strict pass criteria" that I asked for when the 1000-iteration tests were initially added would have been e.g. 1%, which it is now.

While I am happy you are enthusiastic about testing, we have no way to systematically make any guarantees about these tests yet, so doing so is very premature.

cfriedt:
> > First, there have been no modifications for PCIe enumeration rates.
> 
> [46dfbc4#diff-09efc5b2f72c04359614dcae2161cadfb1a8c4c20be97be74436963e3ec74d50R99](https://github.com/tenstorrent/tt-zephyr-platforms/commit/46dfbc45791a8dee2ad4033226de46db6fdb6e09#diff-09efc5b2f72c04359614dcae2161cadfb1a8c4c20be97be74436963e3ec74d50R99) `test_pcie_fw_load_time` is validating that the PCIe enumeration would have completed by the spec deadline- it seems that the change added an acceptable failure rate of 10% for this test, am I missing something there?

Yes (to your last question). The assertion was changed from `fail_count == 0` to `fail_count < 1`, which is the same.

> The issue I am struggling with here is that I do believe the failures we see right now in stress tests (0.2% on tt-smi reset and 0.01% on DMFW ping) are probably the result of a legitimate issue.

Maybe. It's actually pretty hard to say (what) though (if anything) has caused that, whether it can be fixed in a reasonable amount of time, and whether it's even a problem.

> DMFW ping tests have been passing reliably for quite a while, and we just added multiple changes in tree that touch STM32 I2C code.

Well, passing... it's also passing now. What was the failure rate before the change? Was the previous failure rate important for a particular SLA?

> Obviously I haven't dove into this particular failure in depth, but I feel that we should spend time investigating it before we classify the error as acceptable.

Start by quantifying the error.

But also, realize that we have effectively taken an arbitrary number that we have chosen, and subtracted one (or two) from it.

The error rate itself is also outside of 3 standard deviations of what is (presumably) a normally distributed variable, which is generally statistically insignificant.

> My primary concern with #654 is that we are going to miss the introduction of new issues into tree, since we have very limited visibility into the pass rate of these tests outside of the CI job status. If we had a dashboard with "health metrics" for the firmware, I would be much less concerned about an acceptable pass rate

That's a great idea.

> > While I am happy you are enthusiastic about testing, we have no way to systematically make any guarantees about these tests yet, so doing so is very premature.
> 
> Can you expand on what you mean here? Maybe the path forwards is some sort of dashboard where we log failure rates

Same as above -> What was the failure rate before the change? Was the previous failure rate important for a particular SLA?

Also sure, a dashboard is a great idea. We might have a ticket for it already.

danieldegrasse:
Going to close this out, I think the path forwards is a dashboard with better test visibility. #660 is the start of that.

PR #657, opened 2025-09-29
drivers: dma: add dma_arc_hs_cluster driver

ShLiTT:


app/smc/prj.conf
```diff
@@ -58,3 +58,7 @@ CONFIG_RETENTION_BOOTLOADER_INFO_TYPE_MCUBOOT=y
 # Generate unsigned images by default
 # To sign FW, use SB_CONFIG_BOOT_SIGNATURE_KEY_FILE="/path/to/key.pem"
 CONFIG_MCUBOOT_GENERATE_UNSIGNED_IMAGE=y
+
+# DMA
+CONFIG_DMA=y
+CONFIG_DMA_ARC_HS=y
```
cfriedt: I think this still needs to be removed. We should definitely have 1 pr where we're adding the driver in a test, and then another PR where we enable the driver in `app/smc` and in devicetree.
```suggestion
```

boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc_p100a.yaml
```diff
@@ -13,6 +13,7 @@ toolchain:
 supported:
   - smp
   - gpio
+  - dma
```
cfriedt: These are OK - don't worry about them.

drivers/dma/Kconfig.arc_hs
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2025 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+config DMA_ARC_HS
+	bool "ARC DMA driver"
+	default y
+	depends on DT_HAS_SNPS_DESIGNWARE_DMA_ARC_HS_ENABLED && DMA
```
cfriedt: The `&& DMA` can be dropped because of the `if DMA` block in `drivers/dma/Kconfig`

```suggestion
	depends on DT_HAS_SNPS_DESIGNWARE_DMA_ARC_HS_ENABLED
```

drivers/dma/dma_arc_hs.c
```diff
@@ -0,0 +1,1011 @@
...
+
+/* ARC DMA Attribute Flags */
+#define ARC_DMA_NP_ATTR       (1 << 3) /* Enable non posted writes */
+#define ARC_DMA_SET_DONE_ATTR (1 << 0) /* Set done without triggering interrupt */
+
+/* ARC Auxiliary Register Access Functions */
+static inline void arc_write_aux(uint32_t reg, uint32_t val)
+{
+	__asm__ volatile("sr %0, [%1]" : : "r"(val), "r"(reg) : "memory");
+}
```
cfriedt: It would be good to check if Zephyr has any existing functions for doing the same thing.

Same with `arc_read_aux()`

drivers/dma/dma_arc_hs.c
```diff
@@ -0,0 +1,1011 @@
...
+static inline void dma_arc_hs_config_hw(void)
+{
+	uint32_t reg = 0;
+
+	reg = (0xf << 4);                   /* Set LBU read transaction limit to max */
+	reg |= (0x4 << 8);                  /* Set max burst length to 16 (max supported) */
+	arc_write_aux(DMA_S_CTRL_AUX, reg); /* Apply settings above */
+}
+
+static inline void dma_arc_hs_init_channel_hw(uint32_t dma_ch, uint32_t base, uint32_t last)
```
cfriedt: Can remove the `inline` modifier.
```suggestion
static void dma_arc_hs_init_channel_hw(void)
```

drivers/dma/dma_arc_hs.c
```diff
@@ -0,0 +1,1011 @@
...
+{
+	arc_write_aux(DMA_C_CHAN_AUX, dma_ch);
+	arc_write_aux(DMA_C_SRC_AUX, (uint32_t)p_src);
+	arc_write_aux(DMA_C_DST_AUX, (uint32_t)p_dst);
+	arc_write_aux(DMA_C_ATTR_AUX, attr);
+	arc_write_aux(DMA_C_LEN_AUX, len);
+}
+
+/* Queue a transfer on the currently selected channel (for multi-block) */
+static inline void dma_arc_hs_next_hw(const void *p_src, void *p_dst, uint32_t len, uint32_t attr)
```
cfriedt: Can remove the `inline` modifier.
```suggestion
static void dma_arc_hs_next_hw(const void *p_src, void *p_dst, uint32_t len, uint32_t attr)
```

drivers/dma/dma_arc_hs.c
```diff
@@ -0,0 +1,1011 @@
...
+	arc_write_aux(DMA_C_ATTR_AUX, attr);
+	arc_write_aux(DMA_C_LEN_AUX, len);
+}
+
+static inline uint32_t dma_arc_hs_get_handle_hw(void)
+{
+	return arc_read_aux(DMA_C_HANDLE_AUX);
+}
+
+static inline uint32_t dma_arc_hs_poll_busy_hw(void)
```
cfriedt: Can remove the `inline` modifier.
```suggestion
static uint32_t dma_arc_hs_poll_busy_hw(void)
```

drivers/dma/dma_arc_hs.c
```diff
@@ -0,0 +1,1011 @@
...
+{
+	return arc_read_aux(DMA_C_STAT_AUX);
+}
+
+static inline void dma_arc_hs_clear_done_hw(uint32_t handle)
```
cfriedt: Can remove the `inline` modifier.
```suggestion
static void dma_arc_hs_clear_done_hw(uint32_t handle)
```

drivers/dma/dma_arc_hs.c
```diff
@@ -0,0 +1,1011 @@
...
+	const struct device *dev;
+	bool work_initialized;
+};
+
+static int dma_arc_hs_config(const struct device *dev, uint32_t channel, struct dma_config *config)
+{
+	const struct arc_dma_config *dev_config = dev->config;
+	struct arc_dma_data *data = dev->data;
+	struct arc_dma_channel *chan;
+	k_spinlock_key_t key;
```
cfriedt:
```suggestion
```

drivers/dma/dma_arc_hs.c
```diff
@@ -0,0 +1,1011 @@
...
+	uint32_t block_idx = 0;
+	k_spinlock_key_t key, hw_key;
+	uint32_t current_channel = channel;
+
+	if (channel >= dev_config->channels) {
+		LOG_ERR("Invalid channel %u", channel);
+		return -EINVAL;
+	}
+
+	key = k_spin_lock(&data->lock);
```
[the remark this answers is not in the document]
cfriedt: Hmm... I guess because there are early returns it might be tricky. Not a big deal here, but it would be nice to have.

`K_SPINLOCK_BREAK` will break out of a spinlock properly, but it might get confused if already inside of a while loop.

drivers/dma/dma_arc_hs.c
```diff
@@ -0,0 +1,1011 @@
...
+	chan->state = ARC_DMA_PREPARED;
+
+	/* Make a copy of the first block configuration */
+	if (config->head_block) {
+		chan->block_config = *config->head_block;
+		/* Update the config to point to our copy */
+		chan->config.head_block = &chan->block_config;
+	}
+
+	k_spin_unlock(&data->lock, key);
```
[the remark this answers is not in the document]
ShLiTT: why this is not needed? I guess `K_SPINLOCK` will automatically unlock the lock after executing the block?

cfriedt (approved):

PR #659, opened 2025-09-29
scripts: dump_smc_stack.py

jgrowdenTT:
AI created script to dump stacks. Can hexdump the full stack, or try to do symbol analysis based on where it thinks the TCB and current stack offset is.


slindsayTT (approved):
didn't look too carefully but seems reasonably structured.
I'm assuming this is (for now at least) for internal use so I don't see why we wouldn't want to have the script.

PR #660, opened 2025-09-29
docs: add visualization of CI job status

danieldegrasse:
Add a job to visualize the results of all CI runs against `main` for e2e-stess tests. This visualization should help track the state of long tests, so we can see when issues are introduced into the tests

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,8 +16,143 @@ env:
   DOXYGEN_VERSION: 1.14.0
 
 jobs:
+  ci-stats:
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


cfriedt:
Long term, I think it's better to use a service that is probably run by TT devinfra, or possibly Grafana if there is no better alternative, but for the short term, why not?

ShLiTT:
> Long term, I think it's better to use a service that is probably run by TT devinfra, or possibly Grafana if there is no better alternative, but for the short term, why not?

Yes, that’s our long-term goal. Daniel and I will be meeting with Petra next Monday. We’re considering using Grafana as the backend and hosting a web page to pull data in a timely manner.

PR #667, opened 2025-09-30
CI: p300 rescue

jgrowdenTT:


danieldegrasse:
I think the fix to get P300a to really be reliable may be setting up flashing unit test binaries with tt-bootstrap, which is a bigger change. If we can figure out the reason for the binary data mismatches that are loaded by openocd, that would also get us past a lot of this, right?

```
Error: checksum mismatch - attempting binary compare
diff 0 address 0x1001a108. Was 0x92 instead of 0x25
diff 1 address 0x1001a109. Was 0x31 instead of 0x63
diff 2 address 0x1001a10a. Was 0x1d instead of 0x3a
diff 3 address 0x1001a10b. Was 0x30 instead of 0x20
No more differences found.
```

ShLiTT (approved):

jgrowdenTT:
We've had some internal chats. 

I've ran the test that uses JTAG to load up SMC a lot, and find it much more reliable if I lower the clock frequency considerably.  I'm still a little worried I might be changing the wrong thing, and I'm just misinterpreting patterns where there are none. Still, this commit as it is passes all the time vs commits outside of this one generally don't. Still, I advise we attempt to move away from JTAG and use something else to flash, as per Daniel's suggestion.

Furthermore, I identified two tests that cause issue on the P300a. I am very confident that these two tests are causing issue, removing them makes a night-and-day difference. I believe we're aligned that this is a regression, but I'm not sure if it was just made more common, or if it was a new introduction.

 Our next work will be to find it. But for now, the test fails, so we're basically forced to take it out since no immediate fix was apparent.

PR #672, opened 2025-10-01
boards: remove all p100 related files/workflows

ShLiTT:
According to Daniel Kruczek:

> [@Melanie Crain](https://tenstorrent.enterprise.slack.com/team/U024LPM19Q9) confirmed, no need to support Scrappy.  We have instructed everyone to remove those cards from their dev environments.


alexapostolu:
There's another mention [here](https://github.com/ShLiTT/tt-zephyr-platforms/blob/694939cbcf4dc8bd4dd998729fee3c1bbb44eb9b/include/zephyr/drivers/misc/bh_fwtable.h#L30) in the `.c` files, but it's probably harder to remove

PR #673, opened 2025-10-01
boards: remove p100 support

ShLiTT:
Confirmed with Daniel Kruczek, scrappy support can be removed
> [@Melanie Crain](https://tenstorrent.enterprise.slack.com/team/U024LPM19Q9) confirmed, no need to support Scrappy.  We have instructed everyone to remove those cards from their dev environments.


cfriedt:
@ShLiTT - can you fill in the PR description?

cfriedt:
The one thing that kind of concerns me with this is that, afair, we also used the p100 target for characterization boards.

Just wanted to make people aware, that by removing this target, we are also removing support for CB in the lab.

.github/workflows/docs.yml
```diff
@@ -212,7 +212,7 @@ jobs:
 
       - name: Copy CI Results
         run: |
-          for board in p100 p100a p150a p300a; do
+          for board in p100a p150a p300a; do
```
cfriedt, line 215: This should be changed to reference `.github/boards.json` as well.

boards/tenstorrent/tt_blackhole/revision.cmake
```diff
@@ -1,5 +1,5 @@
 set(BOARD_REVISIONS
-    "p100" "p100a" "p150a" "p150b" "p150c" "p300a" "p300b" "p300c" "galaxy")
+    "p100a" "p150a" "p150b" "p150c" "p300a" "p300b" "p300c" "galaxy")
```
cfriedt, line 2: Can this also be updated to reference `.github/boards.json` as well?

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -18,7 +18,7 @@
 #include <zephyr/logging/log.h>
 
 #define BOARDTYPE_ORION 0x37
-#define BOARDTYPE_P100  0x36
+/* #define BOARDTYPE_P100  0x36 - removed, p100 no longer supported */
```
cfriedt, line 21: Just remove it

lib/tenstorrent/bh_arc/regulator_config.c
```diff
@@ -272,6 +272,7 @@ static const RegulatorData serdes_vr_data[] = {
 
 /***** End of common serdes VR settings */
 
+/* p100 configuration removed - p100 no longer supported
```
cfriedt, line 275: We should just remove things instead of commenting them. They are still preserved in git history.

lib/tenstorrent/bh_arc/regulator_config.h
```diff
@@ -26,7 +26,7 @@ typedef struct {
 	uint32_t count;
 } BoardRegulatorsConfig;
 
-extern const BoardRegulatorsConfig p100_regulators_config;
+/* extern const BoardRegulatorsConfig p100_regulators_config; - p100 no longer supported */
```
cfriedt, line 29: Just remove it

.github/workflows/docs.yml
cfriedt: Yes, that would be good to go into another PR. @danieldegrasse - I didn't catch this in your PR earlier, unfortunately.

The main reason to reference the .json file is that we should avoid having arbitrary lists of boards scattered throughout our CI, because the lists do get desynchronized, and it's hit us in the past.

boards/tenstorrent/tt_blackhole/revision.cmake
cfriedt: Sure

cfriedt (approved):

PR #675, opened 2025-10-01
readme: add links to CI history pages

danieldegrasse:
Now that CI history pages are deployed with doc build, link to them from our readme. These CI pages provide historical statistics on the hardware-long workflow, to determine pass rates for tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,11 @@
 [![HW Smoke](https://github.com/tenstorrent/tt-zephyr-platforms/actions/workflows/hardware-smoke.yml/badge.svg?branch=main)](https://github.com/tenstorrent/tt-zephyr-platforms/actions/workflows/hardware-smoke.yml)
 [![HW Soak](https://github.com/tenstorrent/tt-zephyr-platforms/actions/workflows/hardware-long.yml/badge.svg?branch=main)](https://github.com/tenstorrent/tt-zephyr-platforms/actions/workflows/hardware-long.yml)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,11 @@
 [![HW Smoke](https://github.com/tenstorrent/tt-zephyr-platforms/actions/workflows/hardware-smoke.yml/badge.svg?branch=main)](https://github.com/tenstorrent/tt-zephyr-platforms/actions/workflows/hardware-smoke.yml)
 [![HW Soak](https://github.com/tenstorrent/tt-zephyr-platforms/actions/workflows/hardware-long.yml/badge.svg?branch=main)](https://github.com/tenstorrent/tt-zephyr-platforms/actions/workflows/hardware-long.yml)
 
+[P100 CI History](https://docs.tenstorrent.com/tt-zephyr-platforms/p100_ci_stats.html)
```
jgrowdenTT, line 8: LGTM but the graphs looks kinda weird. Y axis goes from 0 to 10 but I think it's pretty clear it should go to 100.

Is it just my browser?

https://docs.tenstorrent.com/tt-zephyr-platforms/p300a_ci_stats.html

jgrowdenTT: Sorry. Just realized it's interactive and I can drag it around.

danieldegrasse: No, I configured the graph that way in the initial build. My thought was that if we kept the Y axis smaller on initial load then intermittent failures (IE 1%) would be more obvious.

If you like I can change the rendering code in this PR to upload graphs with an initial range of 0-100

PR #679, opened 2025-10-01
WIP: lib: tenstorrent: bh_arc: add VF curve computation ARC messages + tests

alayander:
Implements ARC messages to compute points on VF curve given a frequency/voltage.
Also adds tests to verify implementation for VF curve message handler.


lib/tenstorrent/bh_arc/vf_curve.c
```diff
@@ -38,8 +45,60 @@ void InitVFCurve(void)
...
+	if (freq_mhz < 0.0F) {
+		response->data[1] = 0U;
+	} else {
+		response->data[1] = (uint32_t)(freq_mhz + 0.5F);
+	}
+
+	return 0;
+}
+
+REGISTER_MESSAGE(TT_SMC_MSG_GET_VOLTAGE_CURVE_FROM_FREQ, get_voltage_curve_from_freq_handler);
```
jgrowdenTT, line 103: I know this is in draft but before I forget -> please follow the precedent set by `TT_SMC_MSG_POWER_SETTING`

-> Create a struct in `msgqueue.c` outlining your message type, and put it in the `request` union
    -> Add doxygen comments to the struct you've created, linking it back to the function handling requests of that structure.
-> Use the struct in place of `request->data` (e.g. `float input_voltage_mv = request->get_freq_curve_from_voltage.input_voltage_mv;` instead of `float input_voltage_mv = (float)request->data[1];`) 
-> Link the enum in `smc_msg.h` to the struct you've created.

Do this for both `TT_SMC_MSG_GET_VOLTAGE_CURVE_FROM_FREQ` and `TT_SMC_MSG_GET_FREQ_CURVE_FROM_VOLTAGE` since they're currently documented as "not supported"

Then double check the artifacts to make sure all the doxygen links are blue.

cfriedt (changes requested):
Why are we using an ARC message for testing the VF Curve calculation?

We know the 3 floating point constants in advance, we know the math used to compute voltage from an input frequency in advance. We can also just use native_sim to verify that the math works as expected without an ARC message.

Why use an ARC message to make Blackhole effectively a pocket calculator?

What feature are we adding, and how should we go about testing it?

tests/lib/tenstorrent/bh_arc/src/vf_curve.c
```diff
@@ -0,0 +1,82 @@
...
+	req.data[0] = TT_SMC_MSG_GET_FREQ_CURVE_FROM_VOLTAGE;
+	req.data[1] = calculated_voltage;
+
+	msgqueue_request_push(0, &req);
+	process_message_queues();
+	msgqueue_response_pop(0, &rsp);
+
+	calculated_freq = rsp.data[1];
+
+	/* Verify roundtrip accuracy (allow small error due to integer conversion) */
```
cfriedt: Not blocking, but we have [zassert_within()](https://docs.zephyrproject.org/latest/doxygen/html/group__ztest__assert.html#gacc930af1a66e8533c5eca9526e198b4b) for this, specifically.

PR #681, opened 2025-10-01
zephyr: patches: update upstream pwm test patch and local test-conf

alayander:
Remove pwm_api testsuite patch since it was merged upstream: https://github.com/zephyrproject-rtos/zephyr/pull/96442
Also modify local test-conf to be compatible with that implementation.

> [!Note]
> Upstream commit b4d5cd3 already includes `zephyr/drivers-i2c-i2c-stm32-v2-stability-improvements.patch`, so that patch was removed as well. The smbus cancel patch required minor adjustments.

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -279,8 +279,8 @@ def arc_watchdog_test(asic_id):
         # does not currently have specific exception types
         logger.error("DMC reset too early, ARC should still be online")
         return False
-    # Delay 600 more ms. Make sure the ARC has been reset.
-    time.sleep(0.6)
+    # Delay 1000 more ms. Make sure the ARC has been reset.
+    time.sleep(1.0)
     # Delete arc chip object, to make sure pyluwen drops open file descriptors
     del arc_chip
     try:


alayander:
The compliance failure is related to the Kconfig option introduced in the patch/upstream
> ERROR   : Test Kconfig failed: 
> Found references to undefined Kconfig symbols.

cfriedt:
> @alayander - don't forget to modify [this line](https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/west.yml?rgh-link-date=2025-10-02T16%3A47%3A49Z#L20) in `west.yml` if needed. I don't know off-hand if the current commit also includes your PR

I'll add the change to`west.yml`. Should point at revision b4d5cd3c474652478c6b0e55819dcf012afe7d13

app/smc/pytest/e2e_smoke.py
```diff
@@ -279,8 +279,8 @@ def arc_watchdog_test(asic_id):
         # does not currently have specific exception types
         logger.error("DMC reset too early, ARC should still be online")
         return False
-    # Delay 600 more ms. Make sure the ARC has been reset.
-    time.sleep(0.6)
+    # Delay 1000 more ms. Make sure the ARC has been reset.
```
danieldegrasse, line 282: While this isn't technically invalid, it does suggest something might be up with the watchdog functionality? The 0.6 second delay plus 0.5 second stop earlier should result in the ARC being reset- we ask for a 1000 ms timeout expiry, I would expect the timeout to expire within 1100 ms. We may need to see if something changed with the DMC timing code.

PR #683, opened 2025-10-02
fw_table: make p300 tdp more conservative

afongTT:
On request from PCB power team, make p300 TDP limits more conservative, preventing each side from taking more than 50% of the total board power.

Numbers from [SYS-2123](https://tenstorrent.atlassian.net/browse/SYS-2123)


alexapostolu:
Can one side take 45% of total power and the other side take 55% total power?

cfriedt (approved):

afongTT:
> Can one side take 45% of total power and the other side take 55% total power?

The TDP limit is set in a way that loosely targets 50/50, so 45/55 is probably possible.

PR #694, opened 2025-10-03
ci: fix e2e-stress invocation

danieldegrasse:
Fix e2e-stress script invocation so that tests will actually run


alexapostolu (approved):

danieldegrasse:
Will remove WIP commit once the tests have actually started in CI on this PR. No need to run them for the full hour, I just made a mistake with the script call in the action definition

PR #695, opened 2025-10-03
scripts: dmc_reset + pcie_utils: reduce verbosity

cfriedt:
Currently, the dirty reset test logs about 6000 messages for its 1000 test iterations, and probably close to 5500 of those messages do not serve a useful purpose.

It's better to log errors with useful information than to log the same information over and over again for what is just expected behaviour.

This should make it easier to comb through CI logs to determine the root cause of a failure.

Example output prior to this change.
https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/18225967418/job/51897212403?pr=694

Example output after this change:
```shell
INFO: Dirty reset test iteration 0/1000
INFO: Dirty reset test iteration 10/1000
```

### DIFF
--- a/scripts/dmc_reset.py
+++ b/scripts/dmc_reset.py
@@ -171,10 +171,9 @@ def reset_dmc(args):
             "stderr": subprocess.DEVNULL,
         }
 
-    logger.debug("running command: " + " ".join(openocd_cmd))
-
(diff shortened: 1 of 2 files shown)


danieldegrasse (approved):
Thanks for cleaning this up

PR #724, opened 2025-10-10
zephyr: blobs: pcie completion timeout disabled

waynengTT:
-updated fw and checksum files to reflect updates
-verified change with lspci


cfriedt:
* rebased to get rid of merge commit

@waynengTT - just a heads-up, but we don't typically do merge commits in our codebase, and we cannot merge PRs that contain them. Fast-forward / rebase only.

PR #729, opened 2025-10-15
app: smc: pytest: Fw upgrade test

jgrowdenTT:
Add test that upgrades from 18.10 to 19.0. Use the blackhole recovery script to allow the downgrade.

Add version checking; verify against DMC/SMC VERSION files.

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -251,7 +251,13 @@ def upgrade_from_version_test(
 
 def test_upgrade_from_18_10(arc_chip_dut, tmp_path: Path, board_name, unlaunched_dut):
     upgrade_from_version_test(
-        arc_chip_dut, tmp_path, board_name, unlaunched_dut, "18.10.0", (13 << 16), (19 << 16)
+        arc_chip_dut,
+        tmp_path,
+        board_name,
+        unlaunched_dut,
+        "18.10.0",
+        (13 << 16),
+        (19 << 16),
     )
 
 
--- a/app/smc/pytest/e2e_stress.py
+++ b/app/smc/pytest/e2e_stress.py
@@ -223,13 +223,31 @@ def test_dmc_ping(arc_chip_dut, asic_id):
 
 def test_upgrade_from_18x(tmp_path: Path, board_name, unlaunched_dut, arc_chip_dut):
     upgrade_from_version_test(
-        arc_chip_dut, tmp_path, board_name, unlaunched_dut, "18.10.0", (13 << 16), (19 << 16)
+        arc_chip_dut,
+        tmp_path,
+        board_name,
+        unlaunched_dut,
+        "18.10.0",
+        (13 << 16),
+        (19 << 16),
     )
 
     upgrade_from_version_test(
-        arc_chip_dut, tmp_path, board_name, unlaunched_dut, "18.11.0", (14 << 16), (20 << 16)
+        arc_chip_dut,
+        tmp_path,
+        board_name,
+        unlaunched_dut,
+        "18.11.0",
+        (14 << 16),
+        (20 << 16),
     )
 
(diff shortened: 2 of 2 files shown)


app/smc/pytest/e2e_smoke.py
```diff
@@ -583,3 +599,59 @@ def test_aiclk(arc_chip_dut, asic_id):
...
+                recovery_py,
+                tar_recovery,
+                board_name,
+                "--force",
+                "--select-first",
+            ]
+        )
+    except subprocess.CalledProcessError as e:
+        logger.error(f"blackhole_recovery.py failed with error: {e}")
+        assert False
```
danieldegrasse: Might also log process output as an info or debug output here so we can confirm it worked in CI runs

app/smc/pytest/e2e_smoke.py
```diff
@@ -583,3 +599,59 @@ def test_aiclk(arc_chip_dut, asic_id):
         aiclk = arc_chip.arc_msg(ARC_MSG_TYPE_GET_AICLK)[0]
         assert aiclk == clk, f"Failed to set clock to {clk} MHz"
         logger.info(f"AICLK set to {aiclk} MHz successfully")
+
+
+def upgrade_from_version_test(tmp_path: Path, board_name, unlaunched_dut, version):
```
danieldegrasse: Generally looks good- a few things I might add:
- let's validate that the DMFW we are running matches the expected version (we can pull this from telemetry). That way we know that the update completed correctly. We can check CMFW version as well in telemetry.
- we could run this test first by defining it first in the file. If we do that we may be able to skip flashing/launching the DUT later, and improve test execution times by only flashing main once.

danieldegrasse (commented):
Generally looks good- appreciate this being possible to run outside of twister.

One thing I think we should add is a way to cross check that the update worked- I think the best way to do that is the verify the DMFW and CMFW versions in telemetry look right for the version flashed- that would catch a situation where the DMFW or CMFW didn't update for whatever reason

scripts/get_ttzp_version.py
```diff
@@ -8,14 +8,51 @@
...
+    minor = int(version_parts.get("VERSION_MINOR", 0))
+    patchlevel = int(version_parts.get("PATCHLEVEL", 0))
+    tweak = int(version_parts.get("VERSION_TWEAK", 0))
+
+    # Ensure values fit in their respective bit ranges
+    if not 0 <= major <= 255:
+        raise ValueError(f"VERSION_MAJOR {major} must be 0-255")
+    if not 0 <= minor <= 255:
+        raise ValueError(f"VERSION_MINOR {minor} must be 0-255")
+    if not 0 <= patchlevel <= 255:
```
alexapostolu: This might be beyond the scope of the test if it's just testing the bit ranges, but shouldn't this be `0-98` since we can't go beyond the main branch version `99`

jgrowdenTT: Thanks, I wasn't aware of that limitation! Are all fields limitted to 99?

app/smc/pytest/e2e_smoke.py
```diff
@@ -141,6 +175,86 @@ def arc_chip_dut(launched_arc_dut, asic_id):
...
+        assert False
+
+    # flash "base" firmware to update from
+    URL = f"https://github.com/tenstorrent/tt-firmware/releases/download/v{version}/fw_pack-{version}.fwbundle"
+    targz = tmp_path / "fw_pack.tar.gz"
+
+    urlretrieve(URL, targz)
+    try:
+        result = subprocess.run(
+            ["tt-flash", "flash", "--fw-tar", str(targz), "--force"],
```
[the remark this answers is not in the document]
danieldegrasse: We are on 3.4.7 now, yeah. Have to be for next release

danieldegrasse (approved):

PR #738, opened 2025-10-16
DNM: wip: use luwen that respects wait for done

danieldegrasse:
Use luwen version that does not timeout if wait_for_done is false, to see if this resolves BL2 flashing issues.


danieldegrasse:
Not worth pursuing further IMO- with recent versions of tt-flash the error *seems* to be gone from our CI, and since v19.0.0 requires latest version of tt-flash users are unlikely to see whatever this issue was

PR #739, opened 2025-10-16
scripts: tooling: blackhole_recovery: support overriding image variables

danieldegrasse:
Support overriding image variables in blackhole recovery script. Also, force recovery if a job is retried.


danieldegrasse:
This script has been pulled onto all of our self hosted runners as the pre/post job hook- so if you restart a job on a PR, the runner will now run card recovery before and after. This should help fix any cases where the recovery script did not run previously, and the card was in a bad state as a result.

Ideally we would simply automatically recover the card if the job failed, but that doesn't seem to be possible using the environment variables that github jobs have available: https://docs.github.com/en/actions/reference/workflows-and-actions/variables#default-environment-variables

PR #744, opened 2025-10-16
app: smc: make BL2 the default build scheme

danieldegrasse:
Switch BL2 to be the default build scheme for the firmware application. Update tests to support running with BL2.


danieldegrasse:
Commits are pulled into #750

PR #745, opened 2025-10-17
poc: enable hotloading firmware

danieldegrasse:
This proof of concept is in an early stage. It supports hotloading
firmware over the original firmware, provided the memory region at
0x1007e000 is not overwritten by the new firmware.

The load sequence works as follows:
- build firmware
- load firmware using load-fw.py script, providing load address and jump
  address

Examples:

west build -p always -b tt_blackhole@p150a/tt_blackhole/smc --sysbuild app/smc/
./scripts/load-fw.py build/smc/zephyr/zephyr.bin 0x10010400 0x10010000

west build -p always -b tt_blackhole@p150a/tt_blackhole/smc \
               ../zephyr/samples/hello_world/
./scripts/load-fw.py build/zephyr/zephyr.bin 0x10000000 0x10000000

This is still *very* unreliable, sometimes the write to memory fails to complete. Still debugging why this is the case.

### DIFF
--- a/app/smc/boards/tt_blackhole_smc_partitions.dtsi
+++ b/app/smc/boards/tt_blackhole_smc_partitions.dtsi
@@ -32,7 +32,14 @@ slot1_partition: &smc_recovery {};
 		compatible = "zephyr,memory-region", "mmio-sram";
 		zephyr,memory-region = "CSM-APP";
 		device_type = "memory";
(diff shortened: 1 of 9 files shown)


slindsayTT:
I think we can just close this since we aren't likely to proceed.

danieldegrasse:
Closing out PR, fun little experiment though :)

PR #747, opened 2025-10-17
app: smc: pytest: e2e_stress.py: add pvt e2e tests

alexapostolu:
Add PVT e2e test for temp, process and voltage sensors.

### DIFF
--- a/app/smc/pytest/e2e_stress.py
+++ b/app/smc/pytest/e2e_stress.py
@@ -86,9 +86,9 @@ def test_arc_watchdog(arc_chip_dut, asic_id):
             fail_count += 1
 
     report_results(test_name, fail_count, total_tries)
-    assert (
-        fail_count <= failure_fail_count
-    ), f"{test_name} failed {fail_count}/{total_tries} times."
+    assert fail_count <= failure_fail_count, (
+        f"{test_name} failed {fail_count}/{total_tries} times."
+    )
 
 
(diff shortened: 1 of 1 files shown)


app/smc/pytest/e2e_stress.py
```diff
@@ -251,3 +258,89 @@ def test_upgrade_from_18x(tmp_path: Path, board_name, unlaunched_dut, arc_chip_d
...
+            response = arc_chip.arc_msg(
+                ARC_MSG_TYPE_READ_TS, True, False, sensor_id, 0, 5000
+            )
+
+            temp = response[2]
+            if temp < 40 or temp > 70:
+                fail_count += 1
+
+    report_results(test_name, fail_count, total_tries)
+    assert fail_count, f"{test_name} failed {fail_count}/{total_tries} times"
```
cfriedt: This is only checking that `fail_count` is not `None`.

What we shoudl do here is validate that the failure rate is less than a specific number. I.e. that `fail_count / total_tries < max_failure_rate`

cfriedt:
I would suggest defining a maximum failure rate that we are comfortable with (e.g. 1/1000, or 0.1%), such that the maximum failure rate is practically never reached, and then asserting based on that.

cfriedt:
@alexapostolu - LGTM. I think we can probably merge this if you want to mark it ready for review.

Anything like tests that improve stability are fine during the stabilization period.

PR #750, opened 2025-10-17
app: smc: switch to bl2 by default and verify signatures in ci

cfriedt:
* app: smc: ensure bl2 builds sign smc fw and mcuboot using the specified key
* pull in all external structural upgrade dependencies
* scripts: tt_flash_runner: support major downgrades argument
* scripts: ci: pass new flags to tt-flash
* scripts: tooling: blackhole_recovery: allow bypassing adapter prompt
* app: smc: make BL2 the default build process
* app: smc: update testcases to remove BL2 test
* app: smc: use existing SB_CONFIG_BOOT_SIGNATURE_KEY_FILE
* boards: tt_blackhole: galaxy: delete blupdate and dmfwimg partitions
* ci: build-fw: verify signed firmware

Access to signing keys is not permitted in pull requests.

An example of the `build-fw` workflow that does not sign FW is in this PR itself, because it's from a fork of the project. Github secrets are not available on PRs originating from external forks.
https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/18605651443/job/53054414920?pr=750

```shell
Verify firmware signature
Run if [ -z "$SIGNATURE_KEY_FILE" ]; then
SIGNATURE_KEY_FILE is not set, skipping signature verification
```

An example of the `build-fw` workflow that _does_ sign FW is at https://github.com/cfriedt/tt-zephyr-platforms/pull/9. A locally-generated signing key is kept in a secret within the forked project.

> [!NOTE]
> GitHub redacts the contents of the key in logs when referenced via `${{ secrets.SIGNATURE_KEY }}`.
> Workflows should only ever refer to `$SIGNATURE_KEY_FILE` and take care to vet any code that ever interacts with this file.

https://github.com/cfriedt/tt-zephyr-platforms/actions/runs/18605651337/job/53054410060?pr=9

```shell
Run SIGNATURE_KEY_FILE="$(mktemp)"
  echo "***
  ...
  ***" | base64 -d > "$SIGNATURE_KEY_FILE"
  SZ=$(wc -c "$SIGNATURE_KEY_FILE" | awk '{print $1}')
  if [ $SZ -gt 42 ]; then
    echo "SIGNATURE_KEY_FILE=$SIGNATURE_KEY_FILE" >> "$GITHUB_ENV"
  fi
  ...
Verify firmware signature
...
    SIGNATURE_KEY_FILE: /tmp/tmp.Y8xkqThus7
Image was correctly validated
Image version: 0.16.0+0
Image digest: bb8a62af1451d09bc8372e31958f5c87acd006ab58722cc2f957a4f1274c709b
Image was correctly validated
Image version: 0.0.0+0
Image digest: e8870d4d87426b8a9d7ff6391fb29657ef28f78352875f498c20e574db32c9e7
Image was correctly validated
Image version: 0.22.0+0
Image digest: 1b91ab46e3b18fc0ae8e17d66fe49459bfdb492096266f11aec8aaa0e820a496
```

Any workflows that are triggered via schedule, via tag, or manually (not originating from an outside PR) will generate signed binaries with the production key.

> [!NOTE]
> If the production key is not used to sign binaries, then the default (insecure) key is used. Binaries are always signed.

### DIFF
--- a/app/smc/sysbuild/mcuboot-bl2.conf
+++ b/app/smc/sysbuild/mcuboot-bl2.conf
@@ -1,4 +1,5 @@
 # Copyright (c) 2025 Tenstorrent AI ULC
 # SPDX-License-Identifier: Apache-2.0
+CONFIG_BOOT_VALIDATE_SLOT0=n
 CONFIG_BOOT_SWAP_USING_OFFSET=y
 CONFIG_BOOT_BOOTSTRAP=y


app/smc/sysbuild.cmake
```diff
@@ -62,86 +71,70 @@ if(NOT "${BOARD_REVISION}" STREQUAL "galaxy")
...
-  # build, we need to create a tiny Zephyr module in the source directory, and
-  # tell sysbuild to include that module with mcuboot
-  set(mcuboot_EXTRA_ZEPHYR_MODULES "${CMAKE_CURRENT_LIST_DIR}/mcuboot_module" CACHE INTERNAL "mcuboot_module directory")
-endif()
+# This is quite a hack- in order to get the mcuboot hook file we need to override
+# image verification (which allows us to skip signature checks) into the mcuboot
+# build, we need to create a tiny Zephyr module in the source directory, and
+# tell sysbuild to include that module with mcuboot
+set(mcuboot_EXTRA_ZEPHYR_MODULES "${CMAKE_CURRENT_LIST_DIR}/mcuboot_module" CACHE INTERNAL "mcuboot_module directory")
+set_config_string(mcuboot CONFIG_BOOT_SIGNATURE_KEY_FILE "${SB_CONFIG_BOOT_SIGNATURE_KEY_FILE}")
```
danieldegrasse: Should we be checking if `SB_CONFIG_BOOT_SIGNATURE_KEY_FILE` is empty here? Or will it get a default value of the key in the mcuboot repo?

cfriedt:
For anyone who wants to test out this PR, I would use the following steps (obviously change `p100a` if you have a different card):

Install `gh` command line tool as described [here](https://github.com/cli/cli#installation).
```shell
# create a local key
imgtool keygen -k mykey.pem -t rsa-2048

# fetch the code in this branch
gh pr checkout 750

# build FW
west build --sysbuild -p -b tt_blackhole@p100a/tt_blackhole/smc app/smc --DCONFIG_COMPILER_WARNINGS_AS_ERRORS=y -DSB_CONFIG_BOOT_SIGNATURE_KEY_FILE=\"$PWD/mykey.pem\"

# verify keys
for i in $(find build -name '*.signed.bin'); do echo $i; imgtool verify -k mykey.pem $i; done

# check .config files
grep BOOT_SIGNATURE_KEY_FILE $(find build -name '.config')
```

> [!NOTE]
> If flashing with a signing key other than the default one in `bootloader/mcuboot/root-rsa-2048.pem`, later re-flashing with binaries signed with another key will require FW recovery.

app/smc/sysbuild.cmake
danieldegrasse: Perfect- if you have the bandwidth, let's drop the other checks in this file then?

danieldegrasse:
> @danieldegrasse - added `CONFIG_BOOT_VALIDATE_SLOT0=n`. We should probably go back to using `app/dmc/sysbuild/mcuboot.conf at some point.

Agreed. Feel free to pull in the changes from #754 here if you'd like- it includes the changes for the DMC app itself. If not, we can merge that PR after- we will need it for 19.0 to generate the correct recovery images.

PR #756, opened 2025-10-20
fw_table: Update minimum frequency to 500

jgrowdenTT:
Change the ASIC min frequency to 500 for additional power savings


afongTT:
I'm not too sure about this. All of qual was run with AICLK min at 800 MHz. We may be opening ourselves up to functional issues.

@tcolacoTT / @bingliTT what testing should we do if we really want to drop AICLK min to 500 MHz? What if we only go to 500 MHz with some communication from KMD that the device is idle?

jgrowdenTT:
This is going to get tabled until, at least, we have better CI coverage of the implications of this change.

PR #494, opened 2025-08-11
tenstorrent/tt-metal#24359: Update to include UMD tests

tt-rkim:

PR #770, opened 2025-10-21
CI: Fix unlaunched dut import

jgrowdenTT:
Some CI stress fixes:

1) Fix the unlaunched dut import when running via twister.

If twister exists, we do not create the unlaunched_dut fixture in e2e_smoke. So don't bother with trying to import anything. This works to get stress tests working locally via `run-e2e.sh`  and worked in this PR with some yaml mods to run stress on PR (although we still fail for other reasons...)

### DIFF
--- a/app/smc/pytest/e2e_stress.py
+++ b/app/smc/pytest/e2e_stress.py
@@ -21,7 +21,13 @@ from e2e_smoke import (
 
 # Needed to keep ruff from complaining about this "unused import"
 # ruff: noqa: F811
-from e2e_smoke import arc_chip_dut, launched_arc_dut, unlaunched_dut  # noqa: F401
+from e2e_smoke import arc_chip_dut, launched_arc_dut  # noqa: F401
+
+try:
+    from e2e_smoke import unlaunched_dut  # noqa: F401
+except ImportError:
+    pass  # we should have it from twister
+
 
 logger = logging.getLogger(__name__)
 


jgrowdenTT:
> @jgrowdenTT - can you also look at this diff to attempt to fix the .cargo issue?
> 
> I noticed that `HOME` might not have been used correctly before.
> 
> [`a2f8985` (#743)](https://github.com/tenstorrent/tt-zephyr-platforms/pull/743/commits/a2f8985cee7301b5c0ae96449bbde9f0ac998636)

Looks like metal tests are still failing there.

But yeah, if you think the home directory is used inconsistently, fixing that is going to be important... With regards to the validity of this PR, I still don't know if the `chmod` avoidance here is a "robustness" improvement vs "hiding an issue."

I think maybe we try to get yours off the ground, if we end up figuring it out great, if we struggle we can always put this in as a temporary patch.

cfriedt:
> But yeah, if you think the home directory is used inconsistently, fixing that is going to be important

That was actually the root cause.

```shell
chmod -R a+rwX $HOME/.cargo $HOME/.cache
```
expanded to 
```shell
chmod -R a+rwX /github/home/.cargo /github/home/.cache
```

Meanwhile, the working directory was set to `/home/user/tt-metal` and that was hard-coded into several commands.

That resulted in the error
```
chmod: cannot access '/github/home/.cargo': No such file or directory
```

Making `HOME=/home/user/tt-metal` should hopefully ensure things are consistent.

> Looks like metal tests are still failing there.

Just missing a `mkdir -p $HOME/.cargo $HOME/.cache`.

I'm likely going to tack that on to #494

PR #771, opened 2025-10-21
scripts: tt_boot_fs: print image digest if signed

danieldegrasse:
Print the digest of images that are signed with an mcuboot header. This
will improve our ability to verify reproducible builds post release,
since RSA-PSS signatures use a random salt value that makes each signature
non-reproducible


scripts/tt_boot_fs.py
```diff
@@ -807,9 +808,19 @@ def ls(
...
+            hdr_len = 0
+            for char in hdr:
+                if char == "\t":
+                    # Tab stops expand to next 8 byte boundary
+                    if hdr_len % 8 == 0:
+                        hdr_len += 8
+                    else:
+                        hdr_len += 8 - (hdr_len % 8)
+                else:
+                    hdr_len += 1
```
cfriedt: This seems a lot more complicated than what is already there. Is it necessary?

PR #772, opened 2025-10-21
blobs: update wormhole fw blob

afongTT:
ERISC 6.7.2.0
- Add support to trigger retraining on failed training ethernet ports

### DIFF
--- a/doc/release/release-notes-19.0.md
+++ b/doc/release/release-notes-19.0.md
@@ -64,6 +64,9 @@ Major enhancements with this release include:
   * Added eth_reinit state machine to handle fail case when port is up
 * PVT Sensor
   * correct PVT RTIO buffer size and frame count for decode
+* Update Wormhole FW blob
+  * ERISC 6.7.2.0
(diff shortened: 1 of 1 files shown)


zephyr/module.yml
```diff
@@ -197,9 +197,9 @@ blobs:
     description: "Modified bootrom to boot Blackhole"
     doc-url: https://tenstorrent.com
   - path: fw_pack-wormhole.tar.gz
-    sha256: 824d86a2ff0dd195e9a71bc85211e1b06b88332e1859142f9eeef1f828642001
+    sha256: 14f481be865e944327fc9fdc77a30a7fb1abb45ab7783923d675d9bbc7562a59
     type: img
-    version: '18.12.0-rc1'
+    version: '19.0.0-rc3'
```
cfriedt: Hopefully not overly optimistic, but I think our next version will be 19.0.0, so let's use that here
```suggestion
    version: '19.0.0'
```

PR #775, opened 2025-10-22
 app: smc: pytest: add 19.0 upgrade test

danieldegrasse:
~Add test to validate updating from one version of 19.0 firmware to
another version. Since no released version of 19.0 firmware exists, we
currently flash a custom recovery bundle with 19.0 partitions but 18.0
version numbers, and validate upgrades from that.~

Add test to validate we can update from 19.0 to latest built firmware. Since firmware builds do not get signed with production keys on PRs, we need to replace the running copy of MCUBoot on the DMC with one that will accept our signature key.

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -374,18 +374,6 @@ def temperature_sensors_test(arc_chip_dut, asic_id):
     return fail_count
 
 
-def test_upgrade_from_18_10(arc_chip_dut, tmp_path: Path, board_name, unlaunched_dut):
-    upgrade_from_version_test(
-        arc_chip_dut,
-        tmp_path,
-        board_name,
-        unlaunched_dut,
-        "18.10.0",
-        (13 << 16),
-        (19 << 16),
-    )
-
-
 def test_upgrade_from_19_00(arc_chip_dut, tmp_path: Path, board_name, unlaunched_dut):
     upgrade_from_version_test(
         arc_chip_dut,


danieldegrasse:
Everything is passing here, but I'm not sure if we want to merge as is. Personally, I think the fw update tests might be better to run on main only, and not per-pr (for execution time reasons)

@jgrowdenTT, any opinion on this?

jgrowdenTT:
Kinda controversial one...

In the case of the 18x -> 19.main upgrade tests, I think the DUT is the 19.main FW version. We're making sure a person on old FW can upgrade to 19.main

In this case. It sorta looks like we're proving that 19.0 supports going to main, in the sense that 19.0 is the DUT. 

This is critical to test before release because 19.0 is set in stone. Inversely, because 19.0 is set in stone, there's no regression possibilities. So I think this at best goes into stress and at worst, we could consider scrapping it.

I think eventually when we have a 19.1, we do want to verify that 19.0 can go to 19.1. In that case I think we just ammend our existing 18.x test.

jgrowdenTT:
> PR now performs a 19.0 update test without any customized recovery bundle. We can either keep this on smoke, or move it to overnight testing

I think we're spending a lot of our per PR budget on upgrade tests.

How do we feel about keping this one in stress, but moving 18.10 to run nightly only?

jgrowdenTT:
I didn't even realize "pushes to main" was a possibility.

18.10 on pushes to main - sure. 19.0 stays in smoke or moves to pushes to main. I'm really fine with either, it's not set in stone. If we find it's too green we push it back, if we find it's red a lot (for "real" reasons) we move it per PR. 

18.10/18.11/18.12 Stay put in stress.

My 2c.

app/smc/pytest/e2e_smoke.py
```diff
@@ -249,15 +277,16 @@ def upgrade_from_version_test(
     ) == read_telem(0, TAG_CM_FW_VERSION)
 
 
-def test_upgrade_from_18_10(arc_chip_dut, tmp_path: Path, board_name, unlaunched_dut):
```
cfriedt: Would it make sense to keep the old test and add the new test too? Otherwise, `replace_bootloader` would always be true, so there wouldn't be a need for the conditional.

danieldegrasse: We will will use the old tests in nightly, so `replace_bootloader` will be false there

cfriedt (approved):

PR #777, opened 2025-10-22
lib: bh_arc: DMC->SMC ping improvements

jgrowdenTT:
Make the DMC->SMC ping SMBUS command a read-word command instead of a write-word command. This lets the DMC verify that the SMC actually processed the command by verifying against the PEC. Use this to make an effective retry loop on the DMC side.

Before this change, I would fail the `test_dmc_ping` within 1000 iterations quite commonly.

After this change, I have passed 3 million iterations without failure.


lib/tenstorrent/bh_arc/cm2dm_msg.c
ShLiTT: Sorry, it's more like a question...
Is A5A5 some random data for testing since we are doing SMBUS_READ?

danieldegrasse (approved):
Good with using the args approach in a follow up PR

PR #786, opened 2025-10-27
ci: hardware-long: separate metal-test into its own workflow

cfriedt:
Separate metal-test into its own workflow and give it a separate status badge on the landing page.


cfriedt:
> > Separate metal-test into its own workflow and give it a separate status badge on the landing page.
> 
> Do we have an CI run verifying that this works? Doesn't have to run to completion, just would be good to make sure all the syntax here is right

https://github.com/cfriedt/tt-zephyr-platforms/actions/runs/18879157313
https://github.com/cfriedt/tt-zephyr-platforms/actions/runs/18879145319

PR #798, opened 2025-10-29
ci: hardware-long: trigger on PR based on label

danieldegrasse:
If the "long-tests" label is added to a PR, trigger hardware long tests on it.

This has the following conditions:

- the jobs won't be triggered unless a label is added (good)
     - this means we don't get a bunch of skipped jobs for hardware-long in the CI view (good)
- only those with write access to the repo can add labels (good)
- adding another label retriggers the logic for the jobs, and will cancel existing hardware long tests running on the PR (**bad**)
- pushing to the PR won't trigger tests to run again (**bad**)
- only way to rerun tests is to remove and readd the label (meh)

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -2,6 +2,9 @@ name: Hardware Long Tests
 
 # Run long tests once nightly, at 00:00
 on:
+  pull_request:
+    types:
+      - labeled
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch: {}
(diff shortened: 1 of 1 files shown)


danieldegrasse:
With this change, if the head commit of a PR contains `[long tests]` in the commit body, the PR will run e2e stress tests

cfriedt:
Still trying to understand what problem this PR is trying to address.

Typically, we don't run the long tests with PRs, but I can see why you might want to, because otherwise, a change would need to be in-tree before the hardware-long test will check it.

A label is probably a better way to conditionally enable long (aka "soak") tests than a string that might happen to be in a PR title (also PR titles can get annoying if they are filled with magic tags).

danieldegrasse:
I am please to report with the help of Claude I have discovered you can use an event like this as the trigger

```
 pull_request:
    types:
      - labeled
 ```

This has the following conditions:
- the jobs won't be triggered unless a label is added (good)
   - this means we don't get a bunch of skipped jobs for hardware-long in the CI view (good)
- adding another label retriggers the logic for the jobs, and will cancel existing hardware long tests running on the PR (**bad***)
- pushing to the PR won't trigger tests to run again (**bad**)
- only way to rerun tests is to remove and readd the label (meh)

danieldegrasse:
Try adding the label to this PR again if you like- it should trigger the long tests. If you add another label it will cancel the tests (removing the label won't cancel them)

PR #801, opened 2025-10-30
ci: build-fw: build: use a for loop instead of build matrix

cfriedt:
The build matrix unfortunately was N-tuplicating all of the work of fetching the Zephyr SDK, fetching Zephyr sources, fetching module sources, fetching python dependencies, fetching apt dependencies.

Even though some of these things are cached, it's a significant waste of time. For example, for `v19.1.0-rc1`, the build-fw workflow took 8 min 30s.

https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/18927610279

This change should reduce much of the N-tuplicated work.

### DIFF
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -41,11 +41,6 @@ jobs:
 
   build:
     runs-on: ubuntu-24.04
-    needs: genboards
-    strategy:
-      fail-fast: false
-      matrix:
-        board: ${{ fromJson(needs.genboards.outputs.boards) }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


cfriedt:
> Does this run faster if we do it sequentially? I think when I was trying this with recovery fw builds it was slower, even though we did have to fetch less data

It's way slower unfortunately. I guess one contributing factor is that with `--sysbuild -p` (I think `--sysbuild` also requires `-p`) we rebuild everything every time.

If there were only a way to cache all of the dependencies, I think the average build time would be around 3m.

Maybe some of the new tickets I created in the backlog will help.
- west apt package management
- apt package caching

Closing this PR, because it obviously isn't good to consider merging.

PR #809, opened 2025-11-03
drivers: dma: dma_tt_bh_noc.c: replace workq memcpy with noc to noc

alexapostolu:
The upstream test suite uses RAM to RAM for their DMA tests. Currently
we use a memcpy inside of a work queue to test it. This commit changes
it so we go via RAM to tensix to RAM. This passes upstream test suite
while also implicity testing RAM <-> tensix.


alexapostolu:
The fix was to make an upstream patch that allows the TX and RX data to be aligned to 64 bytes

PR #814, opened 2025-11-04
drivers: move to upstream mspi and mspi_nor drivers for flash

danieldegrasse:
Move from the downstream MSPI driver to upstream copy, since we have
upstreamed the changes we need to use this driver.

Move to the upstream MSPI NOR flash driver. We still need some small
patches for the driver, so add those to the patchlist

> [!NOTE]
> This change will remove the dynamic probe functionality from the flash
driver, so this driver will no longer support P100 cards (which have
been removed from tree)


danieldegrasse:
Reopened to restart CI

danieldegrasse:
This CI run includes extended SPI read/write tests: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/19084780972/job/54522455560?pr=814. Since these are passing I'm relatively confident that this change is valid.

danieldegrasse:
> LGTM

Sounds good- let's watch closely (and be suspicious of) any changes we see to flash test results in CI. I *think* this change is safe given the testing, but touching the flash code is always risky

PR #816, opened 2025-11-04
ci: limit P300a jobs to functional runner

danieldegrasse:
Limit P300a jobs to the functional runner until the other runner is properly reconfigured to support job execution

### DIFF
--- a/.github/ci_boards.json
+++ b/.github/ci_boards.json
@@ -11,7 +11,7 @@
   },
   {
     "board": "p300a",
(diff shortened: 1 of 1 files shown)


danieldegrasse:
I think since `yyz-zephyr-01` is already registered with the `p300a-jtag` token this is probably the easiest way to get it reprovisioned without causing issues with CI jobs

cfriedt (approved):
Sure - not 100% sure if we need to reprovision it, but it's definitely still borked (last time I looked)

PR #817, opened 2025-11-05
[Backport v18.12-branch] zephyr: patches: add patch to move i2c_dw stop callback

cfriedt:
Backport b16657cce32f71a07e9504fbc3c0fab4db8ab261 from #802

Currently, our recovery is done using `v18.12.0-rc1`, which still includes the bug that was fixed with #802.
We should probably apply the fix to the `v18.12-branch` and tag a new `v18.12.2` release to pull in the fix and ensure that recovery firmware exhibits the same reliable i2c / smbus communication that we see in `v19.1.0`. Then, rather than using `v18.12.0-rc1` for recovery, we should use `v18.12.2`.

### DIFF
--- a/zephyr/patches.yml
+++ b/zephyr/patches.yml
@@ -173,3 +173,13 @@ patches:
       and rejects all future transfers until cancellation is reversed.
 
       Implementations are included for STM32 I2C and SMBus.
+  - path: zephyr/drivers-i2c-i2c_dw-make-target-stop-callback-after-r.patch
+    sha256sum: f74129867892db21df4f328943ca8855f6fa5ca6a3780574fae6c730290dae8c
+    module: zephyr
+    author: Daniel DeGrasse
+    email: ddegrasse@tenstorrent.com
+    date: 2025-10-30
+    comments: |
+      Ensure that the I2C DesignWare target driver invokes the stop callback
+      after read operations complete. This ensures proper synchronization
+      for I2C target devices.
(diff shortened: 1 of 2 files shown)


danieldegrasse:
It looks like the P300 job is timing out each time. Maybe try extending the timeout on the test temporarily? The P150a test executes in 58 seconds, so it would be interesting to see how long the P300 test is taking. I don't think that 58 second figure includes the time for the flash step to run, so the P300 flash might be why the test is timing out

cfriedt:
The first two runs resulted in a timeout in `e2e_smoke.py::test_dirty_reset`

The last iteration ran into 
```
ERROR: Could not flash device None
---------------------------- live log sessionfinish ----------------------------
DEBUG: Device already disconnected
- generated xml file: /__w/tt-zephyr-platforms/tt-zephyr-platforms/zephyr/twister-smc-e2e/tt_blackhole@p300a_tt_blackhole_smc/zephyr/tt-zephyr-platforms/app/smc/app.vuart-bh-2x/report.xml -
!!!!!!!!!!!!!!!!!! _pytest.outcomes.Exit: Failed to flash DUT !!!!!!!!!!!!!!!!!!
============================ no tests ran in 14.86s ============================
INFO    - /__w/tt-zephyr-platforms/tt-zephyr-platforms/zephyr/twister-smc-e2e/tt_blackhole@p300a_tt_blackhole_smc/zephyr/tt-zephyr-platforms/app/smc/app.vuart-bh-2x/twister_harness.log
INFO    - 
```
https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/19102904268/job/54596417950?pr=817

cfriedt:
I think it would be worth it to try and build the recovery image with the version of FW in this commit, adjust the URLs used when the recovery image is fetched to a URL where this blob is stored, and then run it through CI that way.

If CI passes, and then maybe add another PR later to update the recovery URL to a published artifact.

How would I go about doing that @danieldegrasse ?

cfriedt:
Going to close this, because I don't really want to fight with CI, and we might be switching to a more recent release for recovery anyway, that already includes this fix.

PR #821, opened 2025-11-05
ci: build-fw: fetch tags from manifest repo and zephyr

danieldegrasse:
Fetch with fetch-depth 0 when building firmware. This is desirable because it means that firmware binaries built within the CI job will have the proper "git describe" version string. Without this change, version strings in released builds look like the following:

```
*** Booting tt_blackhole with Zephyr OS d90d71c42c6d ***
*** TT_GIT_VERSION 66ca1d5a6b56 ***
```

While this isn't _required_, without it reproducible builds are a bit harder to achieve (since you have to delete your tags locally first to get the same `git describe` behavior)

### DIFF
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -50,9 +50,19 @@ jobs:
       - uses: actions/checkout@v4
         with:
           path: tt-zephyr-platforms
+          fetch-tags: true
       - uses: ./tt-zephyr-platforms/.github/workflows/prepare-zephyr
         with:
           app-path: tt-zephyr-platforms
+
+      # Fetch tags from repositories in manifest. Needed so that
+      # git describe functions correctly when creating a build.
+      # We can remove this once https://github.com/zephyrproject-rtos/action-zephyr-setup/issues/56
+      # is fixed.
+      - name: Fetch Tags
+        shell: bash
+        run: |
+          west forall -c 'git fetch --unshallow || true'
 
       - name: Generate BOARD
         working-directory: tt-zephyr-platforms


danieldegrasse:
Console output from a build in this PR:
```
*** Booting tt_blackhole with Zephyr OS v4.2.0-4835-gb4d5cd3c4746 ***
*** TT_GIT_VERSION v19.1.0-11-g27f36f0ecb63 ***
*** SDK_VERSION zephyr sdk 0.17.4 ***
Tenstorrent Blackhole CMFW 0.23.99
```

Console output from v19.1.0 (release)
```
*** Booting tt_blackhole with Zephyr OS d90d71c42c6d ***
*** TT_GIT_VERSION 66ca1d5a6b56 ***
*** SDK_VERSION zephyr sdk 0.17.4 ***
Tenstorrent Blackhole CMFW 0.23.0
```

While `66ca1d5a6b56` corresponds to the 19.1.0 tag SHA, this changes the actual string data in the image. So to truly reproduce the 19.1.0 build, you first have to remove all tags from your repository.

.github/workflows/build-fw.yml
```diff
@@ -50,10 +50,18 @@ jobs:
       - uses: actions/checkout@v4
         with:
           path: tt-zephyr-platforms
+          fetch-depth: 0
```
cfriedt: Rather than fetching all of git history, and then having a separate step for fetching tags, why not just fetch the needed commit with tags?

The suggestion just jumped out at me after looking here.
https://github.com/actions/checkout

```suggestion
          fetch-tags: true
```

.github/workflows/build-fw.yml
```diff
@@ -50,10 +50,18 @@ jobs:
...
         with:
           app-path: tt-zephyr-platforms
 
+      # Fetch tags from repositories in manifest. Needed so that
+      # git describe functions correctly when creating a build.
+      - name: Fetch Tags
+        shell: bash
+        run: |
+          west forall -c 'git fetch --unshallow || true'
+
```
cfriedt:
```suggestion
```

.github/workflows/build-fw.yml
danieldegrasse: Didn't know about this, applied

.github/workflows/build-fw.yml
cfriedt: yeah, its just a removal because the tags are already fetched

danieldegrasse: Where? The checkout is going to fetch tags for `tt-zephyr-platforms`, but we also need to fetch tags from Zephyr to get the Zephyr version printed during boot. The zephyr setup action performs a shallow clone by default, which is why I have this step

danieldegrasse:
Rekicked CI as the P150a runner was having issues running the recovery script

PR #822, opened 2025-11-05
msgqueue: add l2cpu_enable flag for power setting command

alayander:
Adds flag to enable/disable L2CPU cores through `power_setting_rqst`.

Modifies msgqueue test to use this flag as well as introduces an emulated clock_control driver needed for msgqueue test run on native sim.

### DIFF
--- a/tests/lib/tenstorrent/bh_arc/src/msgqueue.c
+++ b/tests/lib/tenstorrent/bh_arc/src/msgqueue.c
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <zephyr/drivers/clock_control.h>
+#include <zephyr/drivers/clock_control/clock_control_tt_bh.h>
 #include <zephyr/ztest.h>
 
 #include <tenstorrent/smc_msg.h>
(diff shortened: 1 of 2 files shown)


tests/lib/tenstorrent/bh_arc/src/msgqueue.c
```diff
@@ -97,20 +99,36 @@ ZTEST(msgqueue, test_msgqueue_register_handler)
...
+		      CLOCK_CONTROL_STATUS_OFF);
+	zassert_equal(clock_control_get_status(
+			      pll4, (clock_control_subsys_t)CLOCK_CONTROL_TT_BH_CLOCK_L2CPUCLK_1),
+		      CLOCK_CONTROL_STATUS_OFF);
+	zassert_equal(clock_control_get_status(
+			      pll4, (clock_control_subsys_t)CLOCK_CONTROL_TT_BH_CLOCK_L2CPUCLK_2),
+		      CLOCK_CONTROL_STATUS_OFF);
+	zassert_equal(clock_control_get_status(
+			      pll4, (clock_control_subsys_t)CLOCK_CONTROL_TT_BH_CLOCK_L2CPUCLK_3),
+		      CLOCK_CONTROL_STATUS_OFF);
```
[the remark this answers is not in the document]
jgrowdenTT: I don't think it hurts to have it. it could be useful to debug some issues in case we blame host interface for not doing the right thing on the power cmd or something. 

If it actually does hurt to have it (it's non deterministic, or otherwise tough to upkeep becasue of *reasons*) I think we could scrap it. 

I guess the missing link here is on-device unit tests for our clock control block? That'd be cool. Lets say out of scope of this though.

Power cmd in general is lacking on-device testing but should be covered once we get the needed pyluwen interface for "full" messages.

jgrowdenTT (approved):
Approving, but I wrote a lot of this...

lib/tenstorrent/bh_arc/power.h
cfriedt: This file and the `set_l2cpu_enable()` function could def be namespaced better, and although `#include "power.h"` is done from within the same working directory, we do add `lib/tenstorrent/bh_arc` and `lib/tenstorrent/bh_arc/include` to the global search paths [here](https://github.com/tenstorrent/tt-zephyr-platforms/blob/35bec380edb2fc7d2623bf48d24ca07baa94e637/lib/tenstorrent/bh_arc/CMakeLists.txt#L4).

It might even make sense to add the function declaration to [tenstorrent/bh_arc.h](https://github.com/tenstorrent/tt-zephyr-platforms/blob/main/include/tenstorrent/bh_arc.h) and include it via `#include <tenstorrent/bh_arc.h>`. That way it's kind of implicitly associated with the bh_arc library.

Whether or not we want to namespace the function (e.g. `bh_arc_set_l2cpu_enable()`) might depend on existing conventions for our shell or other handlers.

What do you think @jgrowdenTT ?

cfriedt: Renaming the header should work as well.

Would it make sense to remove the "bh_arc" directory from the global include path? If we have tests or drivers that refer to it, they should be able to locally add the path.

cfriedt: > remove the "bh_arc" directory from the global include path?

That's here
https://github.com/tenstorrent/tt-zephyr-platforms/pull/841

PR #826, opened 2025-11-07
lib: tenstorrent: bh_arc: tensix_init.c: align ram buf to 64 bytes

alexapostolu:
Align RAM buffer to 64 bytes for the NOC2AXI to L1 transaction. This alignment is specified by the data sheet, and without it the data transferred may be shifted by some amount, resulting in garbage values in the L1.


alexapostolu:
@cfriedt Yup I checked, this is the only place

PR #828, opened 2025-11-08
manifest: bump zephyr version to v4.3.0-rc3

cfriedt:
Update to zephyr v4.3.0-rc3 and shed some patches that we no longer need since they are merged upstream already.


cfriedt:
I'm unable to locally reproduce the error that's causing CI to fail in the tests that appear to be passing. Locally, it's been going past 10 iterations successfully. I retried CI once already, so it seems it's not a random error.

```shell
IT=0; RES=0; while [ $RES -eq 0 ]; do IT=$((IT+1)); echo "Iteration $IT"; ./scripts/ci/run-smoke.sh -t smc -p 0 p100a &>/dev/null; RES=$?; done
Iteration 1
Iteration 2
Iteration 3
Iteration 4
Iteration 5
Iteration 6
Iteration 7
Iteration 8
Iteration 9
Iteration 10
```

cfriedt:
Indeed, increasing the buffer size for the test fixes the problem, although our CI is still seeming somewhat fragile due to synchronization issues between jobs (libusb errors, cards not detected on systems, etc).
https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/19211235158/job/54916302111?pr=829

cfriedt:
p300a runner seemed to be unavailable for 45 minutes, so I cancelled those jobs and re-ran.

Issue #703, opened 2025-10-08
Please show an error when build-combined-fwbundle.sh fails


nhuang-tt:
**Describe the bug**
There's no error shown to the user when build-combined-fwbundle.sh  failss

**To Reproduce**
Run build-combined-fwbundle.sh with the disk full for example

**Expected behavior**
It should throw an error.

**Impact**
User is left confused on where the output files are.

**Logs and console output**
```
(.venv) nhuang@bh-db-03-special-nhuang-for-reservation-38995:~/tt-zephyr-platforms-work/tt-zephyr-platforms$ ./scripts/build-combined-fwbundle.sh  
Created TEMP_DIR
Building release 18.12.99 / pack 18.12.99.0
Building tt_blackhole@p100a/tt_blackhole/smc
Removed TEMP_DIR
```

**Environment (please complete the following information):**
N/A

**Additional context**
We just need to remove `>/dev/null 2>&1` from the build script

PR #831, opened 2025-11-10
Remove >/dev/null 2>&1 from build_combined_fwbundle.sh

nhuang-tt:
TIcket
https://github.com/tenstorrent/tt-zephyr-platforms/issues/703

build_combined_fwbundle.sh fails silently and doesn't show the error to the user.

Changes
- Remove >/dev/null 2>&1 from build_combined_fwbundle.sh

### DIFF
--- a/scripts/build-combined-fwbundle.sh
+++ b/scripts/build-combined-fwbundle.sh
@@ -62,8 +62,7 @@ for REV in $BOARD_REVS; do
   fi
 
   echo "Building $BOARD"
-  west build -d "${TEMP_DIR}${REV}" --sysbuild -p -b "$BOARD" app/smc \
-    >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


scripts/build-combined-fwbundle.sh
```diff
@@ -62,8 +62,7 @@ for REV in $BOARD_REVS; do
   fi
 
   echo "Building $BOARD"
-  west build -d "${TEMP_DIR}${REV}" --sysbuild -p -b "$BOARD" app/smc \
-    >/dev/null 2>&1
+  west build -d "${TEMP_DIR}${REV}" --sysbuild -p -b "$BOARD" app/smc
```
cfriedt: It may be better if standard error and standard output were redirected to a file, and in the case of an error, a trap / signal handler were to print the contents of that file.

The redirections were mainly intended to reduce verbosity, so I would like to preserve that if possible.

PR #834, opened 2025-11-12
drivers: remove unused pincontrol driver

alexapostolu:
~~Remove RX and TRI enable from pin controller, as these should be the GPIO driver's responsibility.~~

~~I'll wait until the I2C and UART drivers are Zephyrized first~~

Remove the pin control driver entirely as it's not used


slindsayTT (commented):
Seems pretty straightforward. Are there any affected tests?

alexapostolu:
@slindsayTT I think the issue was some unrelated github issue, tests are green now

PR #848, opened 2025-11-18
ci: metal: run tests that include models

danieldegrasse:
Run test sets including models on blackhole cards as part of metal
nightly tests.

### DIFF
--- a/.github/workflows/metal.yml
+++ b/.github/workflows/metal.yml
@@ -2,6 +2,9 @@ name: Metal 🤘
 
 # Run long tests once nightly, at 00:00
 on:
+  pull_request:
+    types:
+      - labeled
   schedule:
     - cron: "0 0 * * *"
   workflow_dispatch: {}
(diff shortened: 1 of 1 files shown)


danieldegrasse:
After one restart, metal CI passes here: https://github.com/tenstorrent/tt-zephyr-platforms/actions/runs/20350512354/job/58480862327?pr=848

the initial failure on the P100a was `prefill_time_to_token (0.04744458198547363) is much lower than expected 0.06. Please update the expected perf.`. Discussing with the metal infra team, it seems like this may be a case where the threshold for the model simply needs to be adjusted to accept this performance (which is better than expected)

Personally, I think this change is ok to go in

However, it is **worth nothing** that currently to run P300 specific tests (which include running llama-3B-instruct) I had to download the llama-3B-instruct model weights locally on the P300a runners. This isn't a sustainable solution once we move to ephemeral CI, but will work in the short term.

danieldegrasse:
Had to rekick metal test on P100a once here, but the original error appears to be the same performance delta issue, which I think is not related to our firmware

alexapostolu:
@danieldegrasse Should we wait until we can analyze the performance better before merging this PR? Because the perf failures seem to be happening more frequently compared to other CI failures.

danieldegrasse:
> @danieldegrasse Should we wait until we can analyze the performance better before merging this PR? Because the perf failures seem to be happening more frequently compared to other CI failures.

Yes, lets. I'd love to get the tests reliable before we merge

alexapostolu (approved):
Looks good!

PR #855, opened 2025-11-23
drivers: dma: dma_tt_bh_noc: refactor driver

alexapostolu:
* Remove `resume()` and `release()` because our hardware does not support those and is not needed to pass upstream tests.
* Added additional error checking to match with upstream API expectations, and basic code refactoring.
* Added the node to `native_sim` overlay file, allowing us to use `DEVICE_DT_GET` instead of `DEVICE_DT_GET_OR_NULL`.


cfriedt (approved):
I'm surprised that non-`_OR_NULL` variants of DT macros can be used when we build native sim tests for bh_arc, but LGTM

PR #883, opened 2025-12-04
lib:tenstorrent:bh_arc: Set Recovery FW to Single SERDES

waynengTT:
Some p300 cards were not enumerating with recovery/preflash firmware due to both SERDES being configured and second SERDES failing.
In production firmware, only one SERDES is configured and the 2 SERDES configuration is redundant. This change also fixes the failing p300 cases.

### DIFF
--- a/lib/tenstorrent/bh_arc/pcie.c
+++ b/lib/tenstorrent/bh_arc/pcie.c
@@ -469,14 +469,14 @@ static int pcie_init(void)
 	if (IS_ENABLED(CONFIG_TT_SMC_RECOVERY)) {
 		pci0_property_table = (FwTable_PciPropertyTable){
 			.pcie_mode = FwTable_PciPropertyTable_PcieMode_EP,
-			.num_serdes = 2,
+			.num_serdes = 1,
 			.pcie_bar0_size = PCIE_BAR0_SIZE_DEFAULT_MB,
 			.pcie_bar2_size = PCIE_BAR2_SIZE_DEFAULT_MB,
 			.pcie_bar4_size = PCIE_BAR4_SIZE_DEFAULT_MB,
 		};
 		pci1_property_table = (FwTable_PciPropertyTable){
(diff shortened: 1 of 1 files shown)


waynengTT:
Yes, I have tested the recovery firmware with this change locally on a p300 and on a p150 and they were able to boot up without any issues.

lib/tenstorrent/bh_arc/pcie.c
```diff
@@ -468,14 +468,14 @@ static int pcie_init(void)
 	if (IS_ENABLED(CONFIG_TT_SMC_RECOVERY)) {
 		pci0_property_table = (FwTable_PciPropertyTable){
 			.pcie_mode = FwTable_PciPropertyTable_PcieMode_EP,
-			.num_serdes = 2,
+			.num_serdes = 1,
```
danieldegrasse: Regarding the change itself, I guess the issue is P300 specific, because there we need to use a more narrow PCIe lane width? For P300 the main issue would be that we setup both PCIe1 and PCIe0, right? In production firmware each ASIC only sets up one PCIe instance

slindsayTT:
Do we want to proceed with this PR or just ditch it?

afongTT:
Noting that we need this change for recovery FW, as seen in [SYS-3893](https://tenstorrent.atlassian.net/browse/SYS-3893).

afongTT (approved):
This change compiled into the preflash image works on a qb2 p300 system I was testing for [SYS-3829](https://tenstorrent.atlassian.net/browse/SYS-3829).

waynengTT:
Would the way this is implemented affect p150 since it uses 2 SERDES?

afongTT:
> Would the way this is implemented affect p150 since it uses 2 SERDES?

Yes it would affect p150 recovery image, but we have 1 recovery image that we expect to work on all platforms, so I think it's ok.

PR #943, opened 2026-01-12
scripts: tt_boot_fs: support multiple descriptor table definitions

danieldegrasse:
# Overview

Add support for multiple descriptor table definitions within tt_boot_fs. This enables us to define an additional "mutable" descriptor table, chosen to reside at `0x17000`. This table will change with each firmware revision, but the ROM table at `0x0` should no longer change. This will allow us to preserve a bootable change of `rom -> mcuboot -> recovery image` across all firmware updates, providing more reliable boot fallback.

> [!NOTE]
> the commits in this PR **are not** bisect-able. To keep everything bisectable, all the code changes would need to be > moved into one commit, as they are not backwards compatible. I can do this if before we merge. For now, I kept commits smaller to aid inreview.
 
## Technical details

This PR adds a "header table" at `0x120000` (offset was chosen because no firmware data lives at this address in current binaries or in pre 19.0 binary layouts)

The header table includes a magic number, version, and bootfs descriptor table count. Most boot images will have 3 descriptor tables:

- ROM table at 0x0 (fixed across updates)
- Failover table at 0x4000 (fixed across updates)
- Mutable table at 0x170000 (variable with updates)

Table offsets in flash are stored following the bootfs header itself as an array of 32 bit unsigned integers.


scripts/tt_boot_fs.py
```diff
@@ -577,11 +652,40 @@ def from_binary(data: bytes, alignment: int = 0x1000) -> BootFs:
...
-            entries[tag] = BootFs.check_entry(tag, fds[tag], data, alignment)
-        failover = BootFs.check_entry("failover", failover_fd, data, alignment)
+        tables = []
+
+        for offset in table_addrs:
+            order: list[str] = []
+            fds: dict[str, tt_boot_fs_fd] = {}
+            entries: dict[str, FsEntry] = {}
+            # Scan FDs in this table
+            for value in iter_fd(
```
alewycky-tenstorrent: Destructure the tuples returned by iter_fd.

scripts/tt_boot_fs.py
```diff
@@ -621,20 +724,25 @@ def load(path: str, env: dict):
...
         alignment = FileAlignment.loads(data["alignment"])
-        images = {}
-        for ent in data["images"]:
-            ent_name = ent["name"]
-            if ent_name in images:
-                raise ValueError(f"Found duplicate image name '{ent_name}'")
-            images[ent_name] = BootImage.loads(ent_name, ent, alignment, env)
+        tables = []
+        for t in data["tables"]:
+            table_ent = {"header_addr": t["header_addr"], "images": {}}
```
alewycky-tenstorrent: You don't really need table_ent until you go to append to tables. Building the images dict makes the code more like the original.

scripts/tt_boot_fs.py
```diff
@@ -577,11 +652,40 @@ def from_binary(data: bytes, alignment: int = 0x1000) -> BootFs:
...
+                lambda addr, size: data[addr + offset : addr + offset + size]
+            ):
+                tag = value[1].image_tag_str()
+                fds[tag] = value[1]
+                order.append(tag)
+
+            for tag in order:
+                entries[tag] = BootTable.check_entry(tag, fds[tag], data, alignment)
+
+            if offset == FAILOVER_HEAD_ADDR:
```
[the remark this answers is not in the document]
danieldegrasse: Yeah, older firmware images don't have a tag on the "failover" entry, since the ROM doesn't use the tag for anything

PR #964, opened 2026-01-16
ci: add recovery image test

alayander:
Adds a e2e test to validates booting into recovery CMFW.

### DIFF
--- a/app/smc/pytest/recovery.py
+++ b/app/smc/pytest/recovery.py
@@ -7,17 +7,18 @@ import logging
 import pyluwen
 import sys
 import time
-
-from pathlib import Path
-from twister_harness import DeviceAdapter
-
-# Import tt_boot_fs utilities
-sys.path.append(str(Path(__file__).parents[3] / "scripts"))
+import yaml
 
 import tt_boot_fs
 import tt_fwbundle
 
+from intelhex import IntelHex
+from pathlib import Path
 from pcie_utils import rescan_pcie
+from e2e_smoke import DeviceAdapter
+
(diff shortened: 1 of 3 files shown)


alayander:
Previous p300 failures were caused by number of SERDES used for PCIe for recovery image, which was fixed in main

afongTT:
Hmmmm is this test doing something different from `test_mcuboot`:

https://github.com/tenstorrent/tt-system-firmware/blob/b4f8ea2c3d1ae5257ffd14a73352501e05ede9ee/app/smc/pytest/e2e_smoke.py#L1061

alayander:
> Hmmmm is this test doing something different from `test_mcuboot`:
> 
> https://github.com/tenstorrent/tt-system-firmware/blob/b4f8ea2c3d1ae5257ffd14a73352501e05ede9ee/app/smc/pytest/e2e_smoke.py#L1061

I think the only difference is the "corruption path", where this tests a bad image flashed from a bundle. Otherwise I think both are testing similar functionality - if this seems unnecessary I can close this PR.

alayander:
Actually are e2e smoke tests run on the second asic for p300? If not that would explain why this test was failing for p300 prior to the SERDES recovery image fix, while `test_mcuboot` was still passing on p300.

Based on this it seems like we only run on one ASIC, unless I'm missing something:
https://github.com/tenstorrent/tt-system-firmware/blob/b4f8ea2c3d1ae5257ffd14a73352501e05ede9ee/.github/workflows/hardware-smoke.yml#L175-L195

PR #984, opened 2026-02-03
[DNM] ci: add initial gitlab CI file

danieldegrasse:
For now this PR will be a testbed where additional commits are pushed (if gitlab CI triggers on this PR at all, which is not a given)

This PR needs to be created from tt-zephyr-platforms instead of a fork due to restrictions on how gitlab CI triggers PR jobs


danieldegrasse:
No longer needed

PR #1038, opened 2026-02-25
.github: workflows: hardware-long: add manufacturing test job

ShLiTT:
Add a manufacturing-test job to the HW Soak workflow that validates the full manufacturing flash and boot sequence on Blackhole boards:
      1. Erase SPI and flash preflash ihex
      2. Flash assembly test firmware (MCUBoot + app) to DMC via STM32Programmer
      3. Reset DMC and verify PCIe enumeration with assembly test FW
      4. Flash production firmware via tt-flash
      5. Reset DMC and verify PCIe enumeration with production FW

### DIFF
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -393,6 +393,11 @@ jobs:
         shell: bash
         run: |
           ./build-flm.sh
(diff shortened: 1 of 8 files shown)


afongTT:
Continued in #1170.

PR #1103, opened 2026-03-19
app: dmc: add dmc_command_code handler

danieldegrasse:
Add a dmc_command_code handler, which supports parsing command codes
from the host. Commands are encoded within a uint32_t due to limitations
of the SMC <-> DMC interface.

These commands are primarily intended as a debug interface to issue
requests to the DMC when an ASIC is in an unknown state.

### DIFF
--- a/app/dmc/src/main.c
+++ b/app/dmc/src/main.c
@@ -34,6 +34,7 @@
 #include <tenstorrent/tt_smbus_regs.h>
 
 #define RESET_UNIT_ARC_PC_CORE_0 0x80030C00
+#define RESET_UNIT_SCRATCH_0 0x80030060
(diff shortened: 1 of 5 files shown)


jgrowdenTT:
Smoke tests? Debug or no, might be good to know they work so we know we're leaning on tried and tested tooling.

jgrowdenTT:
I think it's more challenging, but not overly so, and If it helps us in the future it might be worth adding. I can maybe try to sketch something up if we think it helps.

danieldegrasse:
Closing this out- if you need the code in this PR, please just let me know. No plans to get rid of my fork, so it should still be there :)

PR #1163, opened 2026-03-31
resolves: SYS-3958 Add host-controllable minimum frequency floor for AICLK

psethiTT:
Implement TT_SMC_MSG_SET_ASIC_HOST_FMIN to allow the host to set a minimum frequency floor that takes precedence over the hardware fmin. This provides a mechanism for the host to prevent the AICLK from dropping below a specified threshold, similar to the existing host_fmax arbiter but for the minimum frequency.

Changes:
- Add set_asic_host_fmin_rqst message structure and handler
- Add host_requested_fmin field to AiclkPPM state
- Update CalculateTargAiclk to enforce host-requested minimum
- Expand AICLK_FMIN_MAX from 800 to 1400 MHz to allow full range
- Support restore_default flag to disable the host-requested floor

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -97,8 +97,12 @@ TT_SMC_MSG_READ_PD = 0x1C
 TT_SMC_MSG_READ_VM = 0x1D
 TT_SMC_MSG_SET_TDP_LIMIT = 0x22
 TT_SMC_MSG_SET_ASIC_HOST_FMAX = 0x23
+TT_SMC_MSG_CHARACTERISATION = 0xC6
 TT_SMC_MSG_COUNTER = 0x35
 TT_SMC_MSG_TOGGLE_GDDR_RESET = 0xB6
+
+# Characterization submessage IDs
+TT_SUB_MSG_SET_HOST_REQUESTED_FMIN = 0x1
 
 # Telemetry tags
 TAG_TDP = 7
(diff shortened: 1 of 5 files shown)


psethiTT:
@afongTT @jgrowdenTT @tcolacoTT if you could please look at the following merge request. 

One concern I have is the change I have made to Fmin range by increasing by max from 800 to 1400. This is then used in set_arb_host_fmin_handler as a check. However, if there are concerns to not touch the Fmin range, we can enforce them on user side while sending python message.

jgrowdenTT:
We're going to talk about this in our next FW meeting. We're discussing a few options:

- Fw build for characterization
- fork
- feature flag in flash

Until then, hold tight. Thanks for your patience <3

jgrowdenTT (changes requested):
As well as the other requested changes, please add a smoke test to e2e for this. Thanks Parth!

app/smc/pytest/e2e_smoke.py
```diff
@@ -1403,6 +1407,104 @@ def test_set_asic_host_fmax(arc_chip_dut, asic_id):
...
+    Verifies that the request succeeds.
+    """
+    arc_chip = pyluwen.detect_chips()[asic_id]
+
+    NEW_HOST_FMIN = 600  # MHz
+
+    # Positive case: set a valid host fmin
+    response = arc_chip.as_bh().arc_msg_buf(
+        [
+            TT_SMC_MSG_CHARACTERISATION,
```
jgrowdenTT: This doesn't look right. each of these entries is on u32 word. The way you've packed the bytes in the command, the REQUESTED_FMIN should be in bits 8-15 of word 0. 

NEW_HOST_FMIN should be in word 1.

Same for below.

psethiTT: thank you, yes I fixed that!

jgrowdenTT (approved):

PR #1170, opened 2026-04-02
.github: workflows: hardware-long: add manufacturing test

alayander:
Continuation of #1038 

Adds a manufacturing-test job to the HW Soak workflow that validates the full manufacturing flash and boot sequence on Blackhole boards.

tests:
- `run-manufacturing-test.sh` passes when run locally on p150a (with artifact assets downloaded)
- Passes in CI on p100a, p150a and p300a on when run as part of long-tests

resolves: SYS-2854

### DIFF
--- a/.github/workflows/build-fw.yml
+++ b/.github/workflows/build-fw.yml
@@ -325,6 +325,9 @@ jobs:
 
   build-preflash:
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        board_rev: [p150a, p300a]
     steps:
       - uses: actions/checkout@v4
         with:
@@ -336,12 +339,12 @@ jobs:
       - name: Build preflash firmware
         shell: bash
         run: |
-          ./tt-system-firmware/scripts/build-preflash.sh
+          ./tt-system-firmware/scripts/build-preflash.sh ${{ matrix.board_rev }}
 
       - name: Upload preflash
         uses: actions/upload-artifact@v4
         with:
-          name: preflash-fw
+          name: preflash-fw-${{ matrix.board_rev }}
           path: |
             preflash*.ihex
             preflash*.bin
@@ -394,6 +397,11 @@ jobs:
         shell: bash
         run: |
           ./build-flm.sh
+      - name: Upload Flash Loader Module
+        uses: actions/upload-artifact@v4
+        with:
+          name: bh-flm
+          path: tt-system-firmware/scripts/tooling/blackhole_recovery/data/bh_flm/build/*.flm
 
   discover-smc-tests:
     runs-on: ubuntu-24.04
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -35,11 +35,15 @@ jobs:
     runs-on: ubuntu-24.04
     outputs:
       matrix-config: ${{ steps.gen.outputs.matrix-config }}
+      mfg-matrix-config: ${{ steps.gen.outputs.mfg-matrix-config }}
     steps:
       - uses: actions/checkout@v4
       - id: gen
         run: |
           echo "matrix-config=$(jq -c '.bh' .github/ci_boards.json)" | tee -a $GITHUB_OUTPUT
+          # Manufacturing test only runs on the standalone PCIe cards (p100a, p150a, p300a)
+          MFG_MATRIX_CONFIG="$(jq -c '[.bh[] | select(.board == "p100a" or .board == "p150a" or .board == "p300a")]' .github/ci_boards.json)"
+          echo "mfg-matrix-config=$MFG_MATRIX_CONFIG" | tee -a $GITHUB_OUTPUT
 
   e2e-stress-test:
     name: BH E2E Stress Test on ${{ matrix.config.board }}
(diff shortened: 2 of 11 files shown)


scripts/ci/run-manufacturing-test.sh
```diff
@@ -0,0 +1,167 @@
...
+
+# ---- Step 1: Erase SPI + flash preflash ihex ----
+echo "=== Step 1: Erase SPI and flash preflash ihex ==="
+PREFLASH_IHEX=$(ls "$PREFLASH_DIR"/preflash-*.ihex | head -1)
+echo "Using preflash image: $PREFLASH_IHEX"
+python3 "$TT_Z_P_ROOT"/scripts/spi_flash.py \
+	--board-name "$BOARD" --no-prompt -v full_erase
+python3 "$TT_Z_P_ROOT"/scripts/spi_flash.py \
+	--board-name "$BOARD" --no-prompt -v \
+	write_from_ihex "$PREFLASH_IHEX"
```
afongTT: I just noticed that this isn't flashing preflash on both ASICs of p300.

alayander: The `spi_flash.py` script handles both ASICs based on the passed board-name and  `boardname_metadata.yaml`

scripts/ci/run-manufacturing-test.sh
```diff
@@ -0,0 +1,173 @@
...
+
+# ---- Step 5: Flash production FW to SPI via pyocd (STM32 SWD) ----
+echo "=== Step 5: Flash production firmware ==="
+FWBUNDLE=$(ls "$FWBUNDLE_DIR"/fw_pack*.fwbundle | head -1)
+echo "Using firmware bundle: $FWBUNDLE"
+python3 "$TT_Z_P_ROOT"/scripts/spi_flash.py \
+	--board-name "$BOARD" --no-prompt -v full_erase
+python3 "$TT_Z_P_ROOT"/scripts/spi_flash.py \
+	--board-name "$BOARD" --no-prompt -v \
+	write_from_fwbundle "$FWBUNDLE"
```
afongTT: It would be better if you could do this over PCIe (by sending SMC msgs to the preflash FW).

Then you also shouldn't need to do Step 6, and DMC should upgrade to the version from the FW Bundle.

alayander:
Cherry-picked commit from: https://github.com/tenstorrent/tt-system-firmware/pull/883

scripts/smc_spi_flash.py
```diff
@@ -0,0 +1,207 @@
...
+        raise FileNotFoundError(fwbundle_path)
+
+    with tempfile.TemporaryDirectory() as tmp:
+        with tarfile.open(fwbundle_path, "r") as tar:
+            _extract_tar_safe(tar, tmp)
+        root = Path(tmp)
+
+        for idx, asic in asics:
+            name = asic["name"]
+            image_bin = root / name / "image.bin"
```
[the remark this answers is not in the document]
alayander: @afongTT is a fix here necessary? The `root` referenced in the code is a temporary directory constructed in the lines above.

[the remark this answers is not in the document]
afongTT: This seems like there isn't actually an issue since `root` and `name` are effectively sanitized.

You might as well be defensive and add the check cycode suggested (that the final path `.startswith(tmp)`.

afongTT (dismissed):

alayander:
~~Will split this into multiple smaller PRs.~~ Actually, was able to significantly refactor and clean up changes - so reopening this PR

.github/workflows/build-fw.yml
```diff
@@ -325,6 +325,9 @@ jobs:
 
   build-preflash:
     runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        board_rev: [p150a, p300a]
```
jgrowdenTT, line 330: p100?

jgrowdenTT:
We need docs on how we can run these tests locally,[ added here.](https://docs.tenstorrent.com/tt-system-firmware/develop/testing/index.html)

scripts/check_card.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
```
jgrowdenTT, line 1: Don't we have somthing like this?

.github/workflows/hardware-long.yml
```diff
@@ -171,3 +175,100 @@ jobs:
...
+    needs: [build-fw-artifact, gen-config]
+    strategy:
+      fail-fast: false
+      matrix:
+        config: ${{ fromJson(needs.gen-config.outputs.mfg-matrix-config) }}
+    runs-on: ${{ matrix.config.board }}
+    container:
+      image: ghcr.io/tenstorrent/tt-system-firmware/ci-image:v19.11.0
+      volumes:
+        - /dev/hugepages-1G:/dev/hugepages-1G
```
jgrowdenTT, line 189: They need hugepages?

jgrowdenTT (approved):
I think if this passes, lets not delay.

But I think we can have improvements in some arears, liek getting more out of tt-flash and carrying less in-repo flashling logic in smc_spi_flash.py

afongTT (approved):

PR #1183, opened 2026-04-14
manufacturing_led: fix board fault led behaviour

alexapostolu:
Manufacturing test requires the board fault LED to be solid on. The commit that introduced a SMC message to blink the LED, accidentally overrode this behavior causing the board fault LED to blink instead of staying solid.

Fix this bug so that the board fault LED is solid on for manufacturing, but for non manufacturing builds still allow the SMC to blink.

Also remove the "led0" node from devicetree as it's the same as the board fault led.

resolves: SYS-3844

### DIFF
--- a/app/dmc/src/main.c
+++ b/app/dmc/src/main.c
@@ -42,8 +42,6 @@ LOG_MODULE_REGISTER(main, CONFIG_TT_APP_LOG_LEVEL);
 
 BUILD_ASSERT(FIXED_PARTITION_EXISTS(bmfw), "bmfw fixed-partition does not exist");
 
-static bool blink_led;
-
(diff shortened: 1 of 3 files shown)


jgrowdenTT (dismissed):

jgrowdenTT:
Just confirming:

A) You spot checked this on the recovery FW to make sure LED is in the right state
B) You verified the blink msg still works via spot check

Also could you please:

A) update release notes
B) Add to the PR description `resolves: SYS-xxxx`

Thanks Alex!

alexapostolu:
@jgrowdenTT Yup I tested it visually and I added release notes

jgrowdenTT (approved):

PR #1189, opened 2026-04-15
jtag_bootrom: trigger memory repair from dmc

afongTT:
As a workaround to issues seen in SYS-2903 (where bootcode hangs on postcode 0x13), use JTAG to trigger memory repair from the DMC every time it resets the ASIC.

Timing of this change on p300c shows ~4.5 ms increase in boot time on the worse chip (chip0, which gets booted second):
- 7 ms **increase** in PERST to bootrom start on chip0
- 2.5 ms **decrease** in bootrom start to PCIe init complete on chip0

This roughly makes sense, as DMC logging is showing memory repair takes about 3.5 ms. We could probably take about 3 ms off this impact if we make bigger changes to the code to kick off memory repair on both chips in parallel.

For reference, on one good p300c, the boot time for chip 0 goes from 50.5 ms to 55.2 ms.

resolves: SYS-2903, SYS-3829


lib/tenstorrent/jtag_bootrom/jtag_bootrom.c
```diff
@@ -84,6 +83,39 @@ void gpio_asic_reset_callback(const struct device *port, struct gpio_callback *c
 static struct gpio_callback preset_cb_data;
 #endif /* IS_ENABLED(CONFIG_JTAG_LOAD_ON_PRESET) */
 
+int jtag_trigger_mem_repair(struct bh_chip *chip)
+{
```
jgrowdenTT, line 87: might want `UNUSED(chip);`

jgrowdenTT (dismissed):

lib/tenstorrent/jtag_bootrom/jtag_bootrom.c
afongTT: Added `ARG_UNUSED` for when `CONFIG_JTAG_LOAD_BOOTROM` is not defined

jgrowdenTT (approved):

PR #1192, opened 2026-04-16
lib: bh_arc: Set serdes per board type

jgrowdenTT:
In recovery FW, use two serdes for p100/p150. Use one serdes for galaxy/p300.


afongTT:
Ok, if we have to do it.

Can you change the board type in the `build-preflash.sh` to p300a then?

afongTT (approved):

PR #1204, opened 2026-04-27
lib: bh_arc: pcie gen1 speed in low power mode

jgrowdenTT:
Add power flag to allow kmd to make pcie speed 1 in low power mode, and 5 in high power mode.

tests: smoke on a p150a. In tt-smi, pcie gen 5 when using UMD backend; pcie gen 1 when using luwen backend
<ticket link TBD>


afongTT (approved):

jgrowdenTT:
Failed dirty reset in smoke :/ I don't think I've seen that before on baseline. 

Might be wise to let this slip the release. I'll try to stress offline.

projkovTT:
Closed with move to follow up PR rebasing to latest main, retested: https://github.com/tenstorrent/tt-system-firmware/pull/1371

PR #1205, opened 2026-04-28
lib: bh_arc: pcie fwtable preset support added

waynengTT:
Added gen3_eq_control_off.f.gen3_eq_pset_req_vec and gen3_eq_control_off.f.gen3_eq_fb_mode as a FW table parameter that can be overwritten by SPI. These vectors apply to Gen3-Gen5.

This allows the preset vector to be configured with the SPI FW table instead of a full firmware flash if the preset values need to be changed.


afongTT:
Need internal MR to merge first.

See [SYS-2216](https://tenstorrent.atlassian.net/browse/SYS-2216)

jgrowdenTT (changes requested):
@waynengTT Discussed with @afongTT I think we need a new libpcie drop, and then appreciate if you could rebase this PR to resolve the conflicts. thanks

jgrowdenTT:
@waynengTT sorry, looks like you need to rebase before we can merge this. Thank you.

PR #1208, opened 2026-04-30
lib: bh_arc: KMD log

jgrowdenTT:
note: This PR does not turn on, or compile in, the KMD log by default. 

This is mostly to get early eyes on this @afongTT @slindsayTT @joelsmithTT @alewycky-tenstorrent 

This adds support for a "KMD PCIE" logging backend via Zephyr's logging infrastructure. KMD sends a message containing the pointer to the host allocated memory buffer to fill logs in. FW initializes header bytes in the buffer. 

FW logs to internal buffer on log push (LOG_INF, LOG_ERR, etc.) logs are framed with metadata at this point.

On timer, or as we approach a full buffer, we  flush that buffer out to the host allocated buffer. We mark the buffer as host owned and interrupt it to process.

(host implementation is something like: loop through all the logs up to the amount the fw indicated, print them, once done mark the buffer as FW owned.) 

Notes:

Zephyr supports multiple logging backends running at once, so we don't lose tt-console logs unless we specifically compile them out.

Tests: (July 10th, 2026)

On a KMD with support for the new logging:

1. smoke
2. metal 3X upstream container tests
3. Verified err/warn/info logs ends up in the right severity in dmesg.

Issues:

1. I seem to occassionally drop the last log right after init. Host correctly resyncs on next push and notes the sequence mismatch
3. I have, on various iterations of this, seen crashes during metal. i'm not totally convinced i have resolved the issues, but I haven't seen them recently. I will continue trying to reproduce. Failure mode is such that a pcie_rescan.sh recovers the connection to the SMC. - _I have not seen this since the beginning of development._

Todo before this can go in IMO

1. Documentation (assuming this design is greenlit)
2. Stress (with supported driver)
3. Flushing logic needs cleanup
4. Galaxy testing

Future Todo:

1. logging sub cmd for enabling/disabling modules per log level.

### DIFF
--- a/doc/services/kmd_logging/index.rst
+++ b/doc/services/kmd_logging/index.rst
@@ -1,0 +1,167 @@
+.. _ttfw_kmd_logging:
+
+KMD Logging Mechanism
+=====================
+
+This page describes how firmware log records are forwarded to the host kernel
+module (KMD) through the ``tt_pcie_log`` backend.
+
+Overview
+--------
+
+The mechanism has two paths:
+
+1. Control path: KMD sends setup and release messages over the ARC message queue
+   (message ID ``TT_SMC_MSG_TT_PCIE_LOG`` / ``0xC7``).
+2. Data path: firmware writes framed log batches into a host DMA buffer and
+   raises a PCIe MSI interrupt so KMD can consume them.
+
+On firmware side, the implementation lives in
+``lib/tenstorrent/bh_arc/tt_pcie_log.c`` and is enabled by
+``CONFIG_TT_PCIE_LOG_BACKEND``.
+
+If the CONFIG is not enabled, FW logs will not be sent to host, but FW will enumerate and operate.
+Other logging backends may be used in conjunction with this backend.
+
+Handshake: Setup And Release
+----------------------------
+
+KMD uses sub-commands in ``tt_pcie_log_rqst``:
+
+``SETUP`` (subcmd ``1``): KMD allocates a coherent DMA buffer and sends DMA
+address (low/high 32-bit words) plus buffer size to firmware. Firmware
+validates size, configures NOC2AXI access, initializes the header, and starts
+periodic flush.
+
+``RELEASE`` (subcmd ``2``): KMD asks firmware to stop host logging. Firmware
+stops the flush timer and returns to local buffering only.
+
+If host is not set up, firmware still formats logs into a local staging buffer.
+Once setup succeeds, subsequent flushes transfer staged data to host memory.
+
+Shared Buffer Contract
+----------------------
+
+Firmware and KMD must agree on these packed headers:
+
+.. list-table:: Buffer Metadata (16 bytes)
+   :header-rows: 1
+   :widths: 20 12 12 56
+
+   * - Field
+     - Type
+     - Size
+     - Meaning
+   * - ``write_offset``
+     - ``uint32_t``
+     - 4
+     - End offset of valid payload bytes in the shared buffer.
+   * - ``buffer_size``
+     - ``uint32_t``
+     - 4
+     - Total host DMA buffer size in bytes.
+   * - ``magic``
+     - ``uint32_t``
+     - 4
+     - Constant ``0x544C4F47`` ("TLOG") for contract validation.
+   * - ``owner``
+     - ``uint8_t``
+     - 1
+     - Ownership byte: ``0`` = FW owns buffer, ``1`` = host owns buffer.
+   * - ``version``
+     - ``uint8_t``
+     - 1
+     - The version of the messaging protocol. Fw currently only supports Version 0.
+   * - ``reserved``
+     - ``uint8_t[2]``
+     - 2
+     - Reserved padding for future use.
+
+.. list-table:: Entry Header (12 bytes)
+   :header-rows: 1
+   :widths: 20 12 12 56
+
+   * - Field
+     - Type
+     - Size
+     - Meaning
+   * - ``msg_size``
+     - ``uint16_t``
+     - 2
+     - Total entry size (entry header + payload). Payload includes a trailing
+       ``\0`` byte.
+   * - ``log_level``
+     - ``uint8_t``
+     - 1
+     - Zephyr log level of the message.
+   * - ``source``
+     - ``uint8_t``
+     - 1
+     - Message source: ``0`` = SMC, ``1`` = DMC.
+   * - ``timestamp``
+     - ``uint32_t``
+     - 4
+     - Firmware timestamp captured for the log entry.
+   * - ``sequence``
+     - ``uint32_t``
+     - 4
+     - Monotonic sequence number used for gap detection.
+
+Payload bytes follow each entry header and are emitted by Zephyr's log output
+formatter in text mode. Firmware appends a trailing ``\0`` to each payload for
+host-side C-string consumers; this NUL byte is included in ``msg_size``.
+
+
+Data Flow
+---------
+
+1. Zephyr logging calls the backend ``process`` callback for each message.
+2. Firmware writes an entry header plus formatted text into a local framed
+   buffer.
+3. Flush occurs periodically (timer + work queue) or when near local capacity.
+4. During flush:
+   - firmware validates host ``magic``
+   - firmware checks ``owner`` is ``0`` (previous batch consumed)
+   - firmware copies bytes to host DMA buffer after the 16-byte buffer header
+   - firmware updates ``write_offset``
+   - firmware sets ``owner = 1`` (handoff to host)
+   - firmware triggers MSI so KMD can process promptly
+5. KMD interrupt handler schedules work, parses entries, prints them to kernel
+   log with level mapping, then clears ``owner`` back to ``0``.
+
+Current Behavior Notes
+----------------------
+
+- Firmware batches data and sends only when host is ready.
+- If host has not yet consumed the previous batch, firmware drops the pending
+  local batch for forward progress.
+- If local staged data exceeds available host payload space, firmware truncates
+  to fit.
+- Data is currently copied as a contiguous batch (not a true host-side circular
+  producer/consumer protocol yet).
+
+Configuration
+-------------
+
+Relevant Kconfig options:
+
+- ``CONFIG_TT_PCIE_LOG_BACKEND``: enable host logging backend.
+- ``CONFIG_TT_PCIE_LOG_BACKEND_BUFFER_SIZE``: local firmware staging size
+  (default ``4096`` bytes).
+
+This backend uses Zephyr text formatting. Other formatters are not supported.
+
+Minimum practical requirement for setup is host buffer size >=
+``sizeof(fw_log_buffer_header) + sizeof(fw_log_entry_header) + 1``.
+
+Troubleshooting
+---------------
+
+- Setup rejected by firmware: check host-provided buffer size and DMA address
+  validity.
+- No host logs appearing: ensure ``CONFIG_TT_PCIE_LOG_BACKEND=y`` in firmware
+  build and confirm KMD successfully sent setup message and enabled logging.
+- Gaps or dropped records: if host processing is slower than firmware
+  production, batches can be dropped by design. Reduce firmware log volume or
+  increase host consumption responsiveness.
+- Ensure the KMD version supports the feature
--- a/include/tenstorrent/smc_msg.h
+++ b/include/tenstorrent/smc_msg.h
@@ -165,6 +165,9 @@ enum tt_smc_msg {
 
 	/** @brief @ref characterisation_rqst "Generic characterization message" */
 	TT_SMC_MSG_CHARACTERISATION = 0xC6,
+
+	/** @brief @ref tt_pcie_log_rqst "FW logging setup/release request" */
+	TT_SMC_MSG_TT_PCIE_LOG = 0xC7,
 };
 
 /** @brief Enumeration of characterization submessage IDs */
--- a/lib/tenstorrent/bh_arc/CMakeLists.txt
+++ b/lib/tenstorrent/bh_arc/CMakeLists.txt
@@ -27,6 +27,7 @@ zephyr_library_sources(
   noc2axi.c
   noc_init.c
   pcie.c
+  pcie_msi.c
   pll.c
   post_code.c
   reset.c
@@ -54,6 +55,8 @@ if(CONFIG_BH_FWTABLE)
   zephyr_library_add_dependencies(nanopb_generated_headers)
 endif()
 
+zephyr_library_sources_ifdef(CONFIG_TT_PCIE_LOG_BACKEND tt_pcie_log.c)
+
 zephyr_linker_sources(DATA_SECTIONS iterables.ld)
 if(CONFIG_ARC)
   zephyr_library_link_libraries(${ZEPHYR_CURRENT_MODULE_DIR}/zephyr/blobs/tt_blackhole_libpciesd.a)
@@ -75,7 +78,6 @@ if(NOT CONFIG_TT_SMC_RECOVERY)
     gddr.c
     i2c_messages.c
     pcie_dma.c
-    pcie_msi.c
     pvt.c
     regulator.c
     regulator_config.c
--- a/lib/tenstorrent/bh_arc/Kconfig.tt_pcie_log
+++ b/lib/tenstorrent/bh_arc/Kconfig.tt_pcie_log
@@ -1,0 +1,27 @@
+# Copyright (c) 2026 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+if LOG
+
+config TT_PCIE_LOG_BACKEND
+	bool "Enable Tenstorrent Host tt_pcie_log backend"
+	depends on !LOG_MODE_IMMEDIATE
+	select LOG_OUTPUT
+	help
+	  Enable tt_pcie_log backend that outputs logs to shared host memory buffer
+	  via DMA for kernel module consumption. When host is not available,
+	  logs are stored in local buffer and transferred when host connects.
+
+if TT_PCIE_LOG_BACKEND
+
+config TT_PCIE_LOG_BACKEND_BUFFER_SIZE
+	int "Local log buffer size in bytes"
+	default 4096
+	help
+	  Size of local buffer to aggregate logs into. Logs from this buffer
+	  are flushed on a timer, or when approaching capacity, and only if the
+	  host has messaged FW the buffer is available.
+
+endif # TT_PCIE_LOG_BACKEND
+
+endif # LOG
--- a/lib/tenstorrent/bh_arc/pcie_msi.c
+++ b/lib/tenstorrent/bh_arc/pcie_msi.c
@@ -61,7 +61,7 @@ void SendPcieMsi(uint8_t pcie_inst, uint32_t vector_id)
 		msi_data += vector_id;
 
 		const uint8_t ring = 0;
-		const uint8_t tlb_num = 0;
+		const uint8_t tlb_num = 12;
 		const uint8_t x = pcie_inst == 0 ? PCIE_INST0_LOGICAL_X : PCIE_INST1_LOGICAL_X;
 		const uint8_t y = PCIE_LOGICAL_Y;
 
--- a/lib/tenstorrent/bh_arc/pcie_msi.c
+++ b/lib/tenstorrent/bh_arc/pcie_msi.c
@@ -42,7 +42,7 @@ uint32_t GetVectorsAllowed(uint32_t mult_msg_en)
 	return 1 << mult_msg_en;
 }
 
-void SendPcieMsi(uint8_t pcie_inst, uint32_t vector_id)
+void SendPcieMsi(uint8_t x, uint8_t y, uint32_t vector_id)
 {
 	BH_PCIE_DWC_PCIE_USP_PF0_MSI_CAP_HDL_PATH_E982B20F_PCI_MSI_CAP_ID_NEXT_CTRL_REG_reg_u
 		pci_msi_cap;
@@ -62,8 +62,6 @@ void SendPcieMsi(uint8_t pcie_inst, uint32_t vector_id)
 
 		const uint8_t ring = 0;
 		const uint8_t tlb_num = 12;
-		const uint8_t x = pcie_inst == 0 ? PCIE_INST0_LOGICAL_X : PCIE_INST1_LOGICAL_X;
-		const uint8_t y = PCIE_LOGICAL_Y;
 
 		NOC2AXITlbSetup(ring, tlb_num, x, y, msi_addr);
 		NOC2AXIWrite32(ring, tlb_num, msi_addr, msi_data);
@@ -89,7 +87,10 @@ static uint8_t send_pcie_msi_handler(const union request *request, struct respon
 	uint8_t pcie_inst = request->send_pci_msi.pcie_inst;
 	uint32_t vector_id = request->send_pci_msi.vector_id;
 
-	SendPcieMsi(pcie_inst, vector_id);
+	const uint8_t x = pcie_inst == 0 ? PCIE_INST0_LOGICAL_X : PCIE_INST1_LOGICAL_X;
+	const uint8_t y = PCIE_LOGICAL_Y;
+
+	SendPcieMsi(x, y, vector_id);
 	return 0;
 }
 
--- a/lib/tenstorrent/bh_arc/tt_pcie_log.c
+++ b/lib/tenstorrent/bh_arc/tt_pcie_log.c
@@ -1,0 +1,472 @@
+/*
+ * Copyright (c) 2026 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/logging/log_backend.h>
+#include <zephyr/logging/log_output.h>
+#include <zephyr/logging/log_ctrl.h>
+#include <zephyr/logging/log_msg.h>
+#include <string.h>
+#include <tenstorrent/smc_msg.h>
+#include <tenstorrent/msgqueue.h>
+#include <zephyr/logging/log_backend_std.h>
+
+#include "chip_info.h"
+#include "noc2axi.h"
+
+/* Forward declaration for PCIe MSI function */
+void SendPcieMsi(uint8_t pcie_inst, uint32_t vector_id);
+
+LOG_MODULE_REGISTER(fw_tt_pcie_log_backend, CONFIG_LOG_DEFAULT_LEVEL);
+
+/* tt_pcie_log sub-commands */
+#define TT_PCIE_LOG_SUBCMD_SETUP   1
+#define TT_PCIE_LOG_SUBCMD_RELEASE 2
+
+/* Log entry header structure (must match KMD definition) */
+struct fw_log_entry_header {
+	uint16_t msg_size;  /* Size of log message including header */
+	uint8_t log_level;  /* FW log level */
+	uint8_t source;     /* 0=SMC, 1=DMC */
+	uint32_t timestamp; /* FW timestamp in milliseconds */
+	uint32_t sequence;  /* Sequence number */
+} __packed;
+
+/* Buffer header at start of shared buffer (must match KMD definition) */
+struct fw_log_buffer_header {
+	uint32_t write_offset; /* Current write position */
+	uint32_t buffer_size;  /* Total buffer size */
+	uint32_t magic;        /* Magic number for validation */
+	uint8_t owner;         /* Buffer owner: FW or host */
+	uint8_t version;       /* FW max supported version */
+	uint8_t reserved[2];   /* Future use */
+} __packed;
+
+#define FW_LOG_SOURCE_SMC 0
+#define FW_LOG_SOURCE_DMC 1
+
+#define FW_LOG_BUFFER_MAGIC      0x544C4F47 /* "TLOG" */
+#define FW_LOG_BUFFER_OWNER_FW   0x0        /* Firmware owns buffer, host has consumed */
+#define FW_LOG_BUFFER_OWNER_HOST 0x1        /* Host owns buffer, firmware has produced data */
+#define FLUSH_THRESHOLD_BYTES    (sizeof(struct fw_log_entry_header) + 256)
+
+/* Static variables to track tt_pcie_log state */
+static bool host_available; /* True when host buffer is set up */
+static uint64_t log_buffer_iova;
+static uint32_t log_buffer_size;
+static uint32_t sequence_number;
+
+#define TT_PCIE_LOG_TLB 11
+
+/* NOC2AXI TLB configuration for direct host memory access */
+#define PCIE_X 19
+#define PCIE_Y 24
+
+static void flush_buffer_to_host(void);
+
+/* Buffer for accumulating framed log entries before host transmission */
+static uint8_t framed_log_buffer[CONFIG_TT_PCIE_LOG_BACKEND_BUFFER_SIZE];
+static uint32_t framed_buffer_pos; /* Current write position in framed buffer */
+
+/* Timer for periodic buffer flushing */
+static struct k_timer flush_timer;
+
+/* Work item for buffer flushing */
+static struct k_work flush_work;
+
+/* Mutex to protect framed_log_buffer access between backend and timer */
+static struct k_mutex buffer_mutex;
+
+/**
+ * @brief Work handler to flush framed buffer to host
+ * @param work Work item (unused)
+ */
+static void flush_work_handler(struct k_work *work)
+{
+	ARG_UNUSED(work);
+	/* Lock mutex to prevent collision with backend_process */
+	if (k_mutex_lock(&buffer_mutex, K_MSEC(100)) != 0) {
+		/* Timeout - skip this flush cycle */
+		return;
+	}
+
+	flush_buffer_to_host();
+
+	k_mutex_unlock(&buffer_mutex);
+}
+
+/**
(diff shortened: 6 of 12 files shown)


lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,409 @@
...
+	/* Create header and reserve space for it */
+	entry_header.msg_size = 0; /* Will be updated after formatting */
+	entry_header.log_level = log_msg_get_level(&msg->log);
+	entry_header.source = FW_LOG_SOURCE_SMC;
+	entry_header.timestamp = msg->log.hdr.timestamp;
+	entry_header.sequence = sequence_number++;
+
+	/* Write header to buffer and advance position */
+	memcpy(framed_log_buffer + framed_buffer_pos, &entry_header, sizeof(entry_header));
+	framed_buffer_pos += sizeof(entry_header);
```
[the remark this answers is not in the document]
jgrowdenTT: i believe I've fixed this.

include/tenstorrent/smc_msg.h
```diff
@@ -162,6 +162,9 @@ enum tt_smc_msg {
 
 	/** @brief @ref characterisation_rqst "Generic characterization message" */
 	TT_SMC_MSG_CHARACTERISATION = 0xC6,
+
+	/** @brief @ref logging_rqst "FW logging setup/release request" */
```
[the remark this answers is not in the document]
jgrowdenTT: Done, thanks.

lib/tenstorrent/bh_arc/CMakeLists.txt
```diff
@@ -53,6 +52,7 @@ zephyr_library_sources(
   noc2axi.c
   noc_init.c
   pcie.c
+  pcie_msi.c
   pll.c
```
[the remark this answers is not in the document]
jgrowdenTT: If we want pcie logging in recovery we need MSI compiled in.

It's not much source code, and PCIE is there in general, right/

(though; we may not want to register the msg)

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	/* Update write_offset to reflect data size */
+	uint32_t new_write_offset = sizeof(struct fw_log_buffer_header) + framed_buffer_pos;
+
+	NOC2AXIWrite32(0, noc2axi_tlb, log_buffer_iova + 0, new_write_offset);
+
+	/* Set flag to indicate new data is ready */
+	NOC2AXIWrite32(0, noc2axi_tlb, log_buffer_iova + 12, FW_LOG_DATA_READY);
+
```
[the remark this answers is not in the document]
jgrowdenTT: Believe this is resolved. Thanks.

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,432 @@
...
+#define TT_PCIE_LOG_SUBCMD_RELEASE 2
+
+/* Log entry header structure (must match KMD definition) */
+struct fw_log_entry_header {
+	uint16_t msg_size;  /* Size of log message including header */
+	uint8_t log_level;  /* FW log level */
+	uint8_t source;     /* 0=SMC, 1=DMC */
+	uint32_t timestamp; /* FW timestamp in milliseconds */
+	uint32_t sequence;  /* Sequence number */
+} __packed;
```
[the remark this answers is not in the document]
jgrowdenTT: ... If you send a 64KB log.

A - Don't do that. 

B - I get the point. Lets say host gives me 1MB of log space. I make a huge buffer in FW. I can now support 64Kb log other than this field being "small" so from an interface contract it'd be nice if I just truncated it to 64KB.  But the detection of  this case is bugged, we'll wrap instead and thus cause all sorts of host side confusion as it starts to tread that log msg data like framed log msgs. Likely dropping the rest of the delivered buffer.

In source code below, can't know the log entry causes a msg_size truncation until i try to put it in the buffer, since formatting subsitution needs to happen.

If I rip apart zephyr's logging  enough, and I find something i can `snprintf` for buffer population instead of relying on `log_output_func`, it's likely i can solve this. This is a good improvement anyways because going char-by-car is really a waste. Although that loses "automatic" support for other logging formats available in zephyr like dictionary based logging.

I _think_ we just leave it and say don't do that.

lib/tenstorrent/bh_arc/Kconfig.tt_pcie_log
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2024 Tenstorrent AI ULC
```
slindsayTT: pretty sure it's 2026 now :)

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,445 @@
...
+#define FW_LOG_SOURCE_SMC 0
+#define FW_LOG_SOURCE_DMC 1
+
+#define FW_LOG_BUFFER_MAGIC   0x544C4F47 /* "TLOG" */
+#define FW_LOG_DATA_READY     0x1        /* Flag: firmware has new data */
+#define FW_LOG_DATA_PROCESSED 0x0        /* Flag: host has processed data */
+
+/* Static variables to track tt_pcie_log state */
+static bool host_available; /* True when host buffer is set up */
+static uint64_t log_buffer_iova;
```
slindsayTT, line 59: I think we might want to have queue of these so that kmd can provide more than one. 
If there's confusion, kmd can just look at the buffers it handed out to see which ones are ready for consumption.

slindsayTT (commented):
mostly looks good to me

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: @joelsmithTT?

No complaints on my end. Currently with the negligible amount we're logging a multi-buffer scheme is overkill. As we turn on more and more logs, multi-buffers would be useful to make sure we're not dropping

joelsmithTT: Single buffer is simpler for the driver.  Do you have a design doc that explains the choice of this technique (vs a single producer/single consumer ring)?

jgrowdenTT: No design doc for multi-buffer.

Single buffer mode design will end up here, as part of this PR. Ring was complicated to resync logs and handle wrapping scenarios. This resyncs trivially (next buffer delivered has a log at the top) and there is no wrap to handle.

jgrowdenTT: Docs are up.

doc/services/kmd_logging/index.rst
```diff
@@ -0,0 +1,161 @@
...
+     - ``uint32_t``
+     - 4
+     - Firmware timestamp captured for the log entry.
+   * - ``sequence``
+     - ``uint32_t``
+     - 4
+     - Monotonic sequence number used for gap detection.
+
+Payload bytes follow each entry header and are emitted by Zephyr's log output
+formatter in text mode when configured.
```
alewycky-tenstorrent: Please NUL-terminate each string.

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,449 @@
...
+	uint16_t msg_size;  /* Size of log message including header */
+	uint8_t log_level;  /* FW log level */
+	uint8_t source;     /* 0=SMC, 1=DMC */
+	uint32_t timestamp; /* FW timestamp in milliseconds */
+	uint32_t sequence;  /* Sequence number */
+} __packed;
+
+/* Buffer header at start of shared buffer (must match KMD definition) */
+struct fw_log_buffer_header {
+	uint32_t write_offset; /* Current write position (circular) */
```
joelsmithTT: Is the comment accurate?

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,449 @@
...
+
+/* Static variables to track tt_pcie_log state */
+static bool host_available; /* True when host buffer is set up */
+static uint64_t log_buffer_iova;
+static uint32_t log_buffer_size;
+static uint32_t sequence_number;
+
+/* NOC2AXI TLB configuration for direct host memory access */
+#define PCIE_X 19 /* PCIe NOC coordinates as specified by user */
+#define PCIE_Y 24
```
joelsmithTT: I don't understand the comment after the `19`.

Do we need to handle parts that don't have NOC translation enabled?

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,449 @@
...
+	uint32_t magic_value;
+
+	if (framed_buffer_pos == 0 || !host_available) {
+		return; /* Nothing to flush or host not available */
+	}
+
+	/* Setup NOC2AXI TLB for host memory access */
+	uint8_t noc2axi_tlb = 0;
+
+	NOC2AXITlbSetup(0, noc2axi_tlb, PCIE_X, PCIE_Y, log_buffer_iova);
```
joelsmithTT: Is this safe -- nothing else can use this window simultaneously?

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,449 @@
...
+/* tt_pcie_log sub-commands */
+#define TT_PCIE_LOG_SUBCMD_SETUP   1
+#define TT_PCIE_LOG_SUBCMD_RELEASE 2
+
+/* Log entry header structure (must match KMD definition) */
+struct fw_log_entry_header {
+	uint16_t msg_size;  /* Size of log message including header */
+	uint8_t log_level;  /* FW log level */
+	uint8_t source;     /* 0=SMC, 1=DMC */
+	uint32_t timestamp; /* FW timestamp in milliseconds */
```
joelsmithTT, line 35: is the comment true?

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,449 @@
...
+}
+
+/**
+ * @brief Character output function for Zephyr log formatting - writes directly to 4KB buffer
+ * @param data Data to append to framed log buffer
+ * @param length Length of data
+ * @param ctx Context (unused)
+ * @return Length processed
+ */
+static int putc_in_frame(uint8_t *data, size_t length, void *ctx)
```
joelsmithTT, line 291: This one is from Claude:

Zephyr's `log_output_write` (`log_output.h:200`) loops `while (len != 0)` on the callback's
return value. `putc_in_frame` returns 0 once `framed_log_buffer` is full, so the log
processing thread spins forever — while holding `buffer_mutex`, which then times out every
flush and every setup attempt.

doc/services/kmd_logging/index.rst
```diff
@@ -0,0 +1,161 @@
...
+Current Behavior Notes
+----------------------
+
+- Firmware batches data and sends only when host is ready.
+- If host has not yet consumed the previous batch, firmware drops the pending
+  local batch for forward progress.
+- If local staged data exceeds available host payload space, firmware truncates
+  to fit and logs a warning.
+- Data is currently copied as a contiguous batch (not a true host-side circular
+  producer/consumer protocol yet).
```
joelsmithTT, line 141: Are you open to adding a version field to `fw_log_buffer_header` in case we want to evolve the protocol?

lib/tenstorrent/bh_arc/tt_pcie_log.c
joelsmithTT: The milliseconds comment I mean.

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: No. I'll clean it up.

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: a) AI slop. Will cleanup.

b) No idea. I've changed it to 2/0.

doc/services/kmd_logging/index.rst
jgrowdenTT: I'll double check what's there. Thanks.

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: Not really sure. I'll ask AI to look into it.

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: Yeah, I think this is right. We're really at the mercy of the flush issued on `328` going through. 

I  don't think I can release the mutex without incurring a race condition. I'll think about it more.

Alternatively, I can just lie and say I did the whole payload. I'll have to look at the ramifications of that.

jgrowdenTT: Or I go back to "if I had `snprintf` this is easy and I know I overflow before I writing to the buffer" which means I could re-investigate whiteboxing the logger structure. 

I've got options and I'll look into them. Thanks Joel.

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: AI seems to think it is; basically routing through k_uptime_get which is ms since boot.

doc/services/kmd_logging/index.rst
jgrowdenTT: It is not NUL terminated. I have made our backend do this.

doc/services/kmd_logging/index.rst
jgrowdenTT: Yes. Discussing offline.

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,461 @@
...
+	 *
+	 * In practice, expectation is that FW buffer size is less than or equal to host's.
+	 */
+	uint32_t available_host_space = log_buffer_size - sizeof(struct fw_log_buffer_header);
+	uint32_t bytes_to_write = MIN(framed_buffer_pos, available_host_space);
+
+	if (framed_buffer_pos > available_host_space) {
+		LOG_WRN("Truncating log data: %u bytes -> %u bytes (host space limit)",
+			framed_buffer_pos, bytes_to_write);
+	}
```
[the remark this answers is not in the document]
jgrowdenTT: Probably correct. I was hoping I could pick this up on a different backend since we still have tt-console vuart backend logging. Ah well.

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,457 @@
...
+
+/* Static variables to track tt_pcie_log state */
+static bool host_available; /* True when host buffer is set up */
+static uint64_t log_buffer_iova;
+static uint32_t log_buffer_size;
+static uint32_t sequence_number;
+
+/* NOC2AXI TLB configuration for direct host memory access */
+/* NOC2AXI TLB configuration for direct host memory access */
+#define PCIE_X 2
```
joelsmithTT: IIRC an earlier commit had this as X=19, Y=24 which should always work provided NOC translation is enabled. If NOC translation is not enabled, the PCIe tile that is connected to the host can be either X=2, Y=0 or X=11, Y=0.

jgrowdenTT: Lets just assume translation is enabled then?

joelsmithTT: It's enabled on all production parts but if we want this feature to be safe on things like the characterization boards and "scrappy" cards then we probably want a check for it to opt out (so that FW doesn't send NOC traffic to a phantom tile).  There's a bit in the telemetry table that tells software whether NOC translation is active so I assume this is an easy check to do from inside the FW.

Alternatively there may be a way to determine the correct X coordinate (2 or 11) to use; unfortunately I don't think the KMD [technique](https://github.com/tenstorrent/tt-kmd/blob/eb8178ed176935d9f15e4a6981873bbc9363707d/blackhole.c#L299-L302) for doing this works from within the ARC.

@afongTT or @alewycky-tenstorrent might have some insights...

joelsmithTT:
Related: https://github.com/tenstorrent/tt-kmd/issues/257

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: We can opt out of the feature altogether in that case, if it's easiest. This is already KCONFIGurable.

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: 0 is a bad choice. I'll pivot to 15. 

I think eventually we should pool these in a zephyr driver.

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: Well. On top of the KCONFIG, I also reject the message if noc translation is disabled. 

Dunno what else we want done here. I think this is fine for now.

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,461 @@
...
+		NOC2AXIWrite32(0, noc2axi_tlb, write_addr + i, word_data);
+	}
+
+	/* Update write_offset to reflect data size */
+	uint32_t new_write_offset = sizeof(struct fw_log_buffer_header) + bytes_to_write;
+
+	NOC2AXIWrite32(0, noc2axi_tlb, log_buffer_iova + 0, new_write_offset);
+
+	/* Hand ownership to host to indicate new data is ready */
+	NOC2AXIWrite32(0, noc2axi_tlb, log_buffer_iova + 12, FW_LOG_BUFFER_OWNER_HOST);
```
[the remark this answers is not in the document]
jgrowdenTT: Fine for now. We can address when we get a new version.

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,467 @@
...
+	}
+
+	/* Mark where this entry starts */
+	entry_start_pos = framed_buffer_pos;
+
+	/* Create header and reserve space for it */
+	entry_header.msg_size = 0; /* Will be updated after formatting */
+	entry_header.log_level = log_msg_get_level(&msg->log);
+	entry_header.source = FW_LOG_SOURCE_SMC;
+	entry_header.timestamp = msg->log.hdr.timestamp;
```
alexapostolu, line 359: I don't know if this is correct but the comment in tt_pcie_log.c says `entry_header.timestamp` is in milliseconds, but that is not enforced here (eg. `msg->log.hdr.timestamp` could be in ticks).

lib/tenstorrent/bh_arc/tt_pcie_log.c
alexapostolu: I think it would be good to create some central document and write down TLB window reservations

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,467 @@
...
+	uint32_t message_start_pos;
+
+	/* Lock mutex to prevent collision with flush timer */
+	if (k_mutex_lock(&buffer_mutex, K_MSEC(10)) != 0) {
+		/* Timeout - drop this data */
+		return;
+	}
+
+	/* Check if we need to flush buffer for space */
+	if (framed_buffer_pos + sizeof(entry_header) + 256 > sizeof(framed_log_buffer)) {
```
alexapostolu: Nit: It might be good to use a macro for 256

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,467 @@
...
+
+	/* Write in 32-bit word chunks for efficiency */
+	for (uint32_t i = 0; i < bytes_to_write; i += 4) {
+		uint32_t word_data = 0;
+		uint32_t bytes_to_copy = MIN(4, bytes_to_write - i);
+
+		/* Pack up to 4 bytes into a 32-bit word */
+		memcpy(&word_data, &framed_log_buffer[i], bytes_to_copy);
+		NOC2AXIWrite32(0, noc2axi_tlb, write_addr + i, word_data);
+	}
```
alexapostolu, line 267: Are we certain `bytes_to_write` is 4 byte aligned? Can we enforce this? If not we could have two loops, the first loop write all the full 32-bit words and the second loop writes the remaining 1-3 bytes if they exist.

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,467 @@
...
+
+static char buf[1];
+LOG_OUTPUT_DEFINE(log_output_tt_pcie, putc_in_frame, buf, 1);
+
+/**
+ * @brief Zephyr log backend process function - called for each individual log message
+ * @param backend The backend instance
+ * @param msg The log message to process
+ */
+static void backend_process(const struct log_backend *const backend, union log_msg_generic *msg)
```
alexapostolu, line 323: I could be wrong but if we configure `LOG_MODE_IMMEDIATE=y` then the mutex locks here would be incorrect according to the API, as this *could* run in an ISR, https://docs.zephyrproject.org/2.7.5/reference/kconfig/CONFIG_LOG_MODE_IMMEDIATE.html

To prevent this I think we can just add `depends on !LOG_MODE_IMMEDIATE` to the `config TT_PCIE_LOG_BACKEND` to prevent that senario

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: I had AI hunt this down earlier. 

Yes for your current firmware build, the value sent in timestamp is in milliseconds.

> Evidence:
> 
> The field is populated directly from Zephyr’s log timestamp at [tt_pcie_log.c:355](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).
> Zephyr selects the low-frequency timestamp source when HW cycles/sec is greater than 1,000,000 at [log_core.c:282](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).
> That low-frequency source is k_uptime_get_32 (milliseconds) at [log_core.c:242](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).
> Your build has 800 MHz HW cycles, so that condition is true: [.config:49](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html).
> So the comment at [tt_pcie_log.c:35](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is correct for this build.
> Caveat: if timestamp source/config is changed later (for example custom timestamp func), the unit could change.
> 

Are you asking, what if we run ARC FW at sub 1MHz? Or is there something I'm missing w.r.t. configuration?

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: It can be 0 (end loop) 1, 2, 3, or 4. In every case I believe the operation is correct. What's the concern?

sorry, that was for bytes_to copy. I'm less sure why bytes_to_write being unaligned is an issue.

jgrowdenTT:
> I think it would be good to create some central document and write down TLB window reservations

I'm kinda of the opinion a pool driver for this is better. I can work on it in my background time, unless someone else wants it

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,468 @@
...
+	/* Check buffer ownership before writing new data */
+	owner_word = NOC2AXIRead32(0, noc2axi_tlb, log_buffer_iova + 12);
+	if ((owner_word & 0xFF) == FW_LOG_BUFFER_OWNER_HOST) {
+		/* Host hasn't processed previous data yet, drop this batch */
+		framed_buffer_pos = 0;
+		return;
+	}
+
+	/* Write entire framed buffer to host starting after header */
+	uint32_t write_addr = log_buffer_iova + sizeof(struct fw_log_buffer_header);
```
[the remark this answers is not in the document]
jgrowdenTT: That doesn't do anything. The addr would be truncated down.  (24B TLB_LOG_SIZE)

The important part is that the TLB itself is pointing to log_buffer_iova.

lib/tenstorrent/bh_arc/tt_pcie_log.c
alexapostolu: Hmm looking into this further, I don't think what I originally said is a problem.

lib/tenstorrent/bh_arc/tt_pcie_log.c
alexapostolu: Yea I think the only cause is if we have a custom timestamp function later on but I don't see any reason why we would

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,472 @@
...
+		LOG_INF("Host tt_pcie_log re-setup: old=0x%llx, new=0x%llx", log_buffer_iova,
+			new_buffer_iova);
+	}
+
+	/* Update to new buffer (could be same address if host reused buffer) */
+	log_buffer_iova = new_buffer_iova;
+	log_buffer_size = buffer_size;
+
+	/* Setup NOC2AXI TLB mapping for host memory access. */
+	uint8_t noc2axi_tlb = TT_PCIE_LOG_TLB;
```
joelsmithTT, line 151: `SendPcieMsi` uses TLB 0, which is used in various places.  I'm not expert enough in how the FW works to know if this is an issue, but it looks like `flush_buffer_to_host` (which calls `SendPcieMsi`) is called from a worker thread.  Is it possible for the worker to try using TLB 0 simultaneously with the main thread doing something with it?

jgrowdenTT: 14 it is, and we _really_ need a pool of these things.

joelsmithTT (commented):
This looks good to me.  I defer official merge approval to folks more familiar with the FW than myself.

lib/tenstorrent/bh_arc/pcie_msi.c
```diff
@@ -61,7 +61,7 @@ void SendPcieMsi(uint8_t pcie_inst, uint32_t vector_id)
 		msi_data += vector_id;
 
 		const uint8_t ring = 0;
-		const uint8_t tlb_num = 0;
+		const uint8_t tlb_num = 14;
 		const uint8_t x = pcie_inst == 0 ? PCIE_INST0_LOGICAL_X : PCIE_INST1_LOGICAL_X;
 		const uint8_t y = PCIE_LOGICAL_Y;
```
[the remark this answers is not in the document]
jgrowdenTT: 12 is used by GDDR. nvm that's 13. 12 it is.

lib/tenstorrent/bh_arc/tt_pcie_log.c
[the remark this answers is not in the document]
afongTT: You could limit the `TT_PCIE_LOG_BACKEND_BUFFER_SIZE` KConfig to a range to block this case indirectly?

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,472 @@
...
+#define FW_LOG_BUFFER_OWNER_HOST 0x1        /* Host owns buffer, firmware has produced data */
+#define FLUSH_THRESHOLD_BYTES    (sizeof(struct fw_log_entry_header) + 256)
+
+/* Static variables to track tt_pcie_log state */
+static bool host_available; /* True when host buffer is set up */
+static uint64_t log_buffer_iova;
+static uint32_t log_buffer_size;
+static uint32_t sequence_number;
+
+#define TT_PCIE_LOG_TLB 15
```
afongTT: Ah, sorry, one more TLB thing... t6py hardcodes using this TLB when accessing NOC via JTAG2AXI. It's the only way to reach NOC using SYS JTAG.
- JTAG2AXI -> SMC -> NOC2AXI TLB 15 -> NOC

Ideally change it to a different one. Maybe make some TLB defines in noc2axi.h to "reserve" the ones we use?

doc/services/kmd_logging/index.rst
```diff
@@ -0,0 +1,167 @@
...
+
+If host is not set up, firmware still formats logs into a local staging buffer.
+Once setup succeeds, subsequent flushes transfer staged data to host memory.
+
+Shared Buffer Contract
+----------------------
+
+Firmware and KMD must agree on these packed headers:
+
+.. list-table:: Buffer Metadata (16 bytes)
```
afongTT, line 47: Is the implication that KMD has to set up this buffer metadata before sending the `SETUP` command?

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,472 @@
...
+	uint32_t available_host_space = log_buffer_size - sizeof(struct fw_log_buffer_header);
+	uint32_t bytes_to_write = MIN(framed_buffer_pos, available_host_space);
+
+	/* Write in 32-bit word chunks for efficiency */
+	for (uint32_t i = 0; i < bytes_to_write; i += 4) {
+		uint32_t word_data = 0;
+		uint32_t bytes_to_copy = MIN(4, bytes_to_write - i);
+
+		/* Pack up to 4 bytes into a 32-bit word */
+		memcpy(&word_data, &framed_log_buffer[i], bytes_to_copy);
```
afongTT, line 265: Could use ARC DMA for this, but not necessary.

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,472 @@
...
+/**
+ * @brief Zephyr log backend panic function - called when system panics
+ * @param backend The backend instance
+ */
+static void backend_panic(const struct log_backend *const backend)
+{
+	ARG_UNUSED(backend);
+
+	/* Try to acquire mutex with a short timeout during panic */
+	if (k_mutex_lock(&buffer_mutex, K_MSEC(10)) == 0) {
```
afongTT, line 393: Is there an ISR hazard here?

Probably not blocking minor corner case.

afongTT (commented):
Just one more minor TLB allocation thing. Otherwise looks good.

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: I think it should be fine with `LOG_MODE_IMMEDIATE=n` which is a dep on the KCONFIG.

doc/services/kmd_logging/index.rst
jgrowdenTT: No; host provides the buffer zeroed, and FW pushes the contents on msg processed.

I think this is still a good approach, especially w.r.t. version, since host can know from buffer metadata what version FW is supporting. And FW can know from msg which version host is requesting..

lib/tenstorrent/bh_arc/tt_pcie_log.c
jgrowdenTT: (╥﹏╥)

alexapostolu (dismissed):

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -0,0 +1,472 @@
...
+	/* Update write_offset to reflect data size */
+	uint32_t new_write_offset = sizeof(struct fw_log_buffer_header) + bytes_to_write;
+
+	NOC2AXIWrite32(0, noc2axi_tlb, log_buffer_iova + 0, new_write_offset);
+
+	/* Hand ownership to host to indicate new data is ready */
+	NOC2AXIWrite32(0, noc2axi_tlb, log_buffer_iova + 12, FW_LOG_BUFFER_OWNER_HOST);
+
+	/* Trigger interrupt to notify host of new data */
+	SendPcieMsi(0, 0);
```
alayander: This should pass in the active pcie instance rather than hardcoded 0, right?

Alternatively we can refactor the function to be able to pass in the coordinates (since the function just uses the instance to determine pcie coordinates) and use `PCIE_X` and `PCIE_Y` similar to above to allow noc translation to handle it

jgrowdenTT: I'll fix it now. Thanks.

alayander (approved):

alexapostolu (approved):

lib/tenstorrent/bh_arc/tt_pcie_log.c
afongTT: I think `LOG_MODE_IMMEDIATE=n` only affects when `backend_process` gets called, not `backend_panic`, which could still be from ISR?

PR #1210, opened 2026-05-05
ci: add scripts and workflow to produce data from CI runs

alexapostolu:
* Add pydantic_models.py to define the format in which CI test data will be stored.
* Add create_pipeline_json.py to download raw data from GitHub Actions workflow runs, parse the data, and produce .json data formatted according to pydantic_models.py.
* Add produce-data workflow to grab data from the workflow that triggered it, run the parser, and upload the .json output to a server over SFTP.

Copied over from #1184

### DIFF
--- a/.github/workflows/produce-data.yml
+++ b/.github/workflows/produce-data.yml
@@ -1,0 +1,179 @@
+name: Produce Data
+
+on:
+  workflow_run:
+    workflows:
+      - "HW Smoke"
+      - "HW Soak"
+      - "Metal 🤘"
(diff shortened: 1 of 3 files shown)


alexapostolu:
the failing Produce Data workflow is fine I think since there's no parent run to attach to in this PR

alexapostolu:
Tested the commands manually, it is able to send the CI data to the database to be graphed. However, the only way we know this works 100% is to merge it into main so it can run on real workflows.

alayander (approved):

PR #1211, opened 2026-05-05
Init Voltage PPM: Make VDD_MIN/MAX configurable via SPI FW table

psethiTT:
Replace hardcoded voltage limits with values read from FW table at initialization. This allows users to program voltage limits via update_spi_data_table.py without reflashing firmware.

### DIFF
--- a/lib/tenstorrent/bh_arc/voltage.c
+++ b/lib/tenstorrent/bh_arc/voltage.c
@@ -7,15 +7,16 @@
 #include <zephyr/sys/util.h>
 #include <tenstorrent/smc_msg.h>
 #include <tenstorrent/msgqueue.h>
+#include <zephyr/drivers/misc/bh_fwtable.h>
 
 #include "voltage.h"
 #include "regulator.h"
 #include "dvfs.h"
 
 /* TODO: Get these from SPI parameters */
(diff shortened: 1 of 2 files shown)


lib/tenstorrent/bh_arc/voltage.c
```diff
@@ -55,8 +56,8 @@ void CalculateTargVoltage(void)
 
 int InitVoltagePPM(void)
 {
-	voltage_arbiter.vdd_min = VDD_MIN;
-	voltage_arbiter.vdd_max = VDD_MAX;
+	voltage_arbiter.vdd_min = tt_bh_fwtable_get_fw_table(fwtable_dev)->chip_limits.vdd_min;
+	voltage_arbiter.vdd_max = tt_bh_fwtable_get_fw_table(fwtable_dev)->chip_limits.vdd_max;
```
afongTT: Can you add some bounds checking, similar to what we have for AICLK PPM?
https://github.com/tenstorrent/tt-system-firmware/blob/0444266c059f8643256a97794d215ecfe125fbec/lib/tenstorrent/bh_arc/aiclk_ppm.c#L34-L37

afongTT (approved):

PR #1215, opened 2026-05-08
loudbox_ci: lower prefil chunck size to avoid exceeding dram

alexapostolu:
Lower prefil chunck size to 32 to avoid Llama-3.3-70B  exceeding DRAM memory limit.


afongTT (approved):

afongTT:
Can't pretend to understand what this does, but seems ok?

alexapostolu:
Yea I confirmed with the Metal team that we can just increase the memory limit to fix this issue.

PR #1216, opened 2026-05-08
wormhole_ci: auto trigger wh ci if wh blob changes

alexapostolu:
Auto trigger WH CI if detects WH blob changes.

### DIFF
--- a/.github/workflows/hardware-smoke.yml
+++ b/.github/workflows/hardware-smoke.yml
@@ -44,6 +44,21 @@ jobs:
           MATRIX_CONFIG_WH="$(jq -c '[.[] | select(.product == "wh") | del(.product)]' .github/ci_boards.json)"
           echo "matrix-config-wh=$MATRIX_CONFIG_WH" | tee -a $GITHUB_OUTPUT
 
+  detect-changes:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-24.04
+    permissions:
+      pull-requests: read
+    outputs:
+      wh-blob-changed: ${{ steps.changes.outputs.wh-blob }}
+    steps:
+      - uses: dorny/paths-filter@v4
+        id: changes
+        with:
+          filters: |
+            wh-blob:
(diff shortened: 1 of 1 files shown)


.github/workflows/hardware-smoke.yml
```diff
@@ -44,6 +44,34 @@ jobs:
...
+      wh-blob-changed: ${{ steps.changes.outputs.wh-blob-changed }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - id: changes
+        shell: bash
+        run: |
+          # Detect whether the Wormhole firmware blob changed in this PR so we
```
afongTT: Thanks, great idea!

I think it would look better with something like: https://github.com/dorny/paths-filter/tree/master#example

alexapostolu: Ah I didn't know about that, done!

.github/workflows/hardware-smoke.yml
```diff
@@ -226,8 +240,8 @@ jobs:
 
   smoke-e2e-wh-test:
     name: WH Smoke E2E Test on ${{ matrix.config.board }}
-    needs: [gen-config]
-    if: github.event_name == 'schedule' || contains(github.event.pull_request.labels.*.name, 'wh-tests')
+    needs: [gen-config, detect-changes]
+    if: github.event_name == 'schedule' || needs.detect-changes.outputs.wh-blob-changed == 'true'
```
afongTT: Oh, sorry, can you keep the label option too?

alexapostolu: Sure yea, done

afongTT: Oops, I think you lost the `needs.detect-changes.outputs.wh-blob-changed == 'true'` part now.

alexapostolu: Fixed

afongTT (approved):

PR #1223, opened 2026-05-13
tt-system-firmware: Zephyr 4.4 migration

jgrowdenTT:
Move to Zephyr 4.4.

Fix all the things. I've sorta forgotten everything but:

- The (incorrect?) stack protect throw -> rework the i2c write to remove the VLA fixes the issue.
- DMA align comes from DTS, not macros, so remove old macros
- Find SDK correctly
- Compliance for a few dts/dtsi files


test-conf/tests/drivers/dma/chan_link_transfer/arc_hs.conf
```diff
@@ -1 +1 @@
-CONFIG_DMA_LINK_ALIGNMENT=32
+CONFIG_STD_C11=y
```
slindsayTT, line 1: hmm, some gemini querying leads to
`To disable GNU extensions and follow strict ISO C or C++ standards, use flags like -std=c11 or -std=c++17 combined with -pedantic`
So maybe we're compiling with that flag now?

slindsayTT (approved):
I can't claim to be expert but nothing immediately leaps out as a problem.

test-conf/tests/drivers/dma/chan_link_transfer/arc_hs.conf
jgrowdenTT: Think this was Daniel... I'll play around with it to see how necessary it is.

This would only apply to the chan_link_transfer test. Not sure about the others.

alexapostolu (approved):

MAINTAINERS.yml
```diff
@@ -555,3 +555,10 @@ Tenstorrent Platforms:
     - "zephyr 🪁"
   description: |
     Maintained upstream.
+
+"West project: tf-psa-crypto":
```
afongTT, line 559: What do we use this for?

lib/tenstorrent/bh_arc/dw_apb_i2c.c
```diff
@@ -536,19 +536,33 @@ uint32_t I2CTransaction(uint32_t id, const uint8_t *write_data, uint32_t write_l
...
-	/* Combines command and data into a single buffer prior to calling the I2CTransaction
-	 * function, which treats the combined buffer as a single transaction.
-	 */
-	uint32_t combined_buf_size = command_byte_size + data_byte_size;
-	uint8_t combined_buf[combined_buf_size];
+	if (asic_state == A3State) {
+		return IC_ABRT_A3_STATE;
+	}
+
+	const uint8_t *cmd_buf = (const uint8_t *)&command;
```
afongTT, line 543: I guess this is the variable length array issue.

Somehow the variable length array in I2CRMWV is ok?...

MAINTAINERS.yml
jgrowdenTT: I _think_ it's linked to MCU boot. This was a Daniel thing.

lib/tenstorrent/bh_arc/dw_apb_i2c.c
jgrowdenTT: Somehow. I wonder if it's because both are on the same stack... or one VLA is feeding into the other... or this VLA needed some math... or etc. 

Again, it's tough to prove that the VLA is the issue but it's enough to say this resolves the issue, and also VLAs are against the rules, so hopefully this is okay.

afongTT (approved):

PR #1231, opened 2026-05-21
lib: bh_arc: eth: add a message for toggling eth tile reset

afongTT:
Add an SMC msg that toggles the ethernet tile reset. This is useful for two purposes:
1. eth tests were previously directly toggling eth reset directly, but
   this now causes issues with SMC FW's eth telemetry polling. Having
   SMC do the reset ensures these processes do not interfere.
2. There may be cases where an individual ethernet tile needs to be
    reset to recover the link.

resolves: SYS-4052

### DIFF
--- a/app/smc/pytest/e2e_smoke.py
+++ b/app/smc/pytest/e2e_smoke.py
@@ -105,6 +105,10 @@ TT_SMC_MSG_SET_ASIC_HOST_FMAX = 0x23
 TT_SMC_MSG_CHARACTERISATION = 0xC6
 TT_SMC_MSG_COUNTER = 0x35
 TT_SMC_MSG_TOGGLE_GDDR_RESET = 0xB6
+TT_SMC_MSG_TOGGLE_ETH_RESET = 0xB0
+
+# ETH toggle reset (eth_tile_reset_rqst) — response[1] uses eth_reset_err from ARC
+ETH_RESET_ERR_INVALID_MASK = 1
 
 # Characterization submessage IDs
 TT_SUB_MSG_SET_HOST_REQUESTED_FMIN = 0x1
@@ -121,6 +125,8 @@ TAG_HOST_AICLK_LIMIT = 70
 NUM_PD = 16
 NUM_VM = 8
 NUM_TS = 8
+
+NUM_ETH = 14
 
 
 def read_telem(arc_chip, telem_idx):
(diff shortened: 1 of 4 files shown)


app/smc/pytest/e2e_smoke.py
```diff
@@ -1245,25 +1241,125 @@ def test_eth_live_status(arc_chip_dut, asic_id):
...
+    assert response[1] == 0
+
+
+def test_eth_toggle_reset_individual(arc_chip_dut, asic_id):
+    """Reset all individual ETH instances."""
+    arc_chip = pyluwen.detect_chips()[asic_id]
+
+    logger.info("Individually reset each ETH instance without SPI FW reload")
+    for eth_inst in range(NUM_ETH):
+        response = send_eth_toggle_reset(arc_chip, 1 << eth_inst, no_fw_reload=True)
```
jgrowdenTT, line 1387: Good test, i was wondering if we can make it a bit stricter? i.e. i'm worried an implementation that didn't reset the eth but just returned `response[0] == 0` happily would pass this. 

Do we have visibility into a heartbeat? 

(this feedback can wait until a later PR, if this is pressing)

app/smc/pytest/e2e_smoke.py
```diff
@@ -1245,25 +1241,125 @@ def test_eth_live_status(arc_chip_dut, asic_id):
...
+        assert response[0] == 0, (
+            f"ETH {eth_inst}: expected success, got status={response[0]} detail={response[1]}"
+        )
+        assert response[1] == 0
+
+    logger.info("Individually reset each ETH instance with SPI FW reload")
+    for eth_inst in range(NUM_ETH):
+        response = send_eth_toggle_reset(arc_chip, 1 << eth_inst)
```
jgrowdenTT, line 1397: Same comment with the SPI FW reload. Past a `0` how do we really know this happened?

include/tenstorrent/msgqueue.h
```diff
@@ -570,6 +570,47 @@ enum gddr_reset_err {
...
 };
 
+/** @brief Host request to reset and reinitialize one or more ETH tiles
+ *
+ * The handler performs the appropriate sequencing to reset and (optionally)
+ * restart ERISC FW on the requested ethernet tiles.
+ *
+ * Request: @c eth_inst_mask is a bitmask of ETH instance indices (bit N selects ETH N).
+ * Bits for harvested instances are silently dropped. An empty effective mask returns success
+ * (no-op). When @c no_fw_reload is 1, SPI FW/cfg reload and @c ReleaseEthReset are skipped.
```
jgrowdenTT: so... The ERISC won't start running again, unless you specified to reload FW? Is eth therefore non functional after this msg is processed with `no_fw_reload == 1`?

...And that's what is being requested? lol.

lib/tenstorrent/bh_arc/eth.c
```diff
@@ -529,6 +533,133 @@ static void EthInit(void)
...
+		for (uint8_t eth_inst = 0; eth_inst < MAX_ETH_INSTANCES; eth_inst++) {
+			if (!IS_BIT_SET(mask, eth_inst)) {
+				continue;
+			}
+
+			rc = LoadEthFw(eth_inst, ring, buf, sizeof(buf), fw_fd.spi_addr,
+				       fw_fd.flags.f.image_size);
+			if (rc < 0) {
+				rsp->data[1] = ETH_RESET_ERR_FW_LOAD;
+				return 1;
```
alexapostolu, line 640: Is it bad if we return immediately after an error without trying to load FW to the rest of the tiles? Maybe we can restructure this so LoadEthFwCfg is only called when LoadEthFw passes, and at the end of the loop it'll return any errors it saw.

lib/tenstorrent/bh_arc/eth.c
```diff
@@ -529,6 +533,133 @@ static void EthInit(void)
...
+		rsp->data[1] = ETH_RESET_ERR_INVALID_MASK;
+		return 1;
+	}
+
+	uint32_t mask = requested & tile_enable.eth_enabled;
+
+	if (mask == 0) {
+		rsp->data[1] = 0;
+		return 0;
+	}
```
alexapostolu, line 568: Is it better if we explicitly return an error when the user accidentally resets a harvested tile?

app/smc/pytest/e2e_smoke.py
afongTT: Yeah, I should probably add a scratch register checking that gets cleared on reset, now I just need to find an appropriate scratch reg...

include/tenstorrent/msgqueue.h
afongTT: Yes, basically the test loads its own FW, but needs the eth tile to start in a clean state.

lib/tenstorrent/bh_arc/eth.c
afongTT: I was thinking I should maybe respond with success (0 in data[0]), and also which eth instances actually got reset (in data[1])?

One benefit would be that "reset all eth" is always `TT_SMC_MSG_TOGGLE_ETH_RESET`, with `eth_inst_mask` = 0x3fff.

lib/tenstorrent/bh_arc/eth.c
afongTT: I think either way we won't start the FW on any instance (and probably don't want to) so we can probably exit early from the FW loading process?

app/smc/pytest/e2e_smoke.py
afongTT: Added a scratch register clearing check.

app/smc/pytest/e2e_smoke.py
afongTT: Added a scratch register clearing check.

lib/tenstorrent/bh_arc/eth.c
afongTT: Changed the definition of data[1] to be "which eth tiles got reset" on message success.

lib/tenstorrent/bh_arc/eth.c
```diff
@@ -529,6 +533,138 @@ static void EthInit(void)
...
 
+static void assert_eth_risc_soft_reset(uint32_t eth_inst, uint32_t ring)
+{
+	const uint32_t kAllRiscSoftReset = 0x47800;
+	uint8_t x, y;
+
+	GetEthNocCoords(eth_inst, ring, &x, &y);
+	NOC2AXITlbSetup(ring, ETH_SETUP_TLB, x, y, ETH_RISC_DEBUG_SOFT_RESET_0);
+	NOC2AXIWrite32(ring, ETH_SETUP_TLB, ETH_RISC_DEBUG_SOFT_RESET_0, kAllRiscSoftReset);
+}
```
alexapostolu, line 544: I'm wondering if it would be good to create a Zephyr reset driver for this (as another Jira ticket). Cause a lot of the reset functionality appears to be the same logic between GDDR, tensix and eth tiles.

include/tenstorrent/msgqueue.h
```diff
@@ -570,6 +570,48 @@ enum gddr_reset_err {
...
```
alexapostolu: Nit: It might be helpful to add the max number of ETH instances in the documentation here

alexapostolu (dismissed):
Looks good, we should wait for James to do another review too

jgrowdenTT (dismissed):

lib/tenstorrent/bh_arc/eth.c
afongTT: Sure, if you want to make a JIRA ticket in the backlog.

include/tenstorrent/msgqueue.h
afongTT: Updated it

alexapostolu (dismissed):

afongTT:
Oops accidentally checked in a testing thing. Fixing it now...

alexapostolu (approved):

PR #1237, opened 2026-05-26
ccfgovr: new persistent cmfw override

alexapostolu:
ccfgovr: new persistent cmfw override


drivers/misc/bh_fwtable/spirom_protobufs/fw_table_override.proto
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright (c) 2026 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+syntax = "proto3";
+
+message FwTableOverride {
```
afongTT, line 15: Seems like this would be better if it used the fw_table.proto, rather than adding a copy of the same thing?

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -209,6 +220,292 @@ static int tt_bh_fwtable_load(const struct device *dev, enum bh_fwtable_e table)
...
+		ccfgovr_apply_eth_property(&base->eth_property_table, &ovr->eth_property_table);
+	}
+	if (ovr->has_product_spec_harvesting) {
+		base->has_product_spec_harvesting = true;
+		ccfgovr_apply_product_spec(&base->product_spec_harvesting,
+					   &ovr->product_spec_harvesting);
+	}
+}
+
+#undef CCFGOVR_MERGE_SCALAR
```
jgrowdenTT: why?

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -209,6 +220,292 @@ static int tt_bh_fwtable_load(const struct device *dev, enum bh_fwtable_e table)
...
+}
+
+static bool ccfgovr_seq_is_newer(uint32_t a, uint32_t b)
+{
+	return ((int32_t)(a - b)) > 0;
+}
+
+#define CCFGOVR_MERGE_SCALAR(base, ovr, field)                                                     \
+	do {                                                                                       \
+		if ((ovr)->has_##field) {                                                          \
```
jgrowdenTT: Could we log if there's a delta?

drivers/misc/bh_fwtable/spirom_protobufs/fw_table_override.proto
alexapostolu: But the difference is we need to add the `optional` field to every element, so I don't think we can use the `fw-table.proto`

drivers/misc/bh_fwtable/bh_fwtable.c
alexapostolu: Since that macro is temporarily used for those specific functions. But I don't have a strong preference for this, I can remove it

drivers/misc/bh_fwtable/spirom_protobufs/fw_table_override.proto
afongTT: I was thinking we would use [pb_decode_ex](https://jpa.kapsi.fi/nanopb/docs/reference.html#pb_decode_ex) with `PB_DECODE_NOINIT | PB_DECODE_NULLTERMINATED`

to override fields of the `FwTable fw_table` directly.

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -209,6 +218,156 @@ static int tt_bh_fwtable_load(const struct device *dev, enum bh_fwtable_e table)
...
+		if (order[i]->hdr.body_len == 0) {
+			return;
+		}
+
+		uint32_t saved_fw_bundle_version = data->fw_table.fw_bundle_version;
+		pb_istream_t stream = pb_istream_from_buffer(body, order[i]->hdr.body_len);
+
+		if (!pb_decode_ex(&stream, FwTable_fields, &data->fw_table,
+				  PB_DECODE_NOINIT | PB_DECODE_NULLTERMINATED)) {
+			LOG_WRN("ccfgovr bank '%s' protobuf decode failed; fw_table may be "
```
afongTT: I think on the partially overwritten case, we should return a failure. Then outside of this function, load CMFWCFG again, so that the fw_table doesn't have the partially overwritten behaviour.

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -209,6 +218,156 @@ static int tt_bh_fwtable_load(const struct device *dev, enum bh_fwtable_e table)
...
+				"partially overridden",
+				order[i]->tag);
+		}
+		data->fw_table.fw_bundle_version = saved_fw_bundle_version;
```
afongTT: Why was fw_bundle_version getting overwritten? Doesn't seem like it should be special?

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -209,6 +218,156 @@ static int tt_bh_fwtable_load(const struct device *dev, enum bh_fwtable_e table)
...
+		rc = flash_read(flash, info->fd.spi_addr + sizeof(info->hdr), body,
+				info->hdr.body_len);
+		if (rc < 0) {
+			LOG_WRN("flash_read(%s body) failed: %d", info->tag, rc);
+			return false;
+		}
+	}
+
+	cksum = info->hdr.magic + info->hdr.seq + info->hdr.body_len;
+	cksum = tt_boot_fs_cksum(cksum, body, info->hdr.body_len);
```
afongTT: Could you use [CRC32](https://docs.zephyrproject.org/latest/doxygen/html/group__crc.html#gafc24e79ed7f978f5bb813091ef318982) for new checksums?

We only use the `tt_boot_fs_cksum` in tt_boot_fs for compatibility with the boot rom. But since this checksum is independent of tt_boot_fs, I'd recommend using a proper checksum.

drivers/misc/bh_fwtable/spirom_protobufs/fw_table_override.proto
afongTT: Should be able to delete this file now?

alexapostolu:
I do have tests for this but they depend on the bh-mod binary so I haven't pushed them yet. Is it worth building the bh-mod binary in CI?

scripts/encode_spirom_bins.py
```diff
@@ -52,6 +53,24 @@ def write_bin_to_file(in_folder, out_folder, bin_file_name, encoded_data):
...
+    body_len = 0
+    body = b""
+    header_no_cksum = (
+        CCFGOVR_MAGIC.to_bytes(4, "little")
+        + seq.to_bytes(4, "little")
+        + body_len.to_bytes(4, "little")
+        + CCFGOVR_HDR_VERSION.to_bytes(4, "little")
+    )
+    crc = zlib.crc32(header_no_cksum + body)
+    return header_no_cksum + crc.to_bytes(4, "little") + body
```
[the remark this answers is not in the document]
afongTT: Seems unnecessary.

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -209,6 +220,172 @@ static int tt_bh_fwtable_load(const struct device *dev, enum bh_fwtable_e table)
...
+		return;
+	}
+
+	rc = flash_read(flash, info->fd.spi_addr, &info->hdr, sizeof(info->hdr));
+	if (rc < 0) {
+		LOG_WRN("flash_read(%s hdr) failed: %d", info->tag, rc);
+		return;
+	}
+
+	info->header_valid = ccfgovr_header_is_plausible(&info->hdr);
```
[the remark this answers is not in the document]
afongTT: Disagree, since we are not using tt_boot_fs flags.image_size to include the payload.

This change ensures validity by:
1. Partition layout gives 4KB to each ccfgovr table (this is not checked anywhere though)
2. `ccfgovr_header_is_plausible` checks `hdr->body_len <= CCFGOVR_DECODE_BODY_MAX`, so we're ok if CCFGOVR_DECODE_BODY_MAX is < the 4KB from step 1
3. The CRC check

The first two parts are a bit fragile. The third one should be pretty good.

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -209,6 +220,172 @@ static int tt_bh_fwtable_load(const struct device *dev, enum bh_fwtable_e table)
...
+		/*
+		 * Allow-listed merge. Add one branch here for each field exposed
+		 * in fw_table_override.proto; keep this list in sync with the
+		 * `optional` declarations there.
+		 */
+		if (ovr.has_chip_limits && ovr.chip_limits.has_tdp_limit) {
+			LOG_INF("CCFGOVR override: chip_limits.tdp_limit = %u",
+				ovr.chip_limits.tdp_limit);
+			data->fw_table.chip_limits.tdp_limit = ovr.chip_limits.tdp_limit;
+		}
```
afongTT, line 379: This is ok for now, but I would like us to test out @alewycky-tenstorrent's suggestion later:

- mark every field optional
- for nanopb create a .options file that sets proto3:true for every field
- prost writes Some(0) as an actual present field rather than skipping as default - we really need this otherwise we can't ever override a field to false
- merge doesn't replace repeating fields, it appends them, so this doesn't work in that case

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -237,6 +414,10 @@ static int tt_bh_fwtable_init(const struct device *dev)
 		return rc;
 	}
 
+#ifdef CONFIG_BH_FWTABLE_CCFGOVR
+	tt_bh_fwtable_apply_ccfgovr(dev);
+#endif
```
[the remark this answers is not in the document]
afongTT: True, I wonder if we should be using native sim for this, e.g. [SYS-3658](https://tenstorrent.atlassian.net/browse/SYS-3658)

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -209,6 +220,172 @@ static int tt_bh_fwtable_load(const struct device *dev, enum bh_fwtable_e table)
...
+		 * to FwTableOverride_fields and get silently skipped.
+		 */
+		FwTableOverride ovr = FwTableOverride_init_zero;
+		pb_istream_t stream = pb_istream_from_buffer(body, order[i]->hdr.body_len);
+
+		if (!pb_decode_ex(&stream, FwTableOverride_fields, &ovr,
+				  PB_DECODE_NULLTERMINATED)) {
+			LOG_WRN("ccfgovr bank '%s' protobuf decode failed; ignoring override",
+				order[i]->tag);
+			return;
```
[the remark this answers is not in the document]
afongTT: @alexapostolu you should probably fix this one

drivers/misc/bh_fwtable/bh_fwtable.c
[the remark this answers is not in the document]
alexapostolu: I created a ticket for this, SYS-4406 and will do once the fw table is in native sim

afongTT (dismissed):

tests/drivers/fw_table/src/main.c
```diff
@@ -0,0 +1,290 @@
...
+
+	out[total++] = (CHIP_LIMITS_PROTO_INDEX << 3) | 2U;
+	out[total++] = (uint8_t)inner_len;
+	memcpy(out + total, inner, inner_len);
+	total += inner_len;
+	while ((total % 4U) != 0U) {
+		out[total++] = 0x00U;
+	}
+	return total;
+}
```
afongTT, line 86: Can you use nanopb to encode this instead? This is pretty wild.

afongTT (approved):

PR #1239, opened 2026-05-28
MRISC 2.16

akataniTT:
- Parameter Sweeping
- - Enable Parameter sweeping by configuring yaml file 
- - When enabled, a predetermined range of values for any controller config register can be swept
- - Eye height info is stored in the training struct
- Updated CA ODT and vrefc based on the data collected during manufacturing with mrisc 2.15
- Updated DRAM DFE for controller 4 to maximize eye height
- Periodically exit powerdown when idle to allow temperature to update


alexapostolu (approved):
Can you add the PR description to the commit message, so it looks something like this,
```
MRISC 2.16

* Parameter Sweeping
  * Enable Parameter sweeping by configuring yaml file
  * When enabled, a predetermined range of values for any controller config register can be swept
  * Eye height info is stored in the training struct
* Updated CA ODT and vrefc based on the data collected during manufacturing with mrisc 2.15
* Updated DRAM DFE for controller 4 to maximize eye height
* Periodically exit powerdown when idle to allow temperature to update


Signed-off-by: Amin Katani <akatani@tenstorrent.com>
```

akataniTT:
@alexapostolu I updated the commit message as requested.

alexapostolu (approved):

PR #1245, opened 2026-06-02
VF Curve: Make process VF margin user-tunable through fw-table

psethiTT:
Wire firmware-table voltage_margin into the process-based VF curve path so voltage tuning is editable via fw table rather than effectively hardcoded deltas.

Rename voltage_margin_mv to fw_table_voltage_margin_mv for clarity and use it consistently in both InitVFCurve and VFCurve.

In process-based mode:

SS path now adds fw_table_voltage_margin_mv + SS_MARGIN_MV. FF path now adds fw_table_voltage_margin_mv + FF_{LOW,HIGH}_MARGIN_MV Also retune process delta constants:

SS: 45 -> 20
FF low: 25 -> 0
FF high: 45 -> 20
Legacy curve behavior is preserved, with only the variable rename and equivalent fw-table margin usage.

### DIFF
--- a/lib/tenstorrent/bh_arc/vf_curve.c
+++ b/lib/tenstorrent/bh_arc/vf_curve.c
@@ -35,9 +35,9 @@
 /* Process-based margin thresholds and values */
 #define RO_SS_THRESHOLD        2500.0F
 #define FREQ_THRESHOLD_MHZ     1200.0F
-#define SS_MARGIN_MV           45.0F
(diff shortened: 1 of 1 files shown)


alexapostolu (approved):

psethiTT:
@afongTT if you can have a look at the changes and check if we can improve on this implementation.

afongTT (approved):

PR #1247, opened 2026-06-02
scripts: build-combined-fwbundle.sh: parallelize build

afongTT:
Build each board type in a background process to parallelize the FW build.


afongTT:
After this change on my Ryzen 7600X system:
```
real    6m53.530s                                                             
user    57m31.465s                     
sys     15m43.133s 
```

Before this change:
```
real    13m45.060s
user    49m40.265s
sys     14m31.485s
```

alexapostolu (approved):

PR #1248, opened 2026-06-03
lib: bh_arc: modify fan curve for fan table rev3.1

alayander:
Implement modified fan curve for rev3.1 of DRAM fan table. Also adds hysteresis mechanism for gddr fan curve and disables initial spin up.

resolves: SYS-2938


lib/tenstorrent/bh_arc/fan_ctrl.c
```diff
@@ -50,22 +52,42 @@ STATIC uint32_t fan_curve(float max_asic_temp, float max_gddr_temp)
...
+		if (max_asic_temp < 72) {
+			fan_speed1 = (uint32_t)(0.833f * max_asic_temp - 29.98f);
+		} else if (max_asic_temp < 75) {
+			fan_speed1 = (uint32_t)(3.333f * max_asic_temp - 209.976f);
+		} else {
+			fan_speed1 = (uint32_t)(4.0f * max_asic_temp - 260.0f);
+		}
 	} else {
 		fan_speed1 = 100;
 	}
```
alexapostolu, line 65: Would it be neater if the if-statements weren't nested

jgrowdenTT (approved):

PR #1250, opened 2026-06-03
tt-system-firmware: zephyr: 4.4.0 plus

jgrowdenTT:
This gets us mctp support and I3c target stupport for the stm32. Also exposes us to a bit of risk upstream

We have to fix the following:

A) Our next fork branch no longer carrying check-compliance changes. Some of these were for diffs, which we don't need support for. Others of these are for managing all the KCONFIGS we couldn't find, which IMO was just a bug and fixed in this PR by further monkey-patching on our compliance script.

B) STM no longer has  it's own config for the transfer timeout and comes with `CONFIG_I2C_TRANSFER_TIMEOUT_MS`

C) We were abusing `user_data` to get `noc_coords` into our dma configuration. This is now done by making a tt-specific driver API to get the coords passed thru. This allows tests to pass and FW to use the driver without mangling or misappropriating the `user_data`


include/zephyr/drivers/dma/dma_tt_bh_noc.h
```diff
@@ -14,14 +14,40 @@ struct tt_bh_dma_noc_coords {
...
-	};
+	const struct tt_bh_dma_noc_driver_api *api = DEVICE_API_GET(tt_bh_dma_noc, dev);
+
+	if (!api->config) {
+		return -ENOSYS;
+	}
+
+	if (cfg == NULL) {
+		return -EINVAL;
+	}
```
alayander: Could add a guard for `coords == NULL` as well.

drivers/dma/dma_tt_bh_noc.c
```diff
@@ -278,19 +278,25 @@ static int tt_bh_dma_noc_config(const struct device *dev, uint32_t channel,
...
 	chan_data->state.last_expected_acks = 0;
 
-	if (config->user_data) {
-		chan_data->coords = *(struct tt_bh_dma_noc_coords *)config->user_data;
-	} else {
-		GetEnabledTensix(&chan_data->coords.source_x, &chan_data->coords.source_y);
-		chan_data->coords.dest_x = 8;
-		chan_data->coords.dest_y = 0;
-	}
+	dma_cfg->channels[channel].coords = *coords;
```
alayander: nit: Can use `chan_data->coords = *coords` matching how the rest of the function handles assignment.

alayander (approved):

PR #1253, opened 2026-06-04
drivers: Use DEVICE_API macro to define api

jgrowdenTT:
Runtime asserts on device compatability verify the API ptr range resides in a section populated by using the DEVICE_API macro. Failure to do so will cause
false detects of an incorrect API being used.

This is necessary for when we rebase on top of 4.4.0 <plus>

### DIFF
--- a/drivers/clock_control/clock_control_tt_bh.c
+++ b/drivers/clock_control/clock_control_tt_bh.c
@@ -627,14 +627,14 @@ static int clock_control_tt_bh_init(const struct device *dev)
 	return 0;
 }
 
-static const struct clock_control_driver_api clock_control_tt_bh_api = {
-	.on = clock_control_tt_bh_on,
-	.off = clock_control_tt_bh_off,
-	.async_on = clock_control_tt_bh_async_on,
-	.get_rate = clock_control_tt_bh_get_rate,
-	.get_status = clock_control_tt_bh_get_status,
-	.set_rate = clock_control_tt_bh_set_rate,
-	.configure = clock_control_tt_bh_configure};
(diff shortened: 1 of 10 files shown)


alayander (approved):

drivers/clock_control/clock_control_tt_bh.c
```diff
@@ -627,14 +627,14 @@ static int clock_control_tt_bh_init(const struct device *dev)
...
-	.set_rate = clock_control_tt_bh_set_rate,
-	.configure = clock_control_tt_bh_configure};
+static DEVICE_API(clock_control,
+		  clock_control_tt_bh_api) = {.on = clock_control_tt_bh_on,
+					      .off = clock_control_tt_bh_off,
+					      .async_on = clock_control_tt_bh_async_on,
+					      .get_rate = clock_control_tt_bh_get_rate,
+					      .get_status = clock_control_tt_bh_get_status,
+					      .set_rate = clock_control_tt_bh_set_rate,
+					      .configure = clock_control_tt_bh_configure};
```
alexapostolu, line 637: The formatting of this seems weird

alexapostolu (approved):

drivers/clock_control/clock_control_tt_bh.c
jgrowdenTT: Want a clang-format off/on? I don't really mind one way or the other.

alexapostolu: clang-format did that? hmm it seems inconsistent with the others but doesn't matter too much

PR #1254, opened 2026-06-04
lib: bh_arc: Test coordinates from func instead of user_data

jgrowdenTT:
We can't use user_data as a coord setter because that could be any sort of struct; upstream tests now use this for a sem.

Instead, solve the issue by opening a new API. Prior calls in our stack must use the API to specifiy the co-ord prior to dma.

Tests specifically gets a co-ord setter that allows the DMA tests to run. This is hacked into non-test code for now.


alexapostolu:
Where do you see upstream tests using user_data as a semaphore? I can't find it in any of the upstream dma tests

jgrowdenTT:
It was added in [this change.](https://github.com/zephyrproject-rtos/zephyr/commit/b0aaa56c1790596a0cebcc7bfccadf8bc4a9820c)

[File in main for reference.](https://github.com/zephyrproject-rtos/zephyr/blob/main/tests/drivers/dma/chan_blen_transfer/src/test_dma.c)

I'm rebased somewhere close to on top of that for the 4.4.0-1 PR.

drivers/dma/dma_tt_bh_noc.c
```diff
@@ -230,6 +232,29 @@ static int noc_dma_transfer(uint32_t cmd, uint32_t ret_coord, uint64_t ret_addr,
...
+	}
+
+	k_spinlock_key_t key = k_spin_lock(&dma_data->lock);
+
+	dma_cfg->channels[channel].coords.source_x = source_x;
+	dma_cfg->channels[channel].coords.source_y = source_y;
+	dma_cfg->channels[channel].coords.dest_x = dest_x;
+	dma_cfg->channels[channel].coords.dest_y = dest_y;
+
+	k_spin_unlock(&dma_data->lock, key);
```
alayander, line 255: Would it be useful to add a flag/guard in `dma_cfg->channels[channel].state` for whether the coordinates were set, which can be checked when attempting `tt_bh_dma_noc_start()`?

drivers/dma/dma_tt_bh_noc.c
alayander: This function header comment is out of date now

drivers/dma/dma_tt_bh_noc.c
jgrowdenTT: Might as well.

afongTT:
I hate that we have to set some state outside of the `dma_api_config` API call, but the API doesn't really give us any other options.

jgrowdenTT:
DNM - Talked with @afongTT and we might have a better soln.

jgrowdenTT:
Unfortunately I need to be on top of my zephyr 4.4.0+ branch in order to drag in some needed device API macros. 

So I'm closing this one.

PR #1263, opened 2026-06-10
lib: bh_arc: add gddr ctf feature

alayander:
Add CTF feature for gddr to trigger a therm trip when:
1) max gddr temperature is greater than or equal to 95 degrees for 1 minute.
2) max gddr temperature reaches 110 degrees.

Also add a native sim test to validate behaviour.

Resolves: SYS-3165, SYS-3374

### DIFF
--- a/app/dmc/src/main.c
+++ b/app/dmc/src/main.c
@@ -217,6 +217,25 @@ static bool process_led_blink_request(struct bh_chip *chip, uint8_t msg_id, uint
 	return false;
 }
 
+static bool process_gddr_therm_trip(struct bh_chip *chip, uint8_t msg_id, uint32_t msg_data)
+{
+	switch (msg_data) {
+	case kGddrThermTripReasonInstantaneous:
+		LOG_ERR("GDDR thermal trip: instantaneous temp >= %dC",
+			GDDR_THERM_TRIP_CRITICAL_TEMP);
+		break;
+	case kGddrThermTripReasonSustained:
+		LOG_ERR("GDDR thermal trip: sustained temp >= %dC for %d minute(s)",
+			GDDR_THERM_TRIP_TEMP, GDDR_THERM_TRIP_DURATION_MIN);
+		break;
+	default:
+		LOG_ERR("GDDR thermal trip: unknown reason %u", msg_data);
+		break;
+	}
+
+	return false;
+}
+
 void process_cm2dm_message(struct bh_chip *chip)
 {
 	typedef bool (*msg_processor_t)(struct bh_chip *chip, uint8_t msg_id, uint32_t msg_data);
@@ -230,6 +249,7 @@ void process_cm2dm_message(struct bh_chip *chip)
 		[kCm2DmMsgIdAutoResetTimeoutUpdate] = process_auto_reset_timeout_update,
 		[kCm2DmMsgTelemHeartbeatUpdate] = process_heartbeat_update,
 		[kCm2DmMsgIdLedBlink] = process_led_blink_request,
+		[kCm2DmMsgIdGddrThermTrip] = process_gddr_therm_trip,
 	};
 
 	for (uint32_t i = 0U; i < kCm2DmMsgCount; i++) {
--- a/lib/tenstorrent/bh_arc/cat.c
+++ b/lib/tenstorrent/bh_arc/cat.c
@@ -6,12 +6,14 @@
 
 #include "cat.h"
 #include "reg.h"
+#include "telemetry.h"
 #include "timer.h"
 
 #include <stdbool.h>
 
 #include <tenstorrent/post_code.h>
 #include <tenstorrent/sys_init_defines.h>
+#include <zephyr/kernel.h>
 #include <zephyr/sys/util.h>
 #include <zephyr/drivers/gpio.h>
 #include <zephyr/init.h>
@@ -48,6 +50,8 @@ typedef union {
 } RESET_UNIT_CATMON_THERM_TRIP_CNTL_reg_u;
 
 #ifndef CONFIG_TT_SMC_RECOVERY
+
+static const int gddr_therm_trip_interval = 100;
 
 static const struct device *const pvt = DEVICE_DT_GET(DT_NODELABEL(pvt));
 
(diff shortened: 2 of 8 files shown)


lib/tenstorrent/bh_arc/cat.h
```diff
@@ -9,4 +9,11 @@
 
 #define T_J_SHUTDOWN 110 /* BH Prod Spec 7.3 */
 
+#define CAT_GDDR_THERM_TRIP_TEMP          95    /* sustained trip threshold, degrees Celsius */
+#define CAT_GDDR_THERM_TRIP_DURATION_MS   60000 /* 1 minute */
+#define CAT_GDDR_THERM_TRIP_CRITICAL_TEMP 110   /* immediate trip threshold, degrees Celsius */
```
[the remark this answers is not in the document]
alayander: The criteria was modified to match the changes in the PR, I've updated the PR description to reflect this.

app/dmc/src/main.c
```diff
@@ -217,6 +217,14 @@ static bool process_led_blink_request(struct bh_chip *chip, uint8_t msg_id, uint
 	return false;
 }
 
+static bool process_gddr_therm_trip(struct bh_chip *chip, uint8_t msg_id, uint32_t msg_data)
+{
+	LOG_ERR("GDDR thermal trip: (%s)",
+		msg_data ? "instantaneous temp >= 110C" : "sustained temp >= 95C for 1 minute");
+
```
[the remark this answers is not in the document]
alayander: The description was updated to match the code.

app/dmc/src/main.c
```diff
@@ -217,6 +217,14 @@ static bool process_led_blink_request(struct bh_chip *chip, uint8_t msg_id, uint
 	return false;
 }
 
+static bool process_gddr_therm_trip(struct bh_chip *chip, uint8_t msg_id, uint32_t msg_data)
+{
+	LOG_ERR("GDDR thermal trip: (%s)",
```
jgrowdenTT: We should apply some self documentation here, as well as some defense against unexpected values

Something like:
```suggestion

char * reason;
if (msg_data == INST_TEMP_TOO_HIGH) {
    reason = "instantaneous temp >= 110c"; //Preferrably a define for 110 pls
}
else if (msg_data == SUSTAINED_TEMP_TOO_HIGH) {
	reason = "sustained temp >= 95C for 1 minute"; // preferrably a define for 95 and 1 minutes
}
else {
	    reason = "unknown reason";
}
	
LOG_ERR("GDDR thermal trip: (%s)", reason)
```

lib/tenstorrent/bh_arc/cat.c
```diff
@@ -22,6 +25,12 @@
 
 LOG_MODULE_REGISTER(cat);
 
+#ifdef CONFIG_ZTEST
```
jgrowdenTT: Prefer not. The function you're externing in the tests, can't it get invoked by the timer advancing?

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -316,6 +316,29 @@ int GetMaxGDDRTemp(void)
 	return max_gddr_temp;
 }
 
+#ifdef CONFIG_ZTEST
```
jgrowdenTT: Prefer not. Prod code and test code should be different.

This is labeled as external, so theoretically it shouldn't matter where you declare this to get it compiled.. So put it in the test.

lib/tenstorrent/bh_arc/telemetry.h
```diff
@@ -338,4 +338,8 @@ void UpdateTelemetryHostAiclkLimit(uint32_t fmax);
 bool GetTelemetryTagValid(uint16_t tag);
 uint32_t GetTelemetryTag(uint16_t tag);
 
+#ifdef CONFIG_ZTEST
```
jgrowdenTT: Same as the advice before, where prod and test code should be different if at all possible, this should be removed.

lib/tenstorrent/bh_arc/cat.c
alayander: I think in the current state, it would require some tricks to modify/access the input and output. I considered using link-time wraps to intercept `GetTelemetryTag` (which is used to read the max gddr temp) and `ReportGddrThermTrip` (which would be used as the output for comparison in the tests) while advancing the timer, but I though it would be cleaner to just just expose the function in general (in latest push). 

Is there an alternative suggestion I should look into instead?

app/dmc/src/main.c
```diff
@@ -217,6 +217,25 @@ static bool process_led_blink_request(struct bh_chip *chip, uint8_t msg_id, uint
 	return false;
 }
 
+static bool process_gddr_therm_trip(struct bh_chip *chip, uint8_t msg_id, uint32_t msg_data)
+{
+	switch (msg_data) {
+	case kGddrThermTripReasonInstantaneous:
```
jgrowdenTT, line 223: ... Do we need to do some action other than log? Or is that later effort?

Getting DMC logs from a trip, in the field, is going to be inconvenient if PCIE is down.

lib/tenstorrent/bh_arc/cat.c
```diff
@@ -115,6 +122,16 @@ static void EnableCAT(uint8_t trim_code, bool shutdown_on_trip)
 	}
 }
 
+void TriggerThermTrip(void)
```
jgrowdenTT: Double check, looks like this can stay static, right?

lib/tenstorrent/bh_arc/cat.c
jgrowdenTT: I think this is good, thanks. There's precedent for externing stuff for test/visibility purposes so this isn't the end of the world. 

for future consideration:
Solution A) I think we advance the timer, change telemetry, verify how the block reacts when the timer advances. 
 
Solution B) You could also create a separate test case for it, and in that test case, include C instead of compiling in lib_bharc. FFF the unknown symbols. That lets you keep `static` at the least, but call it in the test.

I've seen some precedent for this in zephyr. IIRC, AI agents _hate_ it, and maybe TT does to... I don't think it's that bad, and I'm actually quite fond of it, it's just a different level of testing vs what we're doing now.

A and B are not mutally exclusive.

alayander: That makes sense - thanks for explaining further!

app/dmc/src/main.c
alayander: Currently there is no plan for further action beside a log, but perhaps this can be extended on in the future if needed

jgrowdenTT (approved):

app/dmc/src/main.c
afongTT: @alayander can you make a JIRA to take care of this later?

Maybe just do the same thing we do for `therm_trip_count` in telemetry.

alayander: Created SYS-4629

PR #1265, opened 2026-06-11
lib: bh_chip: Fix bus recovery type

jgrowdenTT:
i2c_recover_bus expects a device of type i2c, not smbus.

Tests:

`native_sim` with `CONFIG_ASSERT` popped almost immediately due to this. This resolves the issue.

### DIFF
--- a/include/tenstorrent/bh_arc.h
+++ b/include/tenstorrent/bh_arc.h
@@ -63,6 +63,7 @@ union cm2dmAckWire {
 struct bh_arc {
 	const struct smbus_dt_spec smbus;
 	const struct gpio_dt_spec enable;
+	const struct device *i2c_dev;
 };
 
 typedef struct cm2dmMessageRet {
@@ -85,6 +86,7 @@ int bharc_disable_i2cbus(const struct bh_arc *dev);
 
 #define BH_ARC_INIT(n)                                                                             \
 	{.smbus = SMBUS_DT_SPEC_GET(n),                                                            \
+	 .i2c_dev = DEVICE_DT_GET(DT_PHANDLE(DT_BUS(n), i2c)),                                     \
 	 .enable = COND_CODE_1(DT_PROP_HAS_IDX(n, gpios, 0),	({	\
 			.port = DEVICE_DT_GET(DT_GPIO_CTLR_BY_IDX(n, gpios, 0)),                   \
 			.pin = DT_GPIO_PIN_BY_IDX(n, gpios, 0),                                    \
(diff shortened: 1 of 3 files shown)


alexapostolu:
Could this be a native sim issue? Maybe I'm looking at this wrong but `smbus.bus` seems to be an i2c device since it's being initialized using the `SMBUS_DT_SPEC_GET` macro, and `SMBUS_DT_SPEC_GET` uses [`DT_BUS`](https://docs.zephyrproject.org/latest/doxygen/html/group__devicetree-generic-bus.html#ga1082d31ac2dafdf9e085d4c23f2169dc) which points to i2c

```
&i2c0 {
	status = "okay";

	smbus_target0: smbus@a {
		status = "okay";
		compatible = "zephyr,smbus_target";
		reg = <0xa>;
	};
};
```

jgrowdenTT:
... That would be even more catastrophic (which I could explain), but no, what I have here should be right.

I'll point out  that you're looking at the SMC DTS... We need to look at the DMC side. 

Firstly. We understand that `INIT_CHIP` is running on each `chip` PHANDLE in `chips`, which is calling `BH_ARC_INIT` on each `arc` PHANDLE in turn.

i.e. The `chip` here in `chips`
`tt_blackhole_tt_blackhole_dmc.dts`
```
	chips {
		compatible = "tenstorrent,bh-chips";
		chips = <&chip0>;
		primary = <0>;
	};
```

Points to this `chip` here, which contains an `arc`
`tt_blackhole_dmc_dtsi`
```
chip0: chip0 {
....
arc = <&chip0_arc>;
};
```

Where the thing that `arc` points to is defined as being on the smbus.
```
&smbus3 {
	status = "okay";
	pinctrl-0 = <&i2c3_scl_pa7 &i2c3_sda_pb4>;
	pinctrl-names = "default";

	chip0_arc: bh_arc@A {
		compatible = "tenstorrent,bh-arc";
		reg = <0xA>;
	};
};
```

So; `BH_ARC_INIT(n)` is getting `n==bh_arc@a`. Hopefully not contentious.

But then `DT_BUS(bh_arc@a)` will resolve to the bus that `bh_arc` lives on, not the bus which `smbus3` lives on. And from the device tree, we know the bus the bh_arc@a lives on is `smbus3`

Which we can see in devicetree_generated
`#define DT_N_S_smbus3_S_bh_arc_a_BUS DT_N_S_smbus3`

alexapostolu (approved):
I see, makes sense

PR #1267, opened 2026-06-11
ci: metal: update container version

afongTT:
Update metal container version to v0.72.0, latest stable.


afongTT:
Nevermind, @alexapostolu is already working on this in #1230.

PR #1275, opened 2026-06-16
lib: bh_arc: Apply doppler throttler when aiclk rails

jgrowdenTT:
Apply the doppler nops in the case where we go above the TDP limit, and our AICLK is as low as possible.

At the 1/0 white marker, doppler nops were enabled as a response to me artificially limiting the TDP limit to 60.

At the 0/0 white marker, doppler nops were disabled as a response to me restoring the TDP limit to 150.

The other changes necessary to make this trace were 

- disable doppler on the p150
- TDP trace

image

resolves: SYS-4495

### DIFF
--- a/lib/tenstorrent/bh_arc/throttler.c
+++ b/lib/tenstorrent/bh_arc/throttler.c
@@ -14,6 +14,7 @@
 #include <tenstorrent/msgqueue.h>
 #include "cm2dm_msg.h"
 #include <zephyr/drivers/misc/bh_fwtable.h>
+#include <zephyr/tracing/tracing.h>
 #include "telemetry_internal.h"
 #include "telemetry.h"
 #include "noc2axi.h"
@@ -165,6 +166,7 @@ static void BroadcastKernelThrottleState(void)
 	const uint8_t kNocTlb = 1;
 
 	if (tensixes_enabled) {
+		sys_trace_named_event("kernel_throttle", throttle_counter & 1, 0);
 		NOC2AXITensixBroadcastTlbSetup(kNocRing, kNocTlb, kKernelThrottleAddress,
 					       kNoc2AxiOrderingStrict);
 		NOC2AXIWrite32(kNocRing, kNocTlb, kKernelThrottleAddress, throttle_counter);
--- a/lib/tenstorrent/bh_arc/throttler.c
+++ b/lib/tenstorrent/bh_arc/throttler.c
@@ -345,6 +345,7 @@ static void UpdateDoppler(const TelemetryInternalData *telemetry)
 void CalculateThrottlers(void)
 {
 	TelemetryInternalData telemetry_internal_data;
+	enum aiclk_arb_min arb;
 
 	ReadTelemetryInternal(1, &telemetry_internal_data);
 
@@ -355,6 +356,20 @@ void CalculateThrottlers(void)
 		UpdateThrottler(kThrottlerFastTDC, telemetry_internal_data.vcore_current);
 		UpdateThrottler(kThrottlerTDC, telemetry_internal_data.vcore_current);
 		UpdateThrottler(kThrottlerBoardPower, GetInputPower());
+
+		float current_power = telemetry_internal_data.vcore_power;
+		float tdp_limit = throttler[kThrottlerTDP].limit;
+
+		bool start_nops = GetAiclkTarg() == GetAiclkFmin() && current_power > tdp_limit;
+		bool stop_nops = GetAiclkTarg() == get_aiclk_effective_arb_min(&arb) &&
+				 current_power < tdp_limit;
+
+		bool new_kernel_nops_enabled = ((kernel_nops_enabled || start_nops) && !stop_nops);
+
+		if (new_kernel_nops_enabled != kernel_nops_enabled) {
+			kernel_nops_enabled = new_kernel_nops_enabled;
+			SendKernelThrottlingMessage(kernel_nops_enabled);
+		}
 	}
 
 	UpdateThrottler(kThrottlerThm, telemetry_internal_data.asic_temperature);


lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -355,6 +356,19 @@ void CalculateThrottlers(void)
 		UpdateThrottler(kThrottlerFastTDC, telemetry_internal_data.vcore_current);
 		UpdateThrottler(kThrottlerTDC, telemetry_internal_data.vcore_current);
 		UpdateThrottler(kThrottlerBoardPower, GetInputPower());
+
+		float current_power = telemetry_internal_data.vcore_power;
+		float power_limit = throttler[kThrottlerTDP].limit;
+
+		bool start_nops = GetAiclkTarg() == GetAiclkFmin() && current_power > power_limit;
+		bool stop_nops = GetAiclkTarg() == GetAiclkFmax() && current_power < power_limit;
```
jgrowdenTT: One issue is that this leaves on doppler throttling on when I force kill a model, because the AICLK never got to 1350 before power msg set the freq to 800.

It's likely that, upon re-running a model, the doppler nops will be released since you'd expect AICLK to raise back to 1350 after some period of being idle, but it's tough to say with the other arbiters in the way.

As a proposed fix, we could compare our targ aiclk to what the min arbiters are limiting us too.

`GetAiclkTarg() == get_aiclk_effective_arb_min(&arb)`

alexapostolu:
Could you add this to the release notes and add "resolves: jira#" if there is one

lib/tenstorrent/bh_arc/throttler.c
tcolacoTT: This implementation would imply the order of events on recovering from the throttling are:

- AICLK begins ramping to Fmax
- send broadcast to disable the nops only if AICLK reaches max (and power limit throttler not hitting)

I think my generic questions around this would be:

- is this how it was implemented for p150 (only disable NOPS once clock return to Fmax)?
- Is this how we **want** to implement it? in my mind we'd peel off the NOPs first and then ramp up the AICLK but it's a matter of interpretation of what's "best"

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -355,6 +356,19 @@ void CalculateThrottlers(void)
...
+
+		bool new_kernel_nops_enabled = ((kernel_nops_enabled || start_nops) && !stop_nops);
+
+		if (new_kernel_nops_enabled != kernel_nops_enabled) {
+			kernel_nops_enabled = new_kernel_nops_enabled;
+			SendKernelThrottlingMessage(kernel_nops_enabled);
```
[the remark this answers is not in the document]
jgrowdenTT: I can't re-use the whole func, I think, without grabbing onto existing logic that's out of the current spec.

I could refactor out just this part, with some parameterization for handling the t2/t3 critical logic I suppose.

lib/tenstorrent/bh_arc/throttler.c
jgrowdenTT: - is this how it was implemented for p150 (only disable NOPS once clock return to Fmax)?

[Yep, see here.](https://github.com/tenstorrent/tt-system-firmware/blob/ebd366d587b371fd206ef03a2b769e537e777489/lib/tenstorrent/bh_arc/throttler.c#L329-L340)

- Is this how we want to implement it? in my mind we'd peel off the NOPs first and then ramp up the AICLK but it's a matter of interpretation of what's "best"

The suggestion sounds like it'll cause a loop of us sending the kernel message over and over. Every time you release the NOPs button, you'll incur a power jump, causing AICLK to floor in response, causing the NOPs button to be re-applied. I should be able to graph this as well with a toy FW, if you'd like.

psethiTT: @tcolacoTT adding some thoughts here

- Yes this is how it was implemented for p150. 
- I believe one of the reasons for implementing it this way ie to increase AICLK before stopping NOPs was to let NOC data transfer (that also runs on AICLK) still run at high AICLK while Tensix remains throttled. Although, we have not run any experiments to characterize perf for either approach against real WLs. 

Question - do we know what's the current level of throttling applied by LLK?

tcolacoTT: Thanks James + Parth!

- I think this implementation makes sense for the current problem statement. We'll have finer grain control with AICLK so we'd ideally stabilize around some AICLK + "NOP enabled" to keep to TDP at its limit and eventually restore normal operation once things are okay
- @jgrowdenTT don't think we need the toggle NOP on/off test at this moment it'll be quite bouncy from a control perspective
- if/when we go down the path of these being 2 independent controllers that can throttle power separate conversation
- @psethiTT Radomir's current implementation always applies level 5 throttle (33%) 


@jgrowdenTT before this PR goes in can PPA team get a test FW to see how it behaves on galaxy running the wokrloads/trying to set the power limits?

FYI @mpnair-TT @dbbharatulaTT @dsprutovTT

jgrowdenTT: @tcolacoTT If acceptable, you can take the artifacts from this PR.

E.g.
https://github.com/tenstorrent/tt-system-firmware/actions/runs/27635987590/artifacts/7674922051

If not acceptable, I can provide a dev build myself. Advise what source/conf changes you want and I'd be happy to oblige. 

@alayander FYI no merge until @tcolacoTT and company says so

mpnair-TT: @jgrowdenTT thanks James. Will take this bundle and test. Hope I don't have to do any more changes to this.

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -355,6 +357,19 @@ void CalculateThrottlers(void)
 		UpdateThrottler(kThrottlerFastTDC, telemetry_internal_data.vcore_current);
 		UpdateThrottler(kThrottlerTDC, telemetry_internal_data.vcore_current);
 		UpdateThrottler(kThrottlerBoardPower, GetInputPower());
+
+		float current_power = telemetry_internal_data.vcore_power;
+		float tdp_limit = throttler[kThrottlerTDP].limit;
+
+		bool start_nops = GetAiclkTarg() == GetAiclkFmin() && current_power > tdp_limit;
```
jgrowdenTT, line 365: @afongTT @alewycky-tenstorrent The current convo here was good, but I don't want to lose sight of the issue I brought up.

How do we feel about changing `GetAiclkFmax` to getting the min arbiter limit?  I.e. 800 on idle, 1350 on busy. I believe this solves the issue after force killing the model, we don't reset the NOPs state to off (because we never made it back to 1350.)

lib/tenstorrent/bh_arc/throttler.c
jgrowdenTT: @mpnair-TT and others (@psethiTT, right?) have had concerns on shutting off NOPing at Fmax instead of something else.

I can imagine a bunch of impls that would probably work. "aiclk at 1000" where 1000 is KCONFIG, fwtable, arc_msg, adjustable, or any combo thereof. Flexible, but more knobs aren't always good. 

"aiclk sustained above 1000 for some number of iterations" would account for our oscillation patterns I'm seeing.

The current implmentation was suggested by @alewycky-tenstorrent and also matches our p150a doppler implementation. 

I could sugget, if reviewers are satisfied, running this implementation for one release, but looking for improvements in the near future.  Of course if reviewers are not satisfied I'm happy to implement whatever feesible implementation now. Please advise. Thanks for your time everyone.

doc/release/release-notes-19.11.md
```diff
@@ -10,6 +10,10 @@ We are pleased to announce the release of TT System Firmware version 19.11.0 
 
 ## Blackhole
 
+### Kernel Throttling on TDP Limit
```
jgrowdenTT: Wrong place. I'm taking out the commit.

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -355,6 +357,19 @@ void CalculateThrottlers(void)
...
+		bool stop_nops = GetAiclkTarg() == GetAiclkFmax() && current_power < tdp_limit;
```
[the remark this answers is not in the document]
jgrowdenTT: I am pretty certain this will cause doppler throttling to be spammed.

lib/tenstorrent/bh_arc/throttler.c
afongTT: I can't think of any issues with switching to using effective min arb limit.

jgrowdenTT: @afongTT implemented. traces show, when we come back out of idle, we re-apply no-matmul-throttling.

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -355,6 +358,20 @@ void CalculateThrottlers(void)
 		UpdateThrottler(kThrottlerFastTDC, telemetry_internal_data.vcore_current);
 		UpdateThrottler(kThrottlerTDC, telemetry_internal_data.vcore_current);
 		UpdateThrottler(kThrottlerBoardPower, GetInputPower());
+
+		float current_power = telemetry_internal_data.vcore_power;
+		float tdp_limit = throttler[kThrottlerTDP].limit;
+
+		bool start_nops = GetAiclkTarg() == GetAiclkFmin() && current_power > tdp_limit;
```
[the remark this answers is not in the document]
jgrowdenTT: Nice border case, but I think we want to leave it at fmin.  If the host msgs us saying our FMin is 1000, why are we enabling doppler nops to try to save power? 

@afongTT? @tcolacoTT?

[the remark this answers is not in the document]
tcolacoTT: Fmin is hidden behind SMC_MSG_CHARACTERISATION so I agree this is an ignorable edge case as you mentioned James.

We're really only looking to enforce the chip + system power limit for cases where the infrastructure is limited.

jgrowdenTT:
@alayander FYI @mpnair-TT Has signed off.

If we can get some tt-system-firmware reviewers to stamp, it'd be appreciated. @afongTT @alewycky-tenstorrent @alexapostolu

Any and all feedback is still welcome of course, and I can run whatever toy implementations we want. Thank you!

afongTT (approved):

alayander:
There's a new release notes markdown for 19.12 on main that we can add the release note to, but that can be down in a follow up PR if preferred

PR #1282, opened 2026-06-18
lib: bh_arc: tt_shell: add msgqueue command

jgrowdenTT:
Add a generic msgqueue command. This is useful for situations where we have shell but don't have a link to CSM; e.g. native sim/nucleo DMC with emulated SMC.

Tests

On device:
```
jg-dev:~$ tt msg 0xc0
rsp[0] = 0x00000000
rsp[1] = 0x00000001
rsp[2] = 0x00000000
rsp[3] = 0x00000000
rsp[4] = 0x00000000
rsp[5] = 0x00000000
rsp[6] = 0x00000000
rsp[7] = 0x00000000
```
Also works in native_sim/nucleo DMC with emulated SMC.

resolves: SYS-4594

### DIFF
--- a/lib/tenstorrent/bh_arc/tt_shell.c
+++ b/lib/tenstorrent/bh_arc/tt_shell.c
@@ -6,16 +6,35 @@
 
 #include <zephyr/logging/log.h>
 #include <zephyr/shell/shell.h>
+#include <errno.h>
+#include <limits.h>
 #include <stdlib.h>
 
 #include <tenstorrent/bh_power.h>
+#include <tenstorrent/msgqueue.h>
 
 #include "telemetry.h"
 #include "smbus_target.h"
 #include "gddr.h"
 #include "asic_state.h"
 #include "noc_init.h"
+
 LOG_MODULE_REGISTER(tt_shell, CONFIG_LOG_DEFAULT_LEVEL);
+
+static int parse_u32_arg(const char *arg, uint32_t *value)
+{
+	char *endptr;
+	unsigned long parsed;
+
+	errno = 0;
+	parsed = strtoul(arg, &endptr, 0);
+	if (errno != 0 || endptr == arg || *endptr != '\0' || parsed > UINT32_MAX) {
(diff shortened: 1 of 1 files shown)


lib/tenstorrent/bh_arc/tt_shell.c
```diff
@@ -148,12 +167,51 @@ static int telem_handler(const struct shell *sh, size_t argc, char **argv)
...
+		request.data[i - 1U] = parsed;
+	}
+
+	ret = msgqueue_request_push(0, &request);
+	if (ret != 0) {
+		shell_error(sh, "Failed to queue request (%d)", ret);
+		return ret;
+	}
+
+	process_message_queues();
```
alexapostolu, line 193: We push to the first msgqueue but process all of them. Are there any side effects to this?

jgrowdenTT: Good question, thanks.

- In the upcoming native_sim/nucleo boards, no, there's no other producers. 
- Prod/release doesn't enable tt_shell.

For dev builds for device who enable the shell. PCIE and shell will compete, yeah.

I think there's two issues.

A) race if someone was already process_message_queues - solve by mutex
B) KMD theoretically owns queue 0, not us.  - Tougher to solve. 

Both points are solved if I just don't enable this command on SMC ARC FW (`CONFIG_ARC` latch is convenient for this PR, but I have a better KCONFIG coming up). 

I also think that both points are kinda rare to happen unless you're in scenarios where you really shouldn't be banging out messages on the shell, and besides, you're in a dev build lol. But I'm completely fine scrapping this on device since if we've got vuart shell, we've got luwen/UMD/tt-dal based messaging anyways.

Let me know your thoughts. Thanks for the feedback Alex.

alexapostolu: Yea this makes sense, I think it's fine if we just merge it as is for now

alexapostolu (approved):

PR #1283, opened 2026-06-18
lib: tenstorrent: bh_arc: enable process based vf curve for p300

alayander:
Enable use of process based vf curve for p300. 

Validation details of the new curve can be found in the associated Jira ticket.

Resolves: SYS-4586


alexapostolu (dismissed):

alexapostolu:
This could also go into the release notes

alayander:
Pushed to add release note

alexapostolu (approved):

PR #1291, opened 2026-06-19
lib: bh_arc: add gddr temperature readings to telemetry internal

alayander:
Add gddr temperature readings to telemetry internal data structure by directly reading temperature registers from MRISC L1.

When tracing this change, `ReadTelemetryInternal` took 289-312 μs.
Whereas before this change `ReadTelemetryInternal` took 281-304μs.
Tracing just the added gddr temperature readings took ~6μs, aligning with the above measurements.

Resolves: SYS-4576

### DIFF
--- a/lib/tenstorrent/bh_arc/gddr.c
+++ b/lib/tenstorrent/bh_arc/gddr.c
@@ -14,6 +14,8 @@
 #include "reg.h"
 #include "status_reg.h"
 
+#include <stddef.h>
+
 #include <tenstorrent/bh_power.h>
 #include <tenstorrent/msgqueue.h>
 #include <tenstorrent/post_code.h>
@@ -58,6 +60,7 @@ LOG_MODULE_REGISTER(gddr, CONFIG_TT_APP_LOG_LEVEL);
 static const struct device *const fwtable_dev = DEVICE_DT_GET(DT_NODELABEL(fwtable));
 
 static struct gddr_bist_info gddr_bist;
+static uint8_t gddr_telemetry_version_ok;
 
 struct gddr_bist_info get_gddr_bist_info(void)
 {
@@ -221,6 +224,57 @@ static uint32_t GetDramMask(void)
 		dram_mask &= tt_bh_fwtable_get_fw_table(fwtable_dev)->dram_table.dram_mask;
 	}
 	return dram_mask;
+}
+
+/* Read the top and bottom DRAM die temperatures (degrees Celsius) for a single GDDR instance. */
+static int read_gddr_temperature(uint8_t gddr_inst, struct gddr_inst_temp *temp)
+{
+	if (gddr_inst >= NUM_GDDR || temp == NULL) {
+		return -EINVAL;
+	}
+	if (!IS_BIT_SET(gddr_telemetry_version_ok, gddr_inst)) {
+		return -ENOTSUP;
+	}
+	/* dram_temperature_top and dram_temperature_bottom are adjacent uint16_t fields that */
+	/* share a single 32-bit word: [15:0] = top, [31:16] = bottom */
+	uint32_t temps = MriscL1Read32(
+		gddr_inst,
+		GDDR_TELEMETRY_TABLE_ADDR + offsetof(gddr_telemetry_table_t, dram_temperature_top));
+	temp->top = temps & 0xff;
+	temp->bottom = (temps >> 16) & 0xff;
+	return 0;
+}
+
+int get_gddr_temps(struct gddr_temps *temps)
+{
+	if (temps == NULL) {
+		return -EINVAL;
+	}
+
+	*temps = (struct gddr_temps){0};
+	uint32_t dram_mask = GetDramMask();
+
+	int ret = 0;
+
+	for (int i = 0; i < NUM_GDDR; i++) {
+		if (!IS_BIT_SET(dram_mask, i)) {
+			continue;
+		}
+
+		int rc = read_gddr_temperature(i, &temps->inst[i]);
+
+		if (rc < 0) {
+			LOG_WRN_ONCE("Failed to read GDDR %d temperature while updating telemetry",
+				     i);
+			ret = rc;
+			continue;
+		}
+
+		temps->max_temp =
+			MAX(temps->max_temp, MAX(temps->inst[i].top, temps->inst[i].bottom));
+	}
+
+	return ret;
 }
 
 static int check_mrisc_busy(uint8_t gddr_inst)
@@ -462,10 +516,24 @@ SYS_INIT_APP(InitMrisc);
 
 static int CheckGddrTraining(uint8_t gddr_inst, k_timepoint_t timeout)
 {
+	gddr_telemetry_version_ok &= ~BIT(gddr_inst);
+
 	do {
 		uint32_t poll_val = MriscRegRead32(gddr_inst, MRISC_INIT_STATUS);
 
 		if (poll_val == MRISC_INIT_FINISHED) {
+			uint32_t version =
+				MriscL1Read32(gddr_inst, GDDR_TELEMETRY_TABLE_ADDR +
+								 offsetof(gddr_telemetry_table_t,
+									  telemetry_table_version));
+
+			if (version != GDDR_TELEMETRY_TABLE_T_VERSION) {
+				LOG_ERR("%s[%d]: version mismatch: %d (expected %d)",
+					"GDDR telemetry table", gddr_inst, version,
+					GDDR_TELEMETRY_TABLE_T_VERSION);
+				return -ENOTSUP;
+			}
+			gddr_telemetry_version_ok |= BIT(gddr_inst);
 			return 0;
 		}
 		if (poll_val == MRISC_INIT_FAILED) {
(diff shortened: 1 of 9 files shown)


lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -293,14 +282,46 @@ static void UpdateGddrTelemetry(void)
...
+	for (int i = 0; i < NUM_GDDR; i++) {
+		uint16_t top, bottom;
+
+		if (!IS_BIT_SET(tile_enable.gddr_enabled, i)) {
+			continue;
+		}
+		if (read_gddr_temperature(i, &top, &bottom) < 0) {
+			LOG_WRN_ONCE("Failed to read GDDR temperature while updating telemetry");
+			continue;
+		}
```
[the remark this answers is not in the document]
alayander: This follows the convention used by other gddr functions

jgrowdenTT:
Mmk. I imagine this is fast enough that we can just run it full tilt without worrying about if its worth it.

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -144,6 +146,27 @@ int read_gddr_telemetry_table(uint8_t gddr_inst, gddr_telemetry_table_t *gddr_te
 	return 0;
 }
 
+int read_gddr_temperature(uint8_t gddr_inst, uint16_t *top, uint16_t *bottom)
+{
+	/* Check that version matches expectation */
+	uint32_t version = MriscL1Read32(
+		gddr_inst, GDDR_TELEMETRY_TABLE_ADDR +
+				   offsetof(gddr_telemetry_table_t, telemetry_table_version));
+	if (version != GDDR_TELEMETRY_TABLE_T_VERSION) {
```
jgrowdenTT: Do we have to do this? I feel like just once at init is enough right?

Or are we trying to support hotloads of MRISC FW?

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -293,14 +282,46 @@ static void UpdateGddrTelemetry(void)
...
+
+		/* DDR_x_y_TEMP:
```
jgrowdenTT: Sorry. "while you're here"...

Attach this instead to the doxygen for https://docs.tenstorrent.com/tt-system-firmware/doxygen/group__telemetry__tags.html#ga6abf62859c0e87da7baee6d16e942c9e

That way when people ask what the format is, we can say RTFM and get a gold star.

(doxygen every of GDDR 01, 23, 45, 67 pls and thanks.)

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -293,14 +282,46 @@ static void UpdateGddrTelemetry(void)
...
+		int shift_val = (i % 2) * 16;
+
+		temperature[i / 2] |=
+			((top & 0xff) << (8 + shift_val)) | ((bottom & 0xff) << shift_val);
+	}
+
+	for (int i = 0; i < NUM_GDDR / 2; i++) {
+		telemetry[TAG_GDDR_0_1_TEMP + i] = temperature[i];
+	}
+	telemetry[TAG_MAX_GDDR_TEMP] = GetMaxGDDRTemp();
```
jgrowdenTT: Isn't this now inconsistent with how we deal with, for example, vcore power?

I think that updates internally every at most every 1ms, but only update reported telem every 100ms. Right?

lib/tenstorrent/bh_arc/gddr.c
alayander: This is mirrored behaviour from `UpdateGddrTelemetry` from the regular telemetry loop. 

I agree the only case where this would be necessary is if trying to support hotloads of MRISC FW, I'll confirm whether this is the case.

lib/tenstorrent/bh_arc/telemetry.c
alayander: Confirmed from the build artifact that the doxygen formatted properly.

lib/tenstorrent/bh_arc/telemetry.c
alayander: Yes! - Thanks for catching that

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -293,27 +282,40 @@ static void UpdateGddrTelemetry(void)
...
-		telemetry[TAG_GDDR_0_1_TEMP + i] = temperature[i];
 		telemetry[TAG_GDDR_0_1_CORR_ERRS + i] = corr_errs[i];
 	}
 	telemetry[TAG_GDDR_UNCORR_ERRS] = uncorr_errs;
 	telemetry[TAG_GDDR_STATUS] = status;
 	telemetry[TAG_GDDR_SPEED] = speed;
 }
 
-int GetMaxGDDRTemp(void)
+uint32_t PackGddrTemperatures(uint32_t *packed)
```
alexapostolu: Would it be better is PackGddrTemperatures was moved to gddr.c? I think it would be better in terms of dependancy directions since now telemetry_internal.c won't have to depend on telemetry.c

lib/tenstorrent/bh_arc/gddr.c
alayander: @afongTT do you know if we are trying to support hotloads of MRISC FW?

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -144,6 +146,56 @@ int read_gddr_telemetry_table(uint8_t gddr_inst, gddr_telemetry_table_t *gddr_te
...
+	}
+	/* dram_temperature_top and dram_temperature_bottom are adjacent uint16_t fields that */
+	/* share a single 32-bit word: [15:0] = top, [31:16] = bottom */
+	uint32_t temps = MriscL1Read32(
+		gddr_inst,
+		GDDR_TELEMETRY_TABLE_ADDR + offsetof(gddr_telemetry_table_t, dram_temperature_top));
+	*top = temps & 0xFFFF;
+	*bottom = (temps >> 16) & 0xFFFF;
+	return 0;
+}
```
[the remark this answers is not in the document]
alayander: This is currently being discussed in an above conversation

lib/tenstorrent/bh_arc/gddr.c
afongTT: No, I think we'd be fine to just check this once, maybe right after we `CheckGddrTraining()`.

alayander:
After removing the version check on each temperature update, the new timings are as follows:
`ReadTelemetryInternal` takes 289-312 μs.
Just the gddr temperature readings took ~6μs.

Updated the description with this info.

lib/tenstorrent/bh_arc/telemetry_internal.c
```diff
@@ -54,6 +55,9 @@ void ReadTelemetryInternal(int64_t max_staleness, TelemetryInternalData *data)
 			internal_data.vcore_current * internal_data.vcore_voltage * 0.001f;
 		internal_data.asic_temperature = avg_tmp;
 
+		internal_data.gddr_max_temperature =
+			pack_gddr_temperatures(internal_data.gddr_temperatures);
```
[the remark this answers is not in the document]
alayander: I believe there shouldn't be a concurrency issue, since I believe they are all run on a single system workqueue.

alexapostolu (dismissed):
I think this looks good now. I'll defer to @jgrowdenTT for final approval

jgrowdenTT (approved):
I think this is fine but I'll leave some comments...

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -144,6 +147,53 @@ int read_gddr_telemetry_table(uint8_t gddr_inst, gddr_telemetry_table_t *gddr_te
...
+
+uint32_t pack_gddr_temperatures(uint32_t *packed)
```
jgrowdenTT: This seemed kinda weird to me at first glance. Potentially better is if we pass in a struct instead of returning part and by-ref-returning the other.

so like.
```
int get_gddr_temps(struct gddr_tmp* tmp)  {
...
}
```
where:

```
struct inst_tmp {
    uint8_t top;
    uint8_t bottom;
};

struct gddr_tmp {
uint32_t max_temp
struct inst_tmp[NUM_INST]
}
```

What this should do:

- It should make the top and botom masking/shifting stuff disappear. 
- It should allow us to scale stuff if we wanted (e.g.) max top, max bottom
- It should allow us to propogate the error code from read_gddr_temp out of this function, if necessary (though log and continue is fine.)

Though, I think my proposal pops an issue with getting the data out to telem, where you might have to start shifting stuff again. If you want, play around, but I don't think this is gating.

jgrowdenTT (dismissed):

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -144,6 +147,59 @@ int read_gddr_telemetry_table(uint8_t gddr_inst, gddr_telemetry_table_t *gddr_te
 	return 0;
 }
 
+int read_gddr_temperature(uint8_t gddr_inst, uint16_t *top, uint16_t *bottom)
```
jgrowdenTT: You can just pass this a pointer to the structure that has top and bottom now, instead of 2 different pointers, right?

alexapostolu (approved):

PR #1294, opened 2026-06-22
Recovery split

alexapostolu:
Separate recovery firmware from fwtable. An alternative that was considered was to just use #ifdef everywhere for recovery, but I thought this method of separating recovery form mission on a cmake build level is cleaner

resolves: SYS-4642

### DIFF
--- a/app/smc/src/main.c
+++ b/app/smc/src/main.c
@@ -23,14 +23,18 @@
 #include <zephyr/kernel.h>
 #include <zephyr/logging/log.h>
 #include <zephyr/drivers/watchdog.h>
-#include <zephyr/drivers/misc/bh_fwtable.h>
 #include <zephyr/storage/flash_map.h>
 #include <zephyr/dfu/mcuboot.h>
+#ifdef CONFIG_BH_FWTABLE
+#include <zephyr/drivers/misc/bh_fwtable.h>
+#endif
 
 LOG_MODULE_REGISTER(main, CONFIG_TT_APP_LOG_LEVEL);
 
 static const struct device *const wdt0 = DEVICE_DT_GET(DT_NODELABEL(wdt0));
+#ifdef CONFIG_BH_FWTABLE
 static const struct device *const fwtable_dev = DEVICE_DT_GET(DT_NODELABEL(fwtable));
+#endif
 
 BUILD_ASSERT(PARTITION_EXISTS(cmfw), "cmfw fixed-partition does not exist");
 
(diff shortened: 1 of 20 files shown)


app/smc/src/main.c
```diff
@@ -24,14 +24,18 @@
 #include <zephyr/kernel.h>
 #include <zephyr/logging/log.h>
 #include <zephyr/drivers/watchdog.h>
-#include <zephyr/drivers/misc/bh_fwtable.h>
 #include <zephyr/storage/flash_map.h>
 #include <zephyr/dfu/mcuboot.h>
+#if !defined(CONFIG_TT_SMC_RECOVERY)
```
jgrowdenTT: Correct should be based on `BH_FWTABLE`

lib/tenstorrent/bh_arc/chip_info.h
```diff
@@ -0,0 +1,69 @@
+/*
```
jgrowdenTT, line 1: Shouldn't we instead have a fwtable device that's a stub-out for recovery?

I think most of the changes go away if we go this route, right?

Nah... While this would work, and I think it's better, fwtable isn't using DEVICE_API. So we can't vtable to a static implemenation. But I suppose you could keep the same contract and then just feed that thru the API pointer instead of always going thru bh_fwtable.

Worth playing with IMO.

lib/tenstorrent/bh_arc/telemetry_mission.c
```diff
@@ -0,0 +1,344 @@
+/*
+ * Copyright (c) 2024 Tenstorrent AI ULC
```
jgrowdenTT: Prefer to avoid this spit up, if possible.

What are we trying to solve? If we don't want init and the timers hooked up, just latch those around an ifdef for recovery and mark the others as unused.

lib/tenstorrent/bh_arc/harvesting_mission.c
```diff
@@ -0,0 +1,261 @@
+/*
```
jgrowdenTT: Instead of moving everything, I think ifdefs inside harvesting will accomplish what we want.

lib/tenstorrent/bh_arc/CMakeLists.txt
```diff
@@ -64,11 +34,51 @@ zephyr_library_sources(
 # zephyr-keep-sorted-stop
 )
 
-zephyr_library_add_dependencies(nanopb_generated_headers)
+zephyr_library_sources_ifdef(CONFIG_TT_SMC_RECOVERY chip_info_static.c)
```
alayander: Would it make sense to instead have this gated on 
`zephyr_library_sources_ifndef(CONFIG_BH_FWTABLE chip_info_static.c)`?

lib/tenstorrent/bh_arc/CMakeLists.txt
```diff
@@ -64,11 +34,51 @@ zephyr_library_sources(
...
 endif()
+
+if(NOT CONFIG_TT_SMC_RECOVERY)
+  zephyr_library_named(bh_arc_mission)
+  zephyr_library_include_directories_ifdef(CONFIG_TT_BH_ARC ${ZEPHYR_CURRENT_MODULE_DIR}/include)
+
+  zephyr_library_sources(
+    aiclk_ppm.c
+    bh_power.c
+    chip_info_fwtable.c
```
alayander: Would it make sense to instead have this gated on
`zephyr_library_sources_ifdef(CONFIG_BH_FWTABLE chip_info_fwtable.c)`?

lib/tenstorrent/bh_arc/chip_info.h
```diff
@@ -0,0 +1,56 @@
...
+
+/* PCIe operating mode. Values intentionally match the protobuf
+ * FwTable_PciPropertyTable_PcieMode enum so the fwtable backend can pass them
+ * through unchanged.
+ */
+typedef enum {
+	BH_PCIE_MODE_DISABLED = 0,
+	BH_PCIE_MODE_EP = 1,
+	BH_PCIE_MODE_RP = 2,
+} bh_pcie_mode_t;
```
alayander, line 25: Could be helpful to add some BUILD_ASSERTs to ensure that these values will always match.

lib/tenstorrent/bh_arc/chip_info.h
jgrowdenTT: The other issue is that recovery doesn't have a dedicated DTS.

include/zephyr/drivers/misc/bh_fwtable.h
```diff
@@ -14,12 +14,16 @@ extern "C" {
...
+ * The generated nanopb headers only exist when the bh_fwtable driver is built.
+ * SMC recovery builds disable the driver (see drivers/misc/bh_fwtable/Kconfig),
+ * but still pull in this header transitively for the plain chip enums below
+ * (e.g. PcbType via regulator.h).
  */
+#ifdef CONFIG_BH_FWTABLE
 #include "fw_table.pb.h"
 #include "flash_info.pb.h"
 #include "read_only.pb.h"
+#endif
```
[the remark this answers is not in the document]
jgrowdenTT: @alexapostolu I think this is good feedback, even if it's not currently bricking the build. Up to you.

jgrowdenTT (dismissed):

jgrowdenTT:
Should have the SYS. Or maybe we talk about dropping that requirement.

However, would be appreciated if you could outline your tests in the PR description. Thanks Alex!

alexapostolu:
I added a native sim test, and I tested manually by flashing bad firmware to force recovery and checked PCIE still enumerated

And I added the SYS

include/zephyr/drivers/misc/bh_fwtable.h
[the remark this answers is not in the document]
alexapostolu: Fixed

afongTT:
> I added a native sim test, and I tested manually by flashing bad firmware to force recovery and checked PCIE still enumerated

I think tt-flash still doesn't support flashing out of recovery, right? (I don't know how they'd get the right board type... maybe they can do some searching around SPI for the board ID?...)

alexapostolu:
@afongTT I'm not sure about tt-flash, but I used blackhole-recovery to get out of recovery

alayander (dismissed):

alayander:
> I think tt-flash still doesn't support flashing out of recovery, right? (I don't know how they'd get the right board type... maybe they can do some searching around SPI for the board ID?...)

@afongTT is this something to address in this PR or is it good to merge?

jgrowdenTT (approved):

alayander (approved):

afongTT:
> > I think tt-flash still doesn't support flashing out of recovery, right? (I don't know how they'd get the right board type... maybe they can do some searching around SPI for the board ID?...)
> 
> @afongTT is this something to address in this PR or is it good to merge?

Doesn't need to be addressed in this PR

PR #1295, opened 2026-06-22
.github refactor

alexapostolu:
Simplify .github workflows
* remove redundancy
* cleaner logs
* added workflow dispatch (it wasn't working before due to a small bug)

### DIFF
--- a/.github/ci_boards.json
+++ b/.github/ci_boards.json
@@ -1,62 +1,62 @@
-[
-  {
-    "board": "p100a",
-    "runs-on": "p100a-jtag",
-    "product": "bh",
-    "metal-image": "ghcr.io/tenstorrent/tt-metal/upstream-tests-bh:v0.68.0-dev20260317-16-gbb89a89897",
-    "metal-target": "blackhole",
-    "metal-options": "--cap-add SYS_MODULE --device /dev/tenstorrent --user=root",
-    "kmd_build": "2.8.0"
-  },
-  {
-    "board": "p150a",
-    "runs-on": "p150a-jtag",
-    "product": "bh",
-    "metal-image": "ghcr.io/tenstorrent/tt-metal/upstream-tests-bh:v0.68.0-dev20260317-16-gbb89a89897",
-    "metal-target": "blackhole",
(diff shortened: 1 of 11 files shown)


afongTT:
Do you want to test the labels on this PR?
- wh-tests
- long-tests
- metal-tests

alexapostolu:
@afongTT  Stress, Metal and WH appear to be running as normal

afongTT (approved):

.github/ci_boards.json
```diff
@@ -1,62 +1,62 @@
...
+      "board": "p300a",
+      "supports_smoke": true,
+      "metal-image": "ghcr.io/tenstorrent/tt-metal/upstream-tests-bh-p300:v0.68.0-dev20260317-16-gbb89a89897",
+      "metal-target": "blackhole_p300",
+      "arch-name": "blackhole",
+      "metal-options": "--cap-add SYS_MODULE --device /dev/tenstorrent --user=root",
+      "kmd_build": "2.8.0"
+    },
+    {
+      "board": "quietbox2",
```
psethiTT, line 38: is there a reason why QB2 and LB dont support smoke test? Have we intentionally left them out?

.github/workflows/hardware-long.yml
```diff
@@ -1,46 +1,45 @@
...
   cancel-in-progress: true
 
 jobs:
-  long-test-check:
+  # Single source of truth for whether the long tests should run. Downstream
+  # jobs gate on this via `needs` so the condition lives in exactly one place.
+  should-run:
+    if: >
+      github.event_name == 'schedule'
+      || github.event_name == 'workflow_dispatch'
```
psethiTT, line 21: Confirmed offline with Alex - before this change, workflow_dispatch was declared as a trigger but wasn't included in the gating job's if condition (only schedule and the long-tests label were). As a result, manually triggering the workflow had no effect. This appears to have been an oversight rather than a deliberate choice, so we've added workflow_dispatch to the condition to enable manual runs as intended.

psethiTT (approved):

PR #1306, opened 2026-06-25
drivers: misc: Efuse driver

jgrowdenTT:
Add an efuse driver.

Other than following zephyr style, this gives us a convenient hook to stub out at an emulated efuse instance.

tests:

I've added a test which verifies some telemetry settings match the efuse programmed data.


drivers/misc/bh_efuse/bh_efuse.c
```diff
@@ -0,0 +1,117 @@
...
+				 enum tt_bh_efuse_box_id box_id,
+				 uint32_t offset,
+				 uint32_t *value)
+{
+	ARG_UNUSED(dev);
+
+	if ((uint32_t)box_id >= TT_BH_EFUSE_BOX_ID_NUM) {
+		return -EINVAL;
+	}
+
```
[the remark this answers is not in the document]
jgrowdenTT: Hm. I think the MACRO already takes care of this.

drivers/misc/bh_efuse/Kconfig
```diff
@@ -0,0 +1,30 @@
...
+	help
+	  Enable the Tenstorrent Blackhole eFuse driver API.
+
+if TT_BH_EFUSE
+
+config TT_BH_EFUSE_INIT_PRIORITY
+	int "Tenstorrent Blackhole eFuse init priority"
+	default 90
+	help
+	  Set initialization priority for the eFuse driver.
```
alayander, line 17: Our pvt driver init makes calls to read from efuse and currently has init priority `CONFIG_SENSOR_INIT_PRIORITY` (which is also default 90 from upstream, I believe). So this could potentially cause issues during pvt init, right?

jgrowdenTT: Thanks! Dropped the prio.

drivers/misc/bh_efuse/bh_efuse.c
```diff
@@ -0,0 +1,114 @@
...
+#define EFUSE_CNTL_EFUSE_RD_CNTL_REG_DEFAULT 0x00000001u
+
+static int tt_bh_efuse_read_impl(const struct device *dev, enum tt_bh_efuse_access_type acc_type,
+				 enum tt_bh_efuse_box_id box_id, uint32_t offset, uint32_t *value)
+{
+	ARG_UNUSED(dev);
+
+	if ((uint32_t)box_id >= TT_BH_EFUSE_BOX_ID_NUM) {
+		return -EINVAL;
+	}
```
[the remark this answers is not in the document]
jgrowdenTT: No way this is right, yeah?

drivers/misc/bh_efuse/bh_efuse.c
```diff
@@ -0,0 +1,115 @@
...
+#include <zephyr/drivers/misc/bh_efuse.h>
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/sys/sys_io.h>
+
+LOG_MODULE_REGISTER(tt_bh_efuse, CONFIG_TT_BH_EFUSE_LOG_LEVEL);
+
+#if DT_NUM_INST_STATUS_OKAY(DT_DRV_COMPAT) > 0
+
+#define EFUSE_DFT0_MEM_BASE_ADDR          0x80040000u
```
alexapostolu, line 21: This isn't a functional change, but since it's a Zephyr driver we could move this base address into the devicetree

drivers/misc/bh_efuse/bh_efuse.c
```diff
@@ -0,0 +1,115 @@
...
+
+typedef union {
+	uint32_t val;
+	EFUSE_CNTL_EFUSE_RD_CNTL_reg_t f;
+} EFUSE_CNTL_EFUSE_RD_CNTL_reg_u;
+
+#define EFUSE_CNTL_EFUSE_RD_CNTL_REG_DEFAULT 0x00000001u
+
+static int tt_bh_efuse_read_impl(const struct device *dev, enum tt_bh_efuse_access_type acc_type,
+				 enum tt_bh_efuse_box_id box_id, uint32_t offset, uint32_t *value)
```
alexapostolu, line 53: Should we check the bounds on offset?

alexapostolu (approved):
Looks good, just minor non-blocking things

alayander (approved):

drivers/misc/bh_efuse/bh_efuse.c
jgrowdenTT: I'll address in a new PR since this is merged. Thanks Alex!

drivers/misc/bh_efuse/bh_efuse.c
jgrowdenTT: I think you're right. I was on the fence because the functional macros, I think, are taking care of bounds checking already, but bounds checking here makes it more standalone.

PR #1309, opened 2026-06-26
west.yml: Grab libpldm

jgrowdenTT:
Grab libpldm. Get a version slightly more recent that v0.16 since that version compiles with warnings.

tests:

This commit works to compile some pldm symbols inside the library that I need for nucleo work.

### DIFF
--- a/MAINTAINERS.yml
+++ b/MAINTAINERS.yml
@@ -569,3 +569,10 @@ Tenstorrent Platforms:
   labels: []
   description: |
     Maintained upstream.
+
+"West project: libpldm":
+  status: odd fixes
+  files: []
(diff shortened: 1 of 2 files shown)


alexapostolu (dismissed):

alexapostolu:
Do we need to "maintain" the maintainers file? Is it useful enough where we need to update it

jgrowdenTT:
Compliance will barf if it's not there :/

Not really a reason. It's now easy enough to silence the compliance checkers we don't care about. We could discuss it in FW meeting. For now, this turns our compliance green.

alayander (approved):

PR #1310, opened 2026-06-26
Revert #1289

alayander:
Revert #1289

### DIFF
--- a/app/smc/src/main.c
+++ b/app/smc/src/main.c
@@ -17,6 +17,7 @@
 #include <stdint.h>
 
 #include <app_version.h>
+#include <tenstorrent/msgqueue.h>
 #include <tenstorrent/post_code.h>
 #include <tenstorrent/sys_init_defines.h>
 #include <zephyr/init.h>
@@ -58,6 +59,8 @@ int main(void)
 		}
 	}
 #endif
+
+	init_msgqueue();
 
 #ifdef CONFIG_BH_FWTABLE
 	init_telemetry(APPVERSION);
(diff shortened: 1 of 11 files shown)


jgrowdenTT:
shell command can probably stay.

although i guess that means we need the kconfig. eh. just revert.

jgrowdenTT:
do you wanna label this for stress to run?

jgrowdenTT (dismissed):

app/smc/src/main.c
```diff
@@ -60,9 +61,24 @@ int main(void)
 #endif
 
 #ifdef CONFIG_BH_FWTABLE
-	init_telemetry(APPVERSION);
-	if (tt_bh_fwtable_get_fw_table(fwtable_dev)->feature_enable.fan_ctrl_en) {
-		init_fan_ctrl();
+	init_msgqueue();
```
afongTT, line 63: I think `init_msgqueue()` should be before the `#ifdef`.

Also not sure you need to bring `IS_ENABLED(CONFIG_TT_SMC_RECOVERY)` back. The `#ifdef CONFIG_BH_FWTABLE` should be enough?

alayander: Oh sorry, i thought that came from as a diff from this patch, will fix

app/smc/src/main.c
```diff
@@ -60,9 +61,24 @@ int main(void)
...
+		/* These timers are split out from their init functions since their work tasks have
+		 * i2c conflicts with other init functions.
+		 *
+		 * Note: The above issue would be solved by using Zephyr's driver model.
+		 */
+		StartTelemetryTimer();
+		if (dvfs_enabled) {
+			StartDVFSTimer();
+		}
+		StartGddrThermTripMonitor();
```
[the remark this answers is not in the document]
afongTT: oh yeah, copilot is right. I think in this file, you only needed to add `init_msqueue()` before `#ifdef CONFIG_BH_FWTABLE` (and the header include).

afongTT (approved):

PR #1313, opened 2026-06-27
doc: PLDM service design

jgrowdenTT:
Provide documentation for the PLDM service

Tests:

Docs build locally.

resolves: SYS-4735

### DIFF
--- a/doc/services/pldm/index.rst
+++ b/doc/services/pldm/index.rst
@@ -1,0 +1,641 @@
+.. _ttzp_services_pldm:
+
+PLDM Service
+============
+
+Overview
+--------
+
+The DMC PLDM service exposes platform sensor data to a BMC over
+MCTP-on-I3C. It is implemented as a Zephyr driver.
+
+Acronyms used in this document:
+
+.. list-table::
+   :header-rows: 1
+
+   * - Acronym
+     - Expansion
+     - Description
+   * - BMC
+     - Baseboard Management Controller
+     - Requester side that discovers PLDM capabilities and reads sensor values.
+   * - DAA
+     - Dynamic Address Assignment
+     - I3C enumeration step where the controller assigns runtime addresses to targets.
+   * - DMC
+     - Device Management Controller
+     - Target/responder firmware side in this project.
+   * - DMTF
+     - Distributed Management Task Force
+     - Standards body that publishes PLDM and MCTP specifications used by this service.
+   * - DT
+     - Device Tree
+     - Hardware description data used to configure Zephyr, including PLDM sensor bindings and endpoint wiring.
+   * - I3C
+     - Improved Inter-Integrated Circuit
+     - Physical bus used for target connectivity and message transport in this integration.
+   * - MCTP
+     - Management Component Transport Protocol
+     - Transport layer that carries PLDM messages between requester and responder.
+   * - PDR
+     - Platform Descriptor Record
+     - Structured records describing platform entities and sensors used by PLDM discovery.
+   * - PLDM
+     - Platform Level Data Model
+     - DMTF protocol used here for platform discovery and sensor telemetry commands.
+   * - TID
+     - Terminus ID
+     - PLDM identifier returned by ``GetTID`` for a responder terminus.
+
+Architecture
+------------
+
+Component view
+~~~~~~~~~~~~~~
+
+.. mermaid::
+
+   block
+     columns 6
+     block:BMC_CONTAINER
+       columns 1
+       space
+       BMC["BMC"]
+       space
+     end
+     block:DMC:4
+       columns 5
+       space
+       space
+       LABEL["DMC"]
+       space
+       space
+
+       DMC_TGT_IF["I3C Target Ctrl"]
+       space
+       space
+       space
+       DMC_SENSORS["DMC Sensors"]
+
+       MCTP_BIND["MCTP Target"]
+       space
+       block:PLDM
+         columns 1
+         PLDM_LABEL["PLDM"]
+         PLDM_SENS_BIND["Sensor Bindings"]
+         PDR_REPO["PDR"]
+         PLDM_RESP["Responder"]
+       end
+       space
+       SMC_SENSORS["SMC Sensors"]
+     end
+
+      block:SMC_CONTAINER
+       columns 1
+       space
+       space
+       SMC["SMC"]
+     end
+
+     style BMC_CONTAINER fill:none,stroke:none,stroke-width:0px
+     style LABEL fill:none,stroke:none,stroke-width:0px,font-weight:bold
+     style PLDM_LABEL fill:none,stroke:none,stroke-width:0px,font-weight:bold
+     style SMC_CONTAINER fill:none,stroke:none,stroke-width:0px,font-weight:bold
+
+     BMC --- DMC_TGT_IF
+     DMC_TGT_IF --- MCTP_BIND
+     MCTP_BIND --- PLDM_RESP
+     SMC_SENSORS --- PLDM_SENS_BIND
+     SMC_SENSORS --- SMC
+     DMC_SENSORS --- PLDM_SENS_BIND
+     PLDM_SENS_BIND --- PDR_REPO
+     PDR_REPO --- PLDM_RESP
+
+Request/Response Ladder Example
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. mermaid::
+
+   sequenceDiagram
+     participant BMC as BMC
+     participant I3C as I3C Target Interface
+     participant MCTP as MCTP I3C Target Binding
+     participant RESP as PLDM Responder
+     participant PDR as PDR Repository Builder
+     participant MAP as PLDM Sensor Bindings
+     participant SNS as Sensor Nodes
+
+     BMC->>I3C: PLDM over MCTP frame
+     I3C->>MCTP: Deliver frame payload
+     I3C-->>BMC: I3C Write Acknowledged
+     MCTP->>RESP: Dispatch PLDM request
+
+     alt PDR discovery path
+       RESP->>PDR: GetPDRRepositoryInfo / GetPDR
+       PDR-->>RESP: Repository metadata / record data
+     else Sensor reading path
+       RESP->>MAP: Resolve sensor-id to descriptor
+       MAP->>SNS: sensor_sample_fetch + sensor_channel_get
+       SNS-->>MAP: sensor_value
+       MAP-->>RESP: Scaled PLDM reading
+     end
+
+     RESP-->>MCTP: PLDM response
+     MCTP-->>I3C: MCTP frame
+
+     alt IBI
+       I3C-->>BMC: IBI
+       BMC->>I3C: I3C read
+     else Polling
+       BMC->>I3C: I3C read
+     end
+
+     I3C-->>BMC: MCTP response on bus
+
+I3C Target Interface
+~~~~~~~~~~~~~~~~~~~~
+
+This service uses the STM32 I3C peripheral in target mode. It acts as the physical endpoint for
+incoming management traffic. In the DT, ``zephyr,mctp-i3c-target`` binds to that target-mode
+controller and exposes an MCTP transport endpoint to the PLDM responder.
+
+This interface is responsible for:
+
+- participating in I3C bus enumeration/addressing driven by the BMC
+- receiving/sending MCTP frames over I3C once the endpoint is reachable
+- passing payloads up to the MCTP binding used by the PLDM responder
+
+MCTP I3C Target Binding
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The service uses Zephyr's upstream ``zephyr,mctp-i3c-target`` binding/API as
+the transport shim between the STM32 I3C target peripheral and ``libmctp``.
+At init time, the binding instance is created from DT and then registered as a bus with ``libmctp``.
+
+PLDM Responder
+~~~~~~~~~~~~~~
+
+The PLDM responder is the protocol-facing driver component that sits above the
+MCTP I3C target binding and serves PLDM requests from the BMC.
+
+Responsibilities:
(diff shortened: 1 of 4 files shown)


doc/services/pldm/index.rst
```diff
@@ -0,0 +1,230 @@
...
+       compatible = "vendor,zephyr-voltage-sensor";
+       status = "okay";
+     };
+
+     pldm_pdr0: pldm_pdr0 {
+       compatible = "tenstorrent,pldm-pdr";
+
+       pldm_voltage0: pldm_voltage0 {
+         compatible = "tenstorrent,pldm-numeric-sensor";
+         sensor = <&zephyr_voltage_sensor0>;
```
jgrowdenTT, line 482: Multi-chiplet approach isn't specifically outlined here and I'm hoping it can wait.

For multi-chiplet, I believe the options available are:

A) Have different SENSOR_IDs for each sensor per chiplet. 
B) For each chiplet we want info from, we make another MCTP EID/ PLDM TID. 

Option A is just DT and SENSOR_ID additions to what I have here; it's unclear to me how invasive option B is.

Either way, I didn't set anything in stone, yet.

doc/services/pldm/index.rst
```diff
@@ -0,0 +1,228 @@
...
+         channel-type = <SENSOR_CHANNEL>;                /* See Zephyr: enum sensor_channel */
+         channel-index = <0>;
+         sensor-id = <SENSOR_ID>;                        /* Managed by Tenstorrent and stable */
+         entity-type = <ENT_TYPE>;                       /* See libpldm: enum pldm_entity_id_codes */
+         base-unit = <UNIT>;                             /* See libpldm: enum pldm_sensor_units */
+         entity-instance = <1>;
+       };
```
alexapostolu, line 486: This may be something we can think about later, but should we specify what floating point representation to use? I think PLDM's docs suggest IEEE Standard 754-1985 but Zephyr Sensor API uses [`q31` format](https://docs.zephyrproject.org/latest/doxygen/html/group__math__dsp.html#gadc89a3547f5324b7b3b95adec3806bc0)

alexapostolu (commented):
1. If a sensor read fails, does PLDM have a special case for that? I see the `GetSensorReading` in the docs but I'm not sure how errors would be reported

jgrowdenTT:
Outstanding Qs:

> Can we use MCTP forwarding to address different grendel devices?

Not currently. Yes, lib mctp supports bridging. So we could theoretically bind our i3c target to a backend that bridges to the I3C controller for the SMC via lib mctp. But then we lose the MCTP endpoint on the DMC itself since _everything_ is bridged... Bridging some and Handling other based on EID isn't supported, it looks like. 

With custom routing SW in libmctp and zephyr this is probably possible though. But not something I've tested at all.

So, Endpoint ID where "Endpoint is the DMC + all SMCs" is easiest, but yeah, then our SENSOR_IDs need some thinking; Probably not an issue for BH, very likely an issue for grendel.

> Can we parse sensor commands in application software?

Yes, though I'm not seeing much gain for one approach vs the other. instead of a large amount of device tree goo, you would have one small DTS change to pipe the PLDM sensor into a single handler, and that handler would have a lot of C source goo. 

Linking the sensor and PDR would no longer be automatic. In this design it is. For either design we can autogetnerate boilerplate that's annoying to write.

> Add a ladder diagram showcasing initialization

Done

jgrowdenTT:
> If a sensor read fails, does PLDM have a special case for that? I see the GetSensorReading in the docs but I'm not sure how errors would be reported

We would return the PLDM response with an error completion code instead of a success one, and I believe no additional data. Completion codes are defined in the PLDM packet in [PLDM Base Spec](https://www.dmtf.org/sites/default/files/standards/documents/DSP0240_1.2.0.pdf) section 8.1 on page 25.

The CC enumerations are split up between the various PLDM specs. below 0x80 is BASE and everythign above 0x80 is "PLDM Type specific"; i.e. "platform" in this case.

 I think a few of the CC codes relevant to your question are:

Base Spec:
`PLDM_ERROR`

Platform Spec:
`PLDM_PLATFORM_INVALID_SENSOR_ID` (if you pass a sensor not in the PDR)
`PLDM_SENSOR_UNAVAILABLE`

I think the action item on this PR is to provide a ladder diagram for these cases. Note these ladders are implementation agnostic; If we're using PLDM, no matter how we wire the DTS/SW, we need this behaviour.

Ladder diagram for this is added.

Let me know if this addresses your question @alexapostolu. Thanks!

doc/services/pldm/index.rst
jgrowdenTT: I'll probably need clarification on what you're asking. I _think_ the answer is "we have `units` and `unit modifier`", and I'll add an example of adding `unit-modifier` to the DTS.

But to cover all bases along the way:

A) Telemetry in SMC isn't changing, so if we're getting a float, it's 16.16 fixed point format there

B) The zephyr sensor value I believe is fixed with `int32_t integer` part and `int32_t fractional` part. Not changing that.

C) PLDM has a [unit along with unit_modifier](https://www.dmtf.org/sites/default/files/standards/documents/DSP0248_1.2.0.pdf) (Section 27.4) specifying how the variable should be interpreted.

D) the sensor reading is always an integer

E.g lets assume PLDM has a sensor with units = celcius and `unit_modifier` = -1 for some SMC temperature zephyr sensor.

1) SMC sensor fetch gets a value of `integer = 25`, `fractional = 500000` representing temperature.
2) That is converted to `25.5f` and stored in 16.16 format in `telemetry`: `0x00198000` I believe.
3) PLDM asks for this sensor reading=> pipe through the SMC sensor => Smbus fetches `0x00198000`
4) DMC's "smc-sensor" compatible must then convert that back to a zephyr sensore reading;`integer = 25`, `fractional=500000`
5) PLDM must convert this `25`/`500000` return into 255 (where `255 * 10 ^-1 = 25.5`, where `-1` is our rate_unit)

I can probably ladder diagram this, if you think it's valuable.

alexapostolu (dismissed):
Looks good

As discussed offline, Zephyr deprecated the Sensor fetch+get in favour of read+decode. But that's an implementation detail that doesn't affect any external contracts.

doc/services/pldm/index.rst
```diff
@@ -0,0 +1,379 @@
...
+   sequenceDiagram
+     participant BMC as BMC
+     participant I3C as I3C Target Interface
+     participant MCTP as MCTP I3C Target Binding
+     participant RESP as PLDM Responder
+     participant PDR as PDR Repository Builder
+     participant MAP as PLDM Sensor Bindings
+     participant SNS as Sensor Nodes
+
+     BMC->>I3C: PLDM over MCTP frame
```
afongTT, line 129: Any idea why this is happening when rendering the diagram?

Image

jgrowdenTT: Works on my machine both with the downloaded artifacts and my local build. Not sure. Safari, Firefox, and Chrome.

alewycky-tenstorrent:
Please document where each field in the compact* numeric sensor PDR will come from - DTS, zephyr sensor, computed by the sensor/PLDM translation layer? In particular I wonder about baseUnit and unitModifier. There used to be a base-unit property in your example DTS but now it's gone.

\* I assume we're using compact numeric.

doc/services/pldm/index.rst
afongTT: Hmmm ok weird

jgrowdenTT:
@alewycky-tenstorrent added. 

Current approach I have on deck is just using numeric sensor, though compact-numeric would also work and could be supported with a different compatible. 

General idea is that, most stuff we want outside of a few fields are DTS driven. Some stuff can be inferred from DTS links or perhaps KCONFIG options. Anything we don't currently care about can be statically set.

On the current design I have posted, only feedback I might have is opening up DTS options for accuracy and tolerance, though with defaults available to omit if it's unknown.

doc/services/pldm/index.rst
```diff
@@ -0,0 +1,589 @@
...
+     - Computed from the encoded struct size.
+   * - ``terminus_handle``
+     - Static code
+     - Fixed at ``1``.
+   * - ``validity``
+     - Static code
+     - Set to ``PLDM_TL_PDR_VALID``.
+   * - ``tid``
+     - DTS
+     - Comes from ``pldm-tid`` on ``tenstorrent,pldm-mctp-responder``.
```
afongTT: Something I am confused about is whether TID needs to be unique across the entire system (i.e. from BMC perspective). I think there's also a dynamic TID assignment mechanism?

Also would that need to be unique per tray or per Galaxy? (i.e. does tBMC forward PLDM requests from sBMC?)

doc/services/pldm/index.rst
```diff
@@ -0,0 +1,589 @@
...
+
+- discover IDs from PDRs (``GetPDRRepositoryInfo`` + ``GetPDR``), then use
+  those IDs for ``GetSensorReading``
+- do not infer identity from record order or display name alone
+
+Notes and limits
+----------------
+
+- ``GetPDR`` supports multipart transfer with CRC in final chunk.
+- Current auxiliary name encoding is UTF-16BE with ``en`` language tag.
```
afongTT, line 632: UTF-16BE? Seems weird, but I don't really know?

doc/services/pldm/index.rst
```diff
@@ -0,0 +1,589 @@
...
+
+Compatibility
+-------------
+
+To keep requester integrations stable across firmware revisions, this service
+uses a fixed-ID compatibility model.
+
+Compatibility policy:
+
+- treat ``sensor-id`` as the stable external contract; IDs are defined and documented in firmware source
```
afongTT, line 613: Should these be documented as Doxygen in a C header or something else? (.dtsi?)

doc/services/pldm/index.rst
```diff
@@ -0,0 +1,589 @@
...
+     - Encoded from ``sensor-name`` as UTF-16BE.
+
+Sensor Nodes
+~~~~~~~~~~~~
+
+Any Zephyr sensor device API can be used to source PLDM sensor values. The
+binding from a sensor-compatible device to the PDR must be defined in DT in
+order to be visible via PLDM.
+
+``tenstorrent,bh-arc-telemetry`` acts as a DMC sensor driver to map SMC
```
afongTT, line 423: Wondering if it makes sense (or is even possible) for bh-arc-telemetry to have its own TID?

doc/services/pldm/index.rst
jgrowdenTT: Plan is a header with defines only which can be included in a dts file or C source. and yeah, doxygen.

doc/services/pldm/index.rst
jgrowdenTT: In claude we trust?

doc/services/pldm/index.rst
jgrowdenTT: I don't think that is possible. Is possible (and necessary!) if it's under a different MCTP endpoint.

doc/services/pldm/index.rst
jgrowdenTT: PLDM TID (terminus ID) has a default here. It is plausibly has to be set by BMC for uniqueness. In which case, We must support `SetTID` for this.

Like, It really comes down to "what is the MCTP network", which is a BMC concern; If they're designing for a bunch of point to point MCTP netowrks with 1 DMC, then they don't need to send us SetTID. We'll be on EID=TID=11.

If it's a single uber network with  SBMC(?) -> TBMC-> DMCx8 then yeah, we need SetTID support. IMO it's no skin off our back to support SetTID and make it up to them to call it.

What's the impact on this document? I think:

- clarify this with a * that indicates we can support dynamix TID from BMC?
- A section that outlines every command we plan on adding support for.

(MCTP EID has a similar answer; DTS for default and ctrl msg for dynamic update.)

doc/services/pldm/index.rst
```diff
@@ -0,0 +1,589 @@
...
+     - Physical bus used for target connectivity and message transport in this integration.
+   * - MCTP
+     - Management Component Transport Protocol
+     - Transport layer that carries PLDM messages between requester and responder.
+   * - PDR
+     - Platform Descriptor Record
+     - Structured records describing platform entities and sensors used by PLDM discovery.
+   * - PLDM
+     - Platform Level Data Model
+     - DMTF protocol used here for platform discovery and sensor telemetry commands.
```
njelicTT, line 46: DMTF acronym should also be explained here

njelicTT (approved):

doc/services/pldm/index.rst
afongTT: Looks like UTF-16BE is part of the spec. The `en` language tag doesn't seem like an actual thing.

afongTT (approved):

doc/services/pldm/index.rst
jgrowdenTT: https://www.dmtf.org/sites/default/files/standards/documents/DSP0248_1.2.0.pdf

28.8:

> 
> The nameLanguageTag field can be used to identify the language (such as French, Italian, or English)
> that is associated with the particular sensorName. Table 83 describes the format of this PDR.

> nameLanguageTag [1]
> A null-terminated ISO646 ASCII string that holds a language tag, per RFC4646, that identifies
> the primary language in which the sensorName was defined (for example, "en" for English, "zh-
> cmn-Hans" for simplified Mandarin Chinese, and so on). This field may be used to help select
> which string to use when multiple character encodings for the sensorName are provided.
> special value: null string = 0x0000 = unspecified

For this document. I can link to 28.8 and use nameLanguageTag instead of just "language tag" to make ctrl+f easier. Thoughts?

afongTT: Sounds good. Clearly I didn't read carefully enough.

alexapostolu (approved):

alewycky-tenstorrent (approved):

alayander (approved):

PR #1314, opened 2026-06-28
ccfgovr: add more native sim and e2e pytests

alexapostolu:
ccfgovr: add more native sim and e2e pytests


projkovTT:
@alexapostolu Is this still happening? Pls update and tag reviewers if so or close if not

projkovTT:
I think that `test_ccfgovr_eth_speed_override` needs to be updated to reflect changes made in https://github.com/tenstorrent/tt-system-firmware/pull/1411. Currently failing on all platforms.

.github/workflows/hardware-long.yml
```diff
@@ -79,6 +79,22 @@ jobs:
           app-path: tt-system-firmware
           kmd-version: ${{ matrix.config.kmd_build }}
 
+      - name: Install bh-mod
```
projkovTT, line 82: Why do we want to install bh-mod instead of build?

app/smc/pytest/e2e_smoke.py
```diff
@@ -2159,3 +2057,89 @@ def test_heartbeat_telemetry(arc_chip_dut, asic_id):
...
+    # firmware default, not a leftover override from an earlier run.
+    _reset_ccfgovr(bh_mod, asic_id)
+    default_tdp = _read_tdp_limit(asic_id)
+    target = default_tdp - 5
+    logger.info(f"ccfgovr smoke: default tdp_limit_max={default_tdp}, target={target}")
+
+    # Persist an override and confirm the running firmware applies it.
+    result = _bh_mod_set(bh_mod, f"chip_limits.tdp_limit={target}")
+    assert result.returncode == 0, f"bh-mod set failed, rc={result.returncode}"
+    measured = wait_arc_boot(asic_id).get_telemetry().tdp_limit_max
```
projkovTT, line 2134: We're missing reflash with `unlaunched_dut.launc()` in this updated test. Right now it just lets it reset.

app/smc/pytest/e2e_stress.py
```diff
@@ -597,3 +610,294 @@ def test_tensix_reset_then_burnin(arc_chip_dut, asic_id):
...
+    assert measured_tdp_after_flash == tdp_target, (
+        f"ccfgovr did not survive tt-flash: "
+        f"expected tdp_limit_max={tdp_target}, got {measured_tdp_after_flash}"
+    )
+    measured_kt_after_flash = read_telem(chip, TAG_KERNEL_THROTTLER)
+    logger.info(f"kernel_throttler after re-flash: 0x{measured_kt_after_flash:08x}")
+    assert measured_kt_after_flash == kt_expected, (
+        f"ccfgovr did not survive tt-flash: expected 0x{kt_expected:08x}, "
+        f"got 0x{measured_kt_after_flash:08x}"
+    )
```
projkovTT, line 699: Since we're setting a few configs I think it would be prudent to make sure that we reset to default values after the test.
Mostly I'm thinking of the failures that we sometimes see with the TDP high/low power delta test - persistent ccfgovr would definitely affect that and making sure that we reset just eliminates that possibility.

projkovTT: Oh I see we have `test_ccfgovr_res_all` which does that. If we're sure that we only run bh_mod tests between affecting the fwtable and `test_ccfgovr_res_all` then its fine

.github/workflows/hardware-long.yml
alexapostolu: bh-mod exists prebuilt on another repo github.com/tenstorrent/luwen, so we have to install it from there

projkovTT: Any concerns with versioning? I think that when we were building it was always off latest main. This has the potential to get outdated.

PR #1315, opened 2026-06-30
CI smoke & long RTT logging: skip galaxy, loud box and qb RTT logs 

kaiqihuTT:
This PR updates the hardware CI workflows to make post-failure diagnostics probe-aware by gating RTT collection based on board type, collecting DMC RTT only on boards with ST-Link/SWD access, collecting SMC RTT only on boards with J-Link access, and keeping SMC state dumping as a separate failure-only diagnostic because it does not require a debug probe. The goal is to avoid misleading CI failures where the Print RTT logs step errors only because the required debug hardware is not present, making the workflow better match the actual capabilities of each board configuration while preserving useful post-failure debug behavior.

### DIFF
--- a/.github/workflows/hardware-long.yml
+++ b/.github/workflows/hardware-long.yml
@@ -99,14 +99,22 @@ jobs:
             zephyr/twister-*e2e*/**/update.fwbundle
             zephyr/twister-*e2e*/**/recording.csv
 
-      - name: Print RTT logs
-        if: ${{ failure() }}
+      - name: Print DMC RTT logs
+        # DMC RTT goes over the STLink/SWD adapter; galaxy/loudbox/quietbox2 have none.
+        if: >-
+          ${{ failure()
+          && matrix.config.board != 'bh-galaxy'
+          && matrix.config.board != 'loudbox'
+          && matrix.config.board != 'quietbox2' }}
         working-directory: tt-system-firmware
         run: |
           echo "DMC RTT logs:"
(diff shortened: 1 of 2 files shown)


afongTT:
> This is NOT the finalized change that is meant to be merged to origin/main. It is a test meant to verify RTT logging behaviour with the p150b on hand.

Could you just update the PR description?

afongTT (approved):

.github/workflows/hardware-long.yml
```diff
@@ -99,14 +99,22 @@ jobs:
...
+          && matrix.config.board != 'loudbox'
+          && matrix.config.board != 'quietbox2' }}
         working-directory: tt-system-firmware
         run: |
           echo "DMC RTT logs:"
           python3 ./scripts/dmc_rtt.py -n
-          echo "SMC RTT logs:"
-          python3 ./scripts/smc_console.py --rtt -n
+
+      - name: Dump SMC state
```
psethiTT, line 114: @kaiqihuTT question out of curiosity - why were smc RTT logs removed? Do we not have a way to read them if JTAG is connected?

psethiTT (approved):

.github/workflows/hardware-long.yml
afongTT: We removed all the JLinks (on ARC JTAG) from the CI systems, so we can't get SMC RTT logs anymore.

PR #1316, opened 2026-06-30
drivers/reset: remove unused reset driver

alexapostolu:
Remove unused reset driver.

Even if we did use it, it would be more work to do so then to just write to the reset registers directly.


jgrowdenTT:
I fear this will cause native sim and nucleo DMC builds to break if anything ends up hitting those reset registers. Did you try?

We could continue using `WriteReg`/`ReadReg` in `lib_bharc`. We intercept these calls to the stub out.

Alternativelly, we move to using the driver, which is probably the better SW design anyways. I don't think we need N of them in the device tree, a singular "BH reset" driver would have been enough since the mem range is all back to back and it looks like reset API uses "id" anyways, if I'm reading the docs right.

alexapostolu:
@jgrowdenTT Good points, I'll play around with the driver to try to make it work

PR #1322, opened 2026-07-03
tensix_reset: use noc coord instead of phys coord for clock gate

alexapostolu:
Tensix reset has been failing intermittently in CI. This is due to SetSingleTileClockGate using physical coordinates instead of NOC coordinates, causing improper translation and a random tile being clock gated. This will hang if the random tile is harvested, causing the intermittent CI failure.

Use NOC coords instead of physical coords for SetSingleTileClockGate.

### DIFF
--- a/app/smc/pytest/e2e_stress.py
+++ b/app/smc/pytest/e2e_stress.py
@@ -505,8 +505,8 @@ def test_power_virus(arc_chip_dut, asic_id):
 )
 def test_tensix_reset_then_burnin(arc_chip_dut, asic_id):
     """
-    Reset every non-harvested tensix tile, then run tt-burnin and check
-    that it exits successfully.
+    Reset every non-harvested Tensix tile, then run tt-burnin and check
+    it exits successfully.
     """
     arc_chip = pyluwen.detect_chips()[asic_id]
 
@@ -517,9 +517,9 @@ def test_tensix_reset_then_burnin(arc_chip_dut, asic_id):
 
     # We want to test the tensix reset message on low power.
     # Set high power after the message to allow NOC read/write to function properly.
+    logger.info(f"Resetting {len(all_tiles)} Tensix tiles")
     arc_chip.set_power_state("low")
     for noc_x, noc_y in all_tiles:
-        logger.info(f"Resetting ({noc_x},{noc_y})")
         response = arc_chip.arc_msg(
             TT_SMC_MSG_TOGGLE_SINGLE_TENSIX_RESET,
             arg0=noc_x | (noc_y << 8),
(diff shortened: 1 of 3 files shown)


alexapostolu:
I will fix the compliance errors after Stress runs

lib/tenstorrent/bh_arc/noc_init.c
```diff
@@ -221,9 +221,9 @@ int32_t set_tensix_enable(bool enable)
 	return 0;
 }
 
-void SetSingleTileClockGate(uint8_t phys_x, uint8_t phys_y, bool gate)
+void SetSingleTileClockGate(uint8_t noc0_x, uint8_t noc0_y, bool gate)
```
jgrowdenTT, line 224: The caller in `reset.c` wasn't adjusted (right?) and is still sending what they call `phys_x`/`phys_y` into here.

I.e. the coordinates we're passing in to this func are from:

	`NocLogicalToPhysical(noc_x, noc_y, &phys_x, &phys_y);`

Where the arc msg is passing `noc_x`/`noc_y`, but it is `phys_x` and `phys_y` which are being passed into `noc0_x` and `noc0_y` above.

This doesn't look right; I think either keeping in `phys`, or using this new impl but with with the `noc` vars, aligns with what I expect. But then that's just the same as we had before...

I could be missing something obvious. Would you mind extrapolating?

alexapostolu: Ah yes you're right, I also need to adjust the caller to pass in NOC coords

The reason I think this works is because since hardware translation is on, it requires logical NOC coordinates so it can map those to the physical tile. While the PHYS coordinates are the output of that mapping, which is incorrect since the hardware will try to translate those again.

To expand on this, I think a functionally equivalent solution would be to disable translation and to use the PHYS coordinates.

app/smc/pytest/e2e_stress.py
```diff
@@ -528,15 +528,46 @@ def test_tensix_reset_then_burnin(arc_chip_dut, asic_id):
...
-    result = subprocess.run(
+    burnin_process = subprocess.Popen(
         ["tt-burnin", "--no-reset"],
-        stdin=subprocess.DEVNULL,
-        capture_output=True,
-        timeout=300,
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
     )
```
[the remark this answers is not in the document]
alexapostolu: Will fix after stress runs

lib/tenstorrent/bh_arc/noc_init.c
jgrowdenTT: Thanks Alex!

jgrowdenTT (dismissed):
The red light is just galaxy being a jerk. unrelated to this I imagine since we got nowhere near testing this.

jgrowdenTT (approved):

alexapostolu:
@psethiTT Pinging again for review. Do you mind if I merge this today? That way we can see two nightly Stress runs with this change before Monday

alayander (approved):

psethiTT:
> @psethiTT Pinging again for review. Do you mind if I merge this today? That way we can see two nightly Stress runs with this change before Monday

merged

PR #1323, opened 2026-07-03
ci: bh_glx: use tt-smi -glx_reset

alexapostolu:
Use tt-smi -glx_reset instead of tt-smi -r for resetting BH GLX in CI.


afongTT (dismissed):

jgrowdenTT:
It doesn't look like this survives reset.

PR #1325, opened 2026-07-04
drivers: efuse: use register yaml bindings for reg base

jgrowdenTT:
Use the standard zephyr pattern of using reg bindings to set the base address of each efuse instance.

Tests: Smoke (which has an efuse test, which is hopefully okay enough to catch regressions.)


alexapostolu (dismissed):

alayander (dismissed):

drivers/misc/bh_efuse/bh_efuse.c
alayander, line 1: Could run clang-format on this to resolve the compliance errors

alexapostolu (approved):

alayander (approved):

PR #1327, opened 2026-07-04
soc: tenstorrent: pinctrl

jgrowdenTT:
Add pinctrl driver, and remove some superfluous lib_bharc hooks in our i2c driver.

This is a setup step to using i2c1 driver as intended and removing the use of the dw_apb_i2c lib_bharc code, allowing for easier emulation and potentially better app-side code re-use.

tests:

Pad cfg regs for i2c0 are identical before/after:
```
jg-dev:~$ devmem 0x800301c0
Using data width 32
Read value 0x1fcc3
jg-dev:~$ devmem 0x800300f0
Using data width 32
Read value 0x3
```

Smoke.


alexapostolu (approved):

soc/tenstorrent/tt_blackhole/pinctrl_bh.c
```diff
@@ -0,0 +1,128 @@
...
+#include <errno.h>
+
+#include <zephyr/arch/cpu.h>
+#include <zephyr/device.h>
+#include <zephyr/drivers/pinctrl.h>
+#include <zephyr/arch/common/sys_bitops.h>
+#include <zephyr/sys/sys_io.h>
+#include <zephyr/dt-bindings/pinctrl/tt_blackhole_smc-pinctrl.h>
+
+#define DW_APB_I2C0_REG_MAP_BASE_ADDR     0x80060000u
```
psethiTT, line 16: Since these registers are also defined in dw_apb_i2c.c, does it make sense to define a shared header file? It seems that the dw_apb_i2c.c is defined at bh_arc level  where as pin_ctrl_bh.c is defined at soc level. BH ARC could include from SOC level.

Also just like we removed the use of function I2C init fucntion from SMBUS target, I believe the goal is to phase out dw_apb_i2c.c eventually. In that case then this exercise might be redundant in long term?

jgrowdenTT: Yeah; the hope is that we can remove `dw_apb_i2c.c`. So sharing code between this and something on the culling block isn't too valuable.

PR #1328, opened 2026-07-05
drivers: smbus_target: refactor for easier command registration

jgrowdenTT:
### Overview
Refactor the SMBUS driver so command registration is easier. Provide helper macros for building an item, and provide a bulk register command. Follow zephyr style

This is just starting cleanup to hopefully make the usage of the driver easier for users, and thus more desirable for eventual upstream to zephyr.

### Tests:
smoke shoud run with our SMBUS test on.
Existing native sim lib bh_arc has tests verifying SMBUS behaviour

### DIFF
--- a/lib/tenstorrent/bh_arc/smbus_target.c
+++ b/lib/tenstorrent/bh_arc/smbus_target.c
@@ -24,6 +24,7 @@
 #include <zephyr/drivers/i2c.h>
 #include <zephyr/sys/byteorder.h>
 #include <zephyr/sys/crc.h>
+#include <zephyr/sys/util.h>
 
 /* DMFW to CMFW i2c interface is on I2C0 of tensix_sm */
 #define CM_I2C_DM_TARGET_INST 0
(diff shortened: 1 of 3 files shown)


alexapostolu (approved):

lib/tenstorrent/bh_arc/smbus_target.c
```diff
@@ -222,45 +172,11 @@ static int InitSmbusTarget(void)
...
-				  &smbus_test_read_block_cmd_def);
-	smbus_target_register_cmd(smbus_target, CMFW_SMBUS_TEST_WRITE_BLOCK,
-				  &smbus_test_write_block_cmd_def);
-	smbus_target_register_cmd(smbus_target, CMFW_SMBUS_TEST_WRITE_BLOCK_READ_BLOCK,
-				  &smbus_block_write_block_read_test);
-
-	smbus_target_register_cmd(smbus_target, CMFW_SMBUS_PING_V2, &smbus_ping_v2_cmd_def);
+	if (smbus_target_register_cmds(smbus_target, smbus_cmds, ARRAY_SIZE(smbus_cmds)) < 0) {
+		printk("Failed to register SMBUS target commands\n");
+		return 0;
```
alayander, line 177: I noticed we return 0 on all "failure paths" in this function from before as well, is there a reason we do that instead of returning an error code/return code of the failing call?

alayander (approved):
Just a non-blocking question

lib/tenstorrent/bh_arc/smbus_target.c
jgrowdenTT: I'm not sure. If we return EC here It's plausible we don't survive system initialization. That might be worse than the surviving. I don't mind either way so long as we can save the device. I'd much rather limp to a recoverable-but-busted state than crash to a non-recoverable one. 

Just my 2c. We can look into it later, if we're interested.

PR #1330, opened 2026-07-07
lib: bh_arc: engage kernel NOPs on high GDDR temperature

alayander:
Once AICLK has already been throttled to its floor and the max GDDR temp reaches GDDR_THERM_TRIP_TEMP, stall Tensix kernels via NOPs for thermal trip mitigation. NOPs are released once GDDR cools past a hysteresis band.

image

The GDDR_NOP_TEMP (nop request activation temp) was lowered to 65 for this trace.

Resolves: SYS-4723, SYS-4486

### DIFF
--- a/lib/tenstorrent/bh_arc/throttler.c
+++ b/lib/tenstorrent/bh_arc/throttler.c
@@ -19,6 +19,7 @@
 #include "telemetry.h"
 #include "noc2axi.h"
 #include "tensix_state_msg.h"
+#include <tenstorrent/bh_arc.h>
 
 static uint32_t power_limit;
 
@@ -40,6 +41,9 @@ static uint32_t kernel_throttler_stop_nops_freq_default;
 
 #define kThrottlerAiclkScaleFactor 500.0F
 #define DEFAULT_BOARD_POWER_LIMIT  150
+
+#define GDDR_NOP_TEMP       GDDR_THERM_TRIP_TEMP /* engage NOPs at/above this GDDR temp */
+#define GDDR_NOP_HYSTERESIS 2 /* release NOPs once temp drops this far below engage */
 
 LOG_MODULE_REGISTER(throttler);
 
@@ -319,6 +323,28 @@ static uint16_t *board_power_history_cursor = board_power_history;
 static uint32_t board_power_sum;
 static bool kernel_nops_enabled;
 
+/* Kernel NOPs are a single shared lever driven by independent requesters.
+ * Each source owns exactly one bit with NOP engagement based on the unionized value.
+ */
+static union {
+	uint32_t u32_all;
+	struct {
+		uint32_t doppler: 1;
+		uint32_t kernel_throttler: 1;
+		uint32_t gddr_thermal: 1;
+	};
+} nop_requests;
+
+static void apply_nop_requests(void)
+{
+	bool new_kernel_nops_enabled = (nop_requests.u32_all != 0U);
+
+	if (new_kernel_nops_enabled != kernel_nops_enabled) {
+		kernel_nops_enabled = new_kernel_nops_enabled;
+		SendKernelThrottlingMessage(kernel_nops_enabled);
+	}
+}
+
 static uint8_t t2_count;
 static uint8_t t3_count;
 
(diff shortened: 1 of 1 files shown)


lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -41,6 +42,9 @@ static uint32_t kernel_throttler_stop_nops_freq_default;
 #define kThrottlerAiclkScaleFactor 500.0F
 #define DEFAULT_BOARD_POWER_LIMIT  150
 
+#define GDDR_NOP_TEMP       GDDR_THERM_TRIP_TEMP /* engage NOPs at/above this GDDR temp */
```
alayander, line 45: Currently I've set the threshold to be the sustained therm trip temp (95C), such that the mitigation only occurs when in this critical temperature range.

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -319,6 +323,37 @@ static uint16_t *board_power_history_cursor = board_power_history;
...
+ * still needs them.
+ */
+enum {
+	kNopSrcPower = BIT(0),
+	kNopSrcGddrThermal = BIT(1),
+};
+
+static uint32_t nop_requests;
+
+static void SetNopRequest(uint32_t src, bool active)
```
jgrowdenTT: nit, arguably should follow zephyr style for new funcs.

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -319,6 +323,37 @@ static uint16_t *board_power_history_cursor = board_power_history;
 static uint32_t board_power_sum;
 static bool kernel_nops_enabled;
 
+/* Kernel NOPs are a single shared lever between independent requesters.
+ * Each source latches its own request bit and the shared state is the OR of
+ * all of them, so one source releasing NOPs cannot stomp another source that
+ * still needs them.
+ */
+enum {
```
jgrowdenTT: Use a bitfield struct. unionize with a u32all. scrap enum. SetNopRequests becomes more or less useless because callers can directly do:
```
nop_requests.power = 1; // or 0
...

nop_requests.thermal = 1; // or 0

...
bool kernel_nops_enabled = (nop_requests.u32_all != 0U)
```
etc.

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -319,6 +323,28 @@ static uint16_t *board_power_history_cursor = board_power_history;
...
+
+static void apply_nop_requests(void)
+{
+	bool new_kernel_nops_enabled = (nop_requests.u32_all != 0U);
+
+	if (new_kernel_nops_enabled != kernel_nops_enabled) {
+		kernel_nops_enabled = new_kernel_nops_enabled;
+		SendKernelThrottlingMessage(kernel_nops_enabled);
+	}
+}
```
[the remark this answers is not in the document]
alayander: I believe they run on the same execution context, so there's no actual concurrency. Both are dispatched to the system workqueue via k_work_submit.

jgrowdenTT:
I _think_ this is okay?

Appreciate if you could double check the NOP-on-AICLK floor feature still works as intended.

jgrowdenTT (approved):

alayander:
> Appreciate if you could double check the NOP-on-AICLK floor feature still works as intended.

Yeah I'll confirm this - please hold off on merging this until after @psethiTT

lib/tenstorrent/bh_arc/throttler.c
```diff
@@ -41,6 +42,9 @@ static uint32_t kernel_throttler_stop_nops_freq_default;
 #define kThrottlerAiclkScaleFactor 500.0F
 #define DEFAULT_BOARD_POWER_LIMIT  150
 
+#define GDDR_NOP_TEMP       GDDR_THERM_TRIP_TEMP /* engage NOPs at/above this GDDR temp */
+#define GDDR_NOP_HYSTERESIS 2 /* release NOPs once temp drops this far below engage */
```
psethiTT, line 46: @alayander (as discussed offline as well), it will be good to get some data on what this GDDR_NOP_HYSTERESIS variable should be set to when temperature is around 95C. In other words, how quickly does the GDDR temperature rise again after NOP insertion is disabled when temperature drops from 95C to <=93C. My concern is that we will constantly be entering and exiting NOP insertion state based on how quickly the temperature rises at that temperature range. 

I think creating a test scenario (around that temperature range) for this will be difficult. But a lower temperature (like you tested at 65C) can give give a good idea of how quick temperature rises back to 65C. I believe the slope of rise will be higher at 95C so we can add some "buffer" (1-2C) on top of the hysteresis number at 65C.

psethiTT (commented):
Added a comment about evaluating the hysteresis number

alayander:
Per SYS-4486, this feature will be deferred - so will close ticket for now and revisit in future if needed

PR #1331, opened 2026-07-08
lib: bh_arc: use i2c_dw for i2c1

jgrowdenTT:
Use i2c_dw for i2c1.


jgrowdenTT:
It works locally except for dirty reset in stress. CLosing this til I can figure it out.

PR #1337, opened 2026-07-13
docs: Compatability and feature capability/enablement advertisement 

jgrowdenTT:
Document our compatability strategy in open source.

Add feature capability bits and build in an example (kernel-nops)

tests:

We have smoke tests for kernel nops feature. Whatever tests we are lacking are out of scope for this document.

I verified the documentation builds. Though some links are broken locally, I'm hopeful that once published everything will be okay.

### DIFF
--- a/doc/develop/guides/compatibility.rst
+++ b/doc/develop/guides/compatibility.rst
@@ -1,0 +1,217 @@
+.. _ttzp_compatibility:
+
+Compatibility Guidelines
+========================
+
+This guide documents compatibility guarantees for externally consumed firmware
+interfaces. The goal is to let newer firmware continue working with older
+controllers, and to let newer controllers detect and adapt to older firmware
+without relying on undefined behavior.
+
+Compatibility Model
+-------------------
+
+Tenstorrent firmware uses a stability-first model for published host-facing
+interfaces:
+
+- Existing telemetry tags keep their meaning.
+- Existing host message IDs and request layouts keep their meaning.
+- Existing SMBus register IDs and payload meanings keep their meaning.
+- New functionality is added by extension, not by repurposing an existing
+  field, message, or register.
+
+Controllers must assume that older firmware may not implement newly added
+features. Firmware must assume that older controllers may continue using older,
+previously documented interfaces indefinitely.
+
+Interface Rules
+---------------
+
+Telemetry Interface
+~~~~~~~~~~~~~~~~~~~
+
+The telemetry interface is defined in ``lib/tenstorrent/bh_arc/telemetry.h``.
+
+Compatibility Discovery
+***********************
+
+Controllers can and should detect newly added telemetry entries by reading the
+telemetry table metadata and tag table, rather than assuming a fixed tag set is
+present.
+
+As documented in
+`Procedure to Read Telemetry </tt-system-firmware/services/telemetry/index.html#procedure-to-read-telemetry>`_:
+
+- Read ``telemetry_table.entry_count``.
+- Walk the telemetry ``tag_table``.
+- Use the discovered tag-to-offset mapping to determine whether a given tag is
+  present in the running firmware.
+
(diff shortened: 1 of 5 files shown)


lib/tenstorrent/bh_arc/telemetry.h
```diff
@@ -25,6 +25,58 @@
...
+ * | @ref TAG_FW_ACTIVE_CONFIG_0 | The feature is currently enabled. |
+ */
+typedef struct {
+	/** @brief Kernel-throttler-at-AICLK-floor feature.
+	 *
+	 * This bit is controlled at runtime by the host with
+	 * @ref TT_SMC_MSG_CHARACTERISATION using
+	 * @ref TT_SUB_MSG_SET_KERNEL_THROTTLER_ENABLED.
+	 */
+	uint32_t kernel_nops_at_aiclk_fmin: 1;
```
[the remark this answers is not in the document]
jgrowdenTT: Yep, good advice, but given we already bought into this schema with msgqueue, I think we're going to assume order on the C bitfield definitition. It makes doxygen better which I (currently) value more than portability. Remind me about this in 2 years lol.

Anyways, If there is eventually a target we compile for where bitfield endianness is hosed, then we macroize these fields in reverse order to converve the ABI.

I do not advise to use this header outside of FW.

doc/develop/guides/compatibility.rst
```diff
@@ -0,0 +1,205 @@
...
+This is the preferred discovery mechanism for new telemetry entries.
+
+Compatibility Rules
+*******************
+
+- Do not rename an existing telemetry tag.
+- Do not repurpose an existing telemetry tag to mean something else.
+- Do not change the encoding or field definition of an existing telemetry tag.
+- Do not change the meaning of an existing bit in a published telemetry bitfield.
+- Best practice is to reserve unused bits explicitly, and keep them reading as 0 until they are
```
joelsmithTT, line 59: If FW enforces that all reserved fields are 0 (and will e.g. fail a message request if the controller passes in stack garbage in a reserved field) then it opens up use of those reserved fields in the future.

But if there is no such enforcement, a bugged/broken controller may fail upon a FW change that repurposes a reserved field in an ABI-compatible way.

Is it this case the rule is intended to cover?

jgrowdenTT: Yes!  Also noting, if it's **not** documented as reserved, it's not reserved... i.e. it's a contract break to use those bits later.

jgrowdenTT: I was thinking of specifically telemetry fields. 

In the case of msgqueue. I don't think FW will check other fields == 0; Future versions of the command might specify that data as valid. We have no general way in-band of knowing "new" vs "old" Thus our policy is to accept "reserved field" data as unchecked.

1. Made up example:  `TT_SMC_MSG_BLINKY` msg. We could extend with another word to specify what LED. If new host specifies that field, old FW will just ignore it and do what it always did on the LED it did before.

2. Real example `TT_SMC_MSG_POWER_SETTING` -> We have to accept power bits we don't know anything about as a valid 1. We just can't do anything with the 1 bits that aren't specified. Therefore rejecting the message on bits= 0 is improper.

3. Real example `TT_SMC_MSG_PING_DM`  -> We added the ability to perform a "legacy ping" vs a "new ping" which "mean" the same thing, though "new ping" is more stable on bus.

I would argue that all of these examples follow the compatability guideance for "new controller/old FW" documented below:

`New controllers must discover capabilities and only use features supported by the older firmware. New optional functionality must degrade cleanly.` i.e. "behaviour should fall back safely instead of breaking" and "behaviour is well defined" in any of these three cases.

Maybe this, and the matrix below, needs better wording?

alexapostolu (approved):

alayander (approved):
lgtm

psethiTT (approved):

PR #1338, opened 2026-07-13
drivers: sensor: tenstorrent: bh_arc_telem

jgrowdenTT:
Add a bh_arc_telem sensor driver.

This driver gets configured via DTS to map zephyr channel/index pairs into an SMC telemetry entry. This compatible can be optionally enabled in DMC side DTS files for DMC boards which want to use the zephry sensor API for SMC; currently none exist, though soon PLDM will. We can add it to to the emulated overlay anyways so that we can easily test with it in later.

tests:

native sim tests for this driver was added against lib_bharc emulation. Currently, the best telemetry to test against is temperature since it is stubbed out in PVT. This test gets run in github actions.


drivers/sensor/tenstorrent/bh_arc_telem/bh_arc_telem.c
```diff
@@ -0,0 +1,322 @@
...
+#include <tenstorrent/tt_smbus_regs.h>
+
+#include <errno.h>
+
+#include <telemetry.h>
+
+LOG_MODULE_REGISTER(bh_arc_telem, CONFIG_SENSOR_LOG_LEVEL);
+
+/* Max bytes returned by a single-tag CMFW_SMBUS_TELEMETRY_READ (56-bit payload) */
+#define BH_ARC_TELEM_RCV_BUF_LEN 7U
```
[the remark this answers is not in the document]
jgrowdenTT: Hm. I'd be willing to accept: 

A) smbus_block_read maanges the 255 for us, and copies into this small buffer (since I know how much I want.)
B) Per compatability constraints, we just say we know it's 7 always. 

Inclined to lean towards B.

tests/drivers/sensors/bh_arc_telem/src/main.c
```diff
@@ -0,0 +1,170 @@
...
+	rc = sensor_read(iodev, &test_rtio, read_buf, base_size);
+	if (rc != 0) {
+		return rc;
+	}
+
+	struct sensor_decode_context decode_ctx =
+		SENSOR_DECODE_CONTEXT_INIT(decoder, read_buf, channel_type, channel_index);
+
+	rc = sensor_decode(&decode_ctx, value, 1U);
+	if (rc <= 0) {
```
alexapostolu, line 75: I think the equality can be dropped because the API for `sensor_decode` says return `-ENODATA` if "requested data type not present in the frame". So I think if there's 0 elements then the function should just return `-ENODATA` (which this driver already does) instead of `0`.

drivers/sensor/tenstorrent/bh_arc_telem/bh_arc_telem.c
```diff
@@ -0,0 +1,322 @@
...
+		if (ret != 0) {
+			rtio_iodev_sqe_err(iodev_sqe, ret);
+			return;
+		}
+
+		ret = bh_arc_telem_raw_to_sensor_value(map, raw, &data[i].value);
+		if (ret != 0) {
+			rtio_iodev_sqe_err(iodev_sqe, ret);
+			return;
+		}
```
alexapostolu: I think the conversion must happen in the decoder. The submit function just submits raw data to the buffer

https://docs.zephyrproject.org/latest/hardware/peripherals/sensor/index.html
> SHOULD implement sensor_submit_t in a way that directly reads into the provided buffer avoiding copying of any kind, with few exceptions.
> MUST implement sensor_decoder_apiwith pure stateless functions. All state needed to convert the raw sensor readings into fixed point SI united values must be in the provided buffer.

tests/drivers/sensors/bh_arc_telem/src/main.c
jgrowdenTT: I think what I have is correct, in the sense that if the driver did _not_ return ENODATA but did return 0:

In the current implementation -> this function would retrun ENODATA to the test
In your suggestion -> this function would return 0 (success)
 
 I think we would _want_ to see ENODATA in the case the driver bugged out somehow. Right?

tests/drivers/sensors/bh_arc_telem/src/main.c
```diff
@@ -0,0 +1,211 @@
...
+	rc = read_channel(&temp_iodev, SENSOR_CHAN_DIE_TEMP, 0U, &value);
+	zassert_ok(rc, "negative temperature read failed: %d", rc);
+	zassert_within(sensor_value_to_micro(&value), -5250000LL, TELEM_TOLERANCE_MICRO);
+}
+
+ZTEST(bh_arc_telem, test_temperature_reflects_latest_update)
+{
+	struct sensor_value value;
+	int rc;
+	const int iterations = 50;
```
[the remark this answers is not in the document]
jgrowdenTT: It's emulated. It's pretty quick.

alexapostolu (approved):

alayander (approved):

alexapostolu (approved):

PR #1339, opened 2026-07-14
drivers: pmci: pldm_pdr

jgrowdenTT:
Add pldm_pdr driver and tests.

The PDR is a repository of sensor information that is querriable via PLDM via the platform commands. The PDR is built at compile time via the DTS, though EID/TID are run-time settable.

tests:

There are associated unit tests on the PDR builder in this PR. These tests will run in unit-tests on every PR
Furthermore/ system loopback tests on the STM U3 work as expected which verifies PDR building/verification from a testbench MCTP-over-I3C controller.


drivers/pmci/pldm/pldm_pdr.c
```diff
@@ -0,0 +1,614 @@
...
+		.channel_index = DT_PROP(node_id, channel_index),                                  \
+		.sensor_id = DT_PROP(node_id, sensor_id),                                          \
+		.entity_type = DT_PROP_OR(node_id, entity_type, 135),                              \
+		.entity_instance = DT_PROP_OR(node_id, entity_instance, 1),                        \
+		.container_id = DT_PROP_OR(node_id, container_id, 0),                              \
+		.base_unit = COND_CODE_1(DT_NODE_HAS_PROP(node_id, base_unit),                     \
+					(DT_PROP(node_id, base_unit)),                             \
+					(PLDM_PDR_BASE_UNIT_FROM_CHANNEL_TYPE(                     \
+						DT_PROP(node_id, channel_type)))),                 \
+			 .unit_modifier = DT_PROP(node_id, unit_modifier),                         \
```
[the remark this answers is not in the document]
jgrowdenTT: this is incorrect.

drivers/pmci/pldm/pldm_pdr.c
```diff
@@ -0,0 +1,614 @@
...
+		.channel_type = DT_PROP(node_id, channel_type),                                    \
+		.channel_index = DT_PROP(node_id, channel_index),                                  \
+		.sensor_id = DT_PROP(node_id, sensor_id),                                          \
+		.entity_type = DT_PROP_OR(node_id, entity_type, 135),                              \
+		.entity_instance = DT_PROP_OR(node_id, entity_instance, 1),                        \
+		.container_id = DT_PROP_OR(node_id, container_id, 0),                              \
+		.base_unit = COND_CODE_1(DT_NODE_HAS_PROP(node_id, base_unit),                     \
+					(DT_PROP(node_id, base_unit)),                             \
+					(PLDM_PDR_BASE_UNIT_FROM_CHANNEL_TYPE(                     \
+						DT_PROP(node_id, channel_type)))),                 \
```
alexapostolu, line 692: What's the reasoning for making the unit optional? It seems like making it mandatory would save us from having that really long `PLDM_PDR_BASE_UNIT_FROM_CHANNEL_TYPE` macro

jgrowdenTT: One-time Fat in the driver means we don't have to keep double-specifying in the DTS.

drivers/pmci/pldm/pldm_pdr.c
```diff
@@ -0,0 +1,614 @@
...
+		pdr.supported_thresholds.byte = 0U;
+		pdr.threshold_and_hysteresis_volatility.byte = 0U;
+		pdr.state_transition_interval = 0.0f;
+		pdr.update_interval = 0.0f;
+		pdr.max_readable.value_s32 = INT32_MAX;
+		pdr.min_readable.value_s32 = INT32_MIN;
+		pdr.range_field_format = PLDM_RANGE_FIELD_FORMAT_SINT32;
+		pdr.range_field_support.byte = 0U;
+
+		rc = pldm_pdr_append_record(cfg, data, record_handle, &pdr, sizeof(pdr));
```
alayander: Do we know if the spec requires that the record contain the exact byte sequence (ie. no padding)?
`pldm_numeric_sensor_value_pdr` is libpldm's unpacked type and I believe the struct and sizeof(pdr) could contain some alignment padding

jgrowdenTT: Indeed! fixed, I think. Hardened in UT. Nice catch Alex.

drivers/pmci/pldm/Kconfig
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2026 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+config TT_PMCI_PLDM_PLATFORM
```
jgrowdenTT: This should not be TT scoped, sorry.

alayander (approved):

alexapostolu (approved):

PR #1342, opened 2026-07-14
lib: bh_arc: guard eth alt serdes loading on speed_override

alayander:
Fix alt eth serdes loading to be conditional on speed_override equal to 400.


jgrowdenTT (approved):
As discussed, I'm a-okay with merging this without a corresponding commit in main.

jgrowdenTT (approved):

PR #1344, opened 2026-07-14
.github: workflows: produce-data.yml: only run on main

alexapostolu:
Only produce Ci data on main branch. Sometimes you can trigger stress test on PRs but we don't want to inlcude that in our official CI database.


jgrowdenTT (approved):

psethiTT (commented):
@alexapostolu does it make sense to apply the same change to Metal tests as well? They seem to follow same trigger as HW Soak - Label Triggered on PRs as well as Nightly scheduled on main.

psethiTT (approved):

alexapostolu:
We dont run data for Metal yet, so it's fine for now

PR #1352, opened 2026-07-17
ci: Stop unnecessary actions

jgrowdenTT:
~~Remove the flash before step for our pytest runners. The e2e smoke tests have fixtures that run that are required to run in order to properly load the PR image.~~

That's a lie. We can't control that. 

Also remove the tracing build which is done elsewhere

Also don't cleanup in the e2e-smoke shell script. CI should clean up on the recovery runner. Locally on Dev builds I'm not immediately seeing any issue on re-run just straight on top of pytest.

Tests:

CI smoke

### DIFF
--- a/scripts/ci/run-e2e.sh
+++ b/scripts/ci/run-e2e.sh
@@ -97,13 +97,4 @@ if [[ "$TEST_SET" == *"e2e-flash"* ]]; then
 		--extra-args=SB_CONFIG_BOOT_SIGNATURE_KEY_FILE=\"$KEYFILE\" \
 		-ll DEBUG \
 		$@
-	# Restore a stable DMFW, since the copy flashed by BL2 tests will
-	# leave the DMC flash in a different state than other tests expect
-	# We erase the DMC flash first to ensure no old image fragments remain
-	west build -p always -b $DMC_BOARD $TT_Z_P_ROOT/app/dmc -d $ZEPHYR_BASE/build-dmc \
-		--sysbuild
-	west flash -d $ZEPHYR_BASE/build-dmc --domain mcuboot --erase \
-		--cmd-erase 'flash erase_sector 0 0 last'
-	west flash -d $ZEPHYR_BASE/build-dmc --domain dmc
-	rm -rf $ZEPHYR_BASE/build-dmc
 fi
--- a/app/smc/sample.yaml
+++ b/app/smc/sample.yaml
@@ -97,7 +97,6 @@ tests:
     build_only: true
     sysbuild: true
     tags:
-      - e2e-flash
       - build-ci
     extra_overlay_confs:
       - tracing.conf


alexapostolu (dismissed):

scripts/ci/run-e2e.sh
```diff
@@ -92,18 +92,8 @@ if [[ "$TEST_SET" == *"e2e-flash"* ]]; then
 		--pytest-args=--flash-timeout=240 \
 		--device-serial-pty "$TT_Z_P_ROOT/scripts/smc_console.py -d $CONSOLE_DEV -p" \
 		--failure-script "$TT_Z_P_ROOT/scripts/smc_test_recovery.py --asic-id $ASIC_ID" \
-		--flash-before \
 		--outdir $ZEPHYR_BASE/twister-e2e-flash \
 		--extra-args=SB_CONFIG_BOOT_SIGNATURE_KEY_FILE=\"$KEYFILE\" \
 		-ll DEBUG \
 		$@
-	# Restore a stable DMFW, since the copy flashed by BL2 tests will
```
jgrowdenTT, line 100: I'm concerned I'm missing a case which this is covering for local tests, but on dev builds it's not obvious to me.

I never run thru twister anyways, and this seems to hold up in CI...

alexapostolu: It seems fine to remove since it's passing CI

scripts/ci/run-stress.sh
```diff
@@ -84,7 +84,6 @@ if [[ "$TEST_SET" == *"e2e-stress"* ]]; then
         --device-flash-timeout 120 \
         --pytest-args=--flash-timeout=120 \
         --device-serial-pty "$TT_Z_P_ROOT/scripts/smc_console.py -p -d $CONSOLE_DEV" \
-        --flash-before \
```
alexapostolu, line 87: Can you clarify why you removed this? If we don't need `--flash-before` then I'm fine with removing it, but I don't think this removal is related to the justification giving in the commit message

>   --flash-before        Flash device before attaching to serial port. This is useful for devices that share the same port for programming and serial console, or
                        use soft-USB, where flash must come first. Also, it skips reading remaining logs from the old image run.

jgrowdenTT: You're right. I'm misinterpreting the flag. Thanks Alex.

I guess we put it back; but the issue is that whatever we flash here is in a limbo state with an old DMC and new SMC. 

I"ll think about this more but put back the flag so the rest of this can go in.

jgrowdenTT: ... Do we abandon twister for e2e?

scripts/ci/run-e2e.sh
```diff
@@ -91,19 +91,10 @@ if [[ "$TEST_SET" == *"e2e-flash"* ]]; then
...
-	# leave the DMC flash in a different state than other tests expect
-	# We erase the DMC flash first to ensure no old image fragments remain
-	west build -p always -b $DMC_BOARD $TT_Z_P_ROOT/app/dmc -d $ZEPHYR_BASE/build-dmc \
-		--sysbuild
-	west flash -d $ZEPHYR_BASE/build-dmc --domain mcuboot --erase \
-		--cmd-erase 'flash erase_sector 0 0 last'
-	west flash -d $ZEPHYR_BASE/build-dmc --domain dmc
-	rm -rf $ZEPHYR_BASE/build-dmc
 fi
```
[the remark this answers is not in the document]
alexapostolu: Not true since we have recovery as a pre-hook

scripts/ci/run-stress.sh
alexapostolu: Yea I'm not sure how to solve that problem... I have to think about it more too

alexapostolu (dismissed):

alexapostolu (approved):

alayander (approved):

psethiTT (approved):

PR #1356, opened 2026-07-21
lib: bh_arc: Add per-instance GDDR soft-harvest mask

kaiqihuTT:
The product-spec harvesting path can only disable a count of GDDR instances (dram_disable_count).

Added a soft_harvest_dram_mask field to the fw_table DramTable and apply it in CalculateHarvesting: each set bit clears the corresponding channel from tile_enable.gddr_enabled.

The field defaults to 0 (no-op) when a board's fw_table omits it, enabling testing of harvested-DRAM configs for future bin6 BH Galaxy parts which allow one DRAM to be harvested.

the code changes are tested by first rebuilding the bh-mod (with the bh-mod side changes) and trying to set, reset and get the soft_harvest_dram_mask field and logging in telemetry.c to verify that the value changed.

### DIFF
--- a/drivers/misc/bh_fwtable/spirom_protobufs/fw_table.proto
+++ b/drivers/misc/bh_fwtable/spirom_protobufs/fw_table.proto
@@ -62,6 +62,7 @@ message FwTable {
   message DramTable {
     uint32 dram_mask = 1;
     bool dram_mask_en = 2;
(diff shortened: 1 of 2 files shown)


jgrowdenTT:
Could you please outline how you tested this in the PR description? Thanks Alex!

afongTT (approved):

PR #1358, opened 2026-07-21
multi_bootfs: Use prebuilt binaries for MCUBoot and Recovery

alexapostolu:
This is the first commit that separates static and mutable FDs into
distinct tables.

We want tt-flash to only update the static table when it detects a byte
difference (only will occur when firmware intentionally updates MCUBoot
or Recovery images). This is not possible with our current architecture
as our signing process creates non-reproducable binaries even if the
source code is the same.

The solution is to prebuild the FDs that will live in the static table,
which is what this commit does by providing prebuilt binaries for
MCUBoot and Recovery.

Another solution is for tt-flash to compare the payload for the
binaries, ignoring the signing stuff in the header. But this will take
more work on the tools side, so prebuild the binaries for now.

Script used to generate blobs: #1361


zephyr/module.yml
```diff
@@ -241,3 +241,183 @@ blobs:
           raw/refs/heads/main/zephyr/blobs/grendel-sival-sources.tar.gz"
     description: "Grendel Sival sources"
     doc-url: https://tenstorrent.com
+  - path: tt_blackhole_mcuboot_P100A.bin
+    sha256: 3ba6707a89597de402cc6faed8d4a2925df7a384ca67cd63964fa82b02153ff4
+    type: img
+    version: '19.12.0-rc1'
```
jgrowdenTT: Could we take from an actual release for all these?

scripts/fetch_boot_blobs.py
```diff
@@ -0,0 +1,218 @@
+#!/usr/bin/env python3
+
```
jgrowdenTT: This could be split into a different PR.

zephyr/module.yml
alexapostolu: Good catch, I forgot to change it when 19.12.0 was released

scripts/fetch_boot_blobs.py
alexapostolu: https://github.com/tenstorrent/tt-system-firmware/pull/1361

alayander (approved):

jgrowdenTT (approved):
@psethiTT Per offline discussion, please Defer merge until we branch for the RC. Then it's the next release manager's job.

PR #1361, opened 2026-07-22
scripts: fetch MCUBoot and Recovery blobs from a release

alexapostolu:
Add a script to fetch MCUBoot and Recovery blobs from a release. Also has the option to write it to the west.yml.

Tests: Used this script to generate the binaries here, #1358

### DIFF
--- a/scripts/fetch_boot_blobs.py
+++ b/scripts/fetch_boot_blobs.py
@@ -1,0 +1,294 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Regenerate the signed MCUBoot and recovery blobs from a tt-system-firmware
+release.
+
+The MCUBoot and recovery images under `zephyr/blobs/` are not built from source
+They are the officially signed images published on the GitHub releases page.
+This script pulls a given release, extracts those images from each board's bundle,
+and writes them back out as `zephyr/blobs/tt_blackhole_{mcuboot,recovery}_<BOARD>.bin`.
+
+For each board the `<board>.fwbundle` release asset is downloaded and its
+tt-boot-fs image is parsed. The `cmfw` entry is the signed MCUBoot bootloader
+and the `safeimg` entry is the signed recovery image.
+
+Usage:
+    # Regenerate all board blobs from a release
+    python scripts/fetch_boot_blobs.py 19.12.0-rc2
+
+    # Regenerate and update the sha256/version fields in zephyr/module.yml too
+    python scripts/fetch_boot_blobs.py 19.12.0-rc2 --write
+"""
+
+from __future__ import annotations
+
+import argparse
+import hashlib
+import json
+import logging
+import os
+import re
+import shutil
+import sys
+import tarfile
+import tempfile
+import urllib.error
+import urllib.request
+from pathlib import Path
+import yaml
+
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+import tt_boot_fs
+
+_logger = logging.getLogger(__name__)
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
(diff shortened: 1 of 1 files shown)


scripts/fetch_boot_blobs.py
```diff
@@ -0,0 +1,302 @@
...
+REPO_ROOT = Path(__file__).resolve().parents[1]
+MODULE_YAML = REPO_ROOT / "zephyr" / "module.yml"
+DEFAULT_RELEASE = "19.12.0"
+
+MCUBOOT_TAG = "cmfw"
+RECOVERY_TAG = "safeimg"
+
+# Map each release fwbundle asset stem to the board suffix used in the blob
+# file names (tt_blackhole_mcuboot_<SUFFIX>.bin / _recovery_<SUFFIX>.bin).
+BOARDS: dict[str, str] = {
```
jgrowdenTT: Capitalize maybe?

I'm just thinking I don't want to update yet another thing when we get galaxy 2... grendel... etc.

scripts/fetch_boot_blobs.py
```diff
@@ -0,0 +1,302 @@
...
+import yaml
+
+sys.path.append(str(Path(__file__).resolve().parent))
+import tt_boot_fs
+
+_logger = logging.getLogger(__name__)
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+MODULE_YAML = REPO_ROOT / "zephyr" / "module.yml"
+DEFAULT_RELEASE = "19.12.0"
```
jgrowdenTT: Probably just have the arg required.

scripts/fetch_boot_blobs.py
```diff
@@ -0,0 +1,302 @@
...
+            ):
+                failures.append(board)
+        except Exception as e:  # noqa: BLE001 - report and continue with other boards
+            _logger.error("  failed: %s", e)
+            failures.append(board)
+
+    if args.write and computed:
+        # Only write when every requested board was fetched cleanly, so we never
+        # record a sha256 for a blob that failed to download/extract.
+        if failures:
```
jgrowdenTT: It's a little confusing, maybe I'm mis-reading.

We fetch every board we can no matter what, even if one fails, but we record the failure in a list.

But then, once we record the failure, not matter if `args.write and computed` or otherwise, the first thing we do is quit?

A) I think this resolves to one failure case before we check args.write and computed; the string delta doesn't look like it specifically helps me debug this. Maybe I'm wrong
B) But then if we're saying that, why don't we bail when we originally throw the except? (Just so we can collect the full list of everything that fails to download/process? I don't mind that answer, just confirming.)

scripts/fetch_boot_blobs.py
```diff
@@ -0,0 +1,302 @@
...
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        allow_abbrev=False,
+    )
+    parser.add_argument(
+        "release",
+        nargs="?",
+        default=DEFAULT_RELEASE,
+        help=f"Release version to fetch blobs from (default: {DEFAULT_RELEASE}).",
+    )
+    parser.add_argument(
```
jgrowdenTT: What's the use case of fetching just 1  (vs scrapping the parameter and always fetching all)

scripts/fetch_boot_blobs.py
```diff
@@ -0,0 +1,302 @@
...
+    result: dict[str, str] = {}
+    for blob in data.get("blobs", []):
+        path = blob.get("path")
+        sha = blob.get("sha256")
+        if path and sha:
+            result[Path(path).name] = sha
+    return result
+
+
+def normalize_version(release: str) -> str:
```
jgrowdenTT: Arguably unnecessary, just always take in `19.12.0` for example.

jgrowdenTT (dismissed):
Looks good. Appreciate if you could clarify your tests in the PR description.

PR #1362, opened 2026-07-22
tests: add DMA burst tests that uses all 16 channels

tpattanasuttinontTT:
Add a modified version of zephyr's `chan_blen_len` transfer test that uses 16 channels instead of 1


jgrowdenTT (changes requested):
You need to take out the first commit before merge since that's handled in a different PR.

We're just in the middle of trying to get a release out the door, but once that's done, I'll try to make sure your other PR gets merged in short order. Thanks for your patience!

jgrowdenTT:
Just out of curiosity, did we try upstreaming this? Is this more or less a copy paste of the upstream blen test over multiple channels?

If the delta between this and the base BLEN test is just a KCONFIG I wonder if they'd accept.

tpattanasuttinontTT:
Understood, will take out the first commit. It is more or less a copy of the upstream test. I haven't tried to upstream it but I will then! Thank you

PR #1364, opened 2026-07-22
tests: add DMA link tests that uses all 16 channels

tpattanasuttinontTT:
Add a modified version of zephyr's `chan_link_transfer` transfer test that uses 16 channels. (channels 0 & 1, 2 & 3... are linked together)


jgrowdenTT (changes requested):
Have to wait for the other PRs.

I assume however this is only blocked on the DMA side changes. you don't need to be on top of the `all_chan_blen_transfer` PR. theoretically anyways.

PR #1380, opened 2026-07-27
bh: fw_table: add soft GDDR harvest mask override

kaiqihuTT:
Add dram_table.soft_harvest_dram_mask to the BH fw_table and CCFGOVR override flow so specific GDDR channels can be soft-harvested at runtime. This updates the fw_table and override protobufs, applies the override in tt_bh_fwtable_apply_ccfgovr(), and feeds the mask into harvesting by clearing the selected bits from tile_enable.gddr_enabled. 

Testing: used bh-mod set options to verify that setting different dram soft harvest masks will apply the new mask 
in fw_table. Verified via telemetry logging

### DIFF
--- a/drivers/misc/bh_fwtable/bh_fwtable.c
+++ b/drivers/misc/bh_fwtable/bh_fwtable.c
@@ -394,6 +394,14 @@ void tt_bh_fwtable_apply_ccfgovr(const struct device *dev)
 				ovr.feature_enable.kernel_throttler_at_floor_en;
 		}
 
+		if (ovr.has_dram_table && ovr.dram_table.has_soft_harvest_dram_mask) {
+			LOG_INF("CCFGOVR override: dram_table.soft_harvest_dram_mask = 0x%x",
+				ovr.dram_table.soft_harvest_dram_mask);
+			data->fw_table.dram_table.soft_harvest_dram_mask =
+				ovr.dram_table.soft_harvest_dram_mask;
+			data->fw_table.has_dram_table = true;
+		}
+
 		return;
 	}
 
--- a/drivers/misc/bh_fwtable/spirom_protobufs/fw_table_override.proto
+++ b/drivers/misc/bh_fwtable/spirom_protobufs/fw_table_override.proto
@@ -19,10 +19,11 @@ message FwTableOverride {
    * (FwTable field 1) is deliberately reserved here so tt-mod cannot forge
    * a bundle version.
    */
-  reserved 1, 4, 5, 6, 7, 8, 9, 10;
+  reserved 1, 4, 6, 7, 8, 9, 10;
 
   optional ChipLimits chip_limits = 2;
   optional FeatureEnable feature_enable = 3;
+  optional DramTable dram_table = 5;
 
   message ChipLimits {
 
@@ -37,5 +38,12 @@ message FwTableOverride {
     reserved 1, 2, 3, 4, 5, 6, 7, 8, 9;
 
     optional bool kernel_throttler_at_floor_en = 10;
+  }
+
+  message DramTable {
+
+    reserved 1, 2;
+
+    optional uint32 soft_harvest_dram_mask = 3;
   }
 }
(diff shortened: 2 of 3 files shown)


lib/tenstorrent/bh_arc/harvesting.c
```diff
@@ -238,6 +238,14 @@ static int CalculateHarvesting(void)
 			tile_enable.eth_enabled = 0;
 		}
 
+		/* Soft-harvest specific GDDR instances requested via the fw_table.
+		 * A set bit in soft_harvest_dram_mask forces that channel harvested.
+		 */
+		if (tt_bh_fwtable_get_fw_table(fwtable_dev)->has_dram_table) {
+			tile_enable.gddr_enabled &= ~tt_bh_fwtable_get_fw_table(fwtable_dev)
+							     ->dram_table.soft_harvest_dram_mask;
+		}
```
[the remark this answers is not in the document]
kaiqihuTT: Not a concern. After bitwise AND, only the least sig 8 bits gets assigned back to gadder_enable, maintain it's uint8_t type

jgrowdenTT:
Appreciate if you could give a better title, outline how you tested, link to a sys ticket (if applicable).

Thanks!

You still need to outline, in the PR description, how you tested it, and link it to your sys ticket. Thanks Alex.

drivers/misc/bh_fwtable/spirom_protobufs/fw_table_override.proto
```diff
@@ -19,10 +19,12 @@ message FwTableOverride {
    * (FwTable field 1) is deliberately reserved here so tt-mod cannot forge
    * a bundle version.
    */
-  reserved 1, 4, 5, 6, 7, 8, 9, 10;
+  reserved 1, 4, 6, 7, 8, 9;
```
projkovTT: Should we be reusing reserved field numbers? Why not use new ones

alexapostolu: Good question - This is the override table (`fw_table_override.proto`), and the way we designed the override table was to "reserve" all the fields of the original table until we needed them. So if we want to expose a new field, we would unreserve that field in the override proto table.

jgrowdenTT:
Can you please outline your tests, and link to a sys ticket if applicable. Thanks!

drivers/misc/bh_fwtable/bh_fwtable.c
```diff
@@ -394,6 +394,22 @@ void tt_bh_fwtable_apply_ccfgovr(const struct device *dev)
...
+			data->fw_table.dram_table.soft_harvest_dram_mask =
+				ovr.dram_table.soft_harvest_dram_mask;
+			data->fw_table.has_dram_table = true;
+		}
+
+		if (ovr.has_product_spec_harvesting &&
+		    ovr.product_spec_harvesting.has_dram_disable_count) {
+			LOG_INF("CCFGOVR override: product_spec_harvesting.dram_disable_count = %u",
+				ovr.product_spec_harvesting.dram_disable_count);
+			data->fw_table.product_spec_harvesting.dram_disable_count =
```
jgrowdenTT: Can't we infer this from the bitmask?

Not a big deal, but isnt' this wrong if something is harvested in efuse? 

@afongTT?

alexapostolu (commented):
Why can't `dram_mask` be used? Could you add a comment to differentiate the purpose of `dram_mask` and `soft_harvest_dram_mask`?

NVM - as discussed in standup there is a potential overlap here but we won't worry about it for this release

afongTT:
> Why can't `dram_mask` be used? Could you add a comment to differentiate the purpose of `dram_mask` and `soft_harvest_dram_mask`?

`dram_mask` is GDDR tile enabled, but don't load MRISC FW on to it.
`soft_harvest_dram_mask` is GDDR tile disabled.

drivers/misc/bh_fwtable/bh_fwtable.c
afongTT: The original implementation of `product_spec_harvesting.dram_disable_count` did not take this into account. It would silently continue with e.g. 7 GDDR if requested `dram_disable_count = 0` and efuse harvested one DRAM.

@kaiqihuTT it would be prudent to `LOG_WARN` in harvesting.c when we can't achieve `desired_gddr_count`

afongTT (dismissed):

afongTT (approved):

PR #1382, opened 2026-07-27
.github: ci_boards: upgrade metal to v0.74

psethiTT:
Upgrades the upstream Metal testing container from v0.68.0-dev to v0.74.0
across all Blackhole boards, and switches quietbox2/loudbox to their
dedicated `-qb-ge` and `-loudbox` images. Also uses the 8B Llama model
instead of 70B on p300, which is faster and sufficient for our testing as
confirmed by the Metal team.

CI-only change -- no firmware sources are touched, so the rc3 binaries are
identical to rc2. The purpose of the new RC is to re-run qualification
against the updated Metal container.

Adding label - metal-tests so that metal runs


alexapostolu:
Oh sorry @psethiTT I forgot to create this backport PR. Usually the person that wants a feature also creates the backport branch. But this looks good, thanks

alexapostolu (approved):

PR #1384, opened 2026-07-29
boards: tenstorrent: galaxy_2 Add spi3 node and max7221 subnode

projkovTT:
Add spi3 node to galaxy2 dmc dts.
Add max7221 subnode on spi3 (binds to max7219 driver `zephyr/drivers/display/display_max7219.c`)
- The driver is interchangeable between max7219/max7221.
- Datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/max7219-max7221.pdf

Verification:
- Verified spi3, max7221 nodes exist in compiled device tree
- Verify existence of maxim driver symbols in built `zephr.elf`

Open question: do we want to define a cs-gpio instead of using nss? The display is the only device on SPI3.

### DIFF
--- a/boards/tenstorrent/tt_blackhole_glx2_dmc/tt_blackhole_glx2_dmc.dts
+++ b/boards/tenstorrent/tt_blackhole_glx2_dmc/tt_blackhole_glx2_dmc.dts
@@ -260,6 +260,22 @@
 	};
 };
 
+&spi3 {
+	pinctrl-0 = <&spi3_nss_pa4 &spi3_sck_pc10
+		     &spi3_miso_pb4 &spi3_mosi_pd6>;
+	pinctrl-names = "default";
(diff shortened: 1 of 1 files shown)


jgrowdenTT (approved):

alexapostolu (approved):
Should you also enable `CONFIG_DISPLAY=y` in this PR or is that out of scope?

I think NSS should be fine since this is the only decide in spi3

projkovTT:
> Should you also enable `CONFIG_DISPLAY=y` in this PR or is that out of scope?

Maybe? I was thinking that it could be left for when we start actually using the display driver in firmware. I was building with `-DCONFIG_DISPLAY=y` to test.

> I think NSS should be fine since this is the only decide in spi3

Agreed.

alayander (approved):

PR #1387, opened 2026-07-30
fw_table: add gddr ctf thresholds and enable to fw table

alayander:
Add gddr therm trip thresholds (ie. temp, critical temp and duration) to chip limits and add gddr therm trip enable to feature enable of firmware table. Also add characterization submessage to enable/disable this feature, with default set to disabled.

Follow up PRs will include: KMD logging, ccfgovr support

Tests:
- Characterization submsg enables and disables feature as reported in telemetry
- Enablement causes gddr ctf trip (with modified threshold for test), but does not trip when disabled

Tested on p150a

Resolves: SYS-4907

### DIFF
--- a/lib/tenstorrent/bh_arc/cat.c
+++ b/lib/tenstorrent/bh_arc/cat.c
@@ -17,6 +17,7 @@
 #include <zephyr/kernel.h>
 #include <zephyr/sys/util.h>
 #include <zephyr/drivers/gpio.h>
+#include <zephyr/drivers/misc/bh_fwtable.h>
 #include <zephyr/init.h>
 #include <zephyr/sys/util.h>
 #include <zephyr/drivers/sensor.h>
@@ -55,6 +56,15 @@ typedef union {
 #ifndef CONFIG_TT_SMC_RECOVERY
 
 static const int gddr_therm_trip_interval = 100;
+
+static bool gddr_therm_trip_enabled;
+
+/* Live trip thresholds, seeded from chip_limits at init */
+static int gddr_therm_trip_temp;
+static int gddr_therm_trip_critical_temp;
+static int64_t gddr_therm_trip_duration_ms;
+
+static const struct device *const fwtable_dev = DEVICE_DT_GET(DT_NODELABEL(fwtable));
 
 static const struct device *const pvt = DEVICE_DT_GET(DT_NODELABEL(pvt));
 
@@ -208,18 +218,19 @@ static int CATInit(void)
 }
 SYS_INIT_APP(CATInit);
 
-int MonitorGddrThermTrip(int64_t now, int max_temp)
+int EvaluateGddrThermTrip(int64_t now, int max_temp, int trip_temp, int critical_temp,
+			  int64_t duration_ms)
 {
 	static bool tracking;
 	static int64_t over_temp_start_time;
 
-	if (max_temp >= CAT_GDDR_THERM_TRIP_CRITICAL_TEMP) {
-		LOG_ERR("Max GDDR temp %dC >= %dC", max_temp, CAT_GDDR_THERM_TRIP_CRITICAL_TEMP);
+	if (critical_temp != 0 && max_temp >= critical_temp) {
+		LOG_ERR("Max GDDR temp %dC >= %dC", max_temp, critical_temp);
 		tracking = false;
 		return max_temp;
 	}
 
-	if (max_temp < CAT_GDDR_THERM_TRIP_TEMP) {
+	if (trip_temp == 0 || max_temp < trip_temp) {
 		tracking = false;
 		return 0;
 	}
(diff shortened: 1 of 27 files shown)


projkovTT (dismissed):

jgrowdenTT:
Can we test to verify that if catmon is enabled here, the KMD log message gets out?

If it doesn't get out I don't think it's blocking the PR, but we'll need a follow up ticket to address.

lib/tenstorrent/bh_arc/cat.c
```diff
@@ -256,7 +272,18 @@ static void gddr_therm_trip_trigger_handler(struct k_work *work)
 }
 
 static K_WORK_DELAYABLE_DEFINE(gddr_therm_trip_trigger_work, gddr_therm_trip_trigger_handler);
-#endif /* CONFIG_TT_BH_ARC_GDDR_THERM_TRIP_ACTION */
+
+uint8_t CatSetGddrThermTripEnabled(uint32_t enabled)
```
jgrowdenTT: bool param

lib/tenstorrent/bh_arc/telemetry.h
```diff
@@ -70,8 +70,17 @@ typedef struct {
...
 	uint32_t kernel_nops_at_aiclk_fmin: 1;
 
+	/** @brief GDDR thermal-trip action feature.
+	 *
+	 * When enabled, sustained or critical GDDR over-temperature engages a
+	 * CATMON thermal trip (shutting the ASIC down). This bit is controlled
+	 * at runtime by the host with @ref TT_SMC_MSG_CHARACTERISATION using
+	 * @ref TT_SUB_MSG_SET_GDDR_THERM_TRIP_ENABLED.
+	 */
+	uint32_t gddr_therm_trip: 1;
```
jgrowdenTT, line 80: I see this updated, but it's not obvious to me where this is used. This should be the source of truth variable about whether or not we are going to perform the catmon stuff, not just for tracking. 

i.e. if I'm in GDB and set this bit high, the feature is enabled.

jgrowdenTT (changes requested):
Just using the feature bit as the global tracking is the feedback I'd like dealt with now. Otherwise should be good to merge.

alayander:
Addressed the feature bit as global tracking, I also have a follow up ticket (SYS-4905) to add the KMD logging

alexapostolu:
Any LOG_* will automatically go through KMD logging / dmesg, so I don't think there's anything you need to manually do for that

jgrowdenTT:
> Any LOG_* will automatically go through KMD logging / dmesg, so I don't think there's anything you need to manually do for that

Agreed, sorry for my lack of clarity. What I'm specifically concerned with is that if `cat.c` downs the device/the system and we don't msg the log out, we're then left with a device inaccessible via PCIE with nothing in dmesg indicating why, and a long hunt a head of us in figuring out why. We need that message visible before we down the system.

lib/tenstorrent/bh_arc/cat.c
```diff
@@ -208,36 +216,42 @@ static int CATInit(void)
...
 
 	if (!tracking) {
 		tracking = true;
 		over_temp_start_time = now;
-	} else if (now - over_temp_start_time >= CAT_GDDR_THERM_TRIP_DURATION_MS) {
-		LOG_ERR("Max GDDR temp %dC >= %dC for >= %d ms", max_temp, CAT_GDDR_THERM_TRIP_TEMP,
-			CAT_GDDR_THERM_TRIP_DURATION_MS);
+	} else if (now - over_temp_start_time >= duration_ms) {
+		LOG_ERR("Max GDDR temp %dC >= %dC for >= %lld ms", max_temp, trip_temp,
+			(long long)duration_ms);
```
jgrowdenTT, line 241: Fine for now but I think there's macros for porting this to u64 we probably need to use.

lib/tenstorrent/bh_arc/telemetry.c
```diff
@@ -260,6 +260,16 @@ void UpdateTelemetryKernelThrottler(bool enabled, uint32_t stop_nops_freq)
 	telemetry[TAG_KERNEL_THROTTLER] = (enabled ? 1U : 0U) | ((stop_nops_freq & 0xFFFFU) << 16U);
 }
 
+void UpdateTelemetryGddrThermTrip(bool enabled)
+{
+	telemetry_feature_flags_0_t active_config = {
+		.u32_all = telemetry[TAG_FW_ACTIVE_CONFIG_0],
+	};
+
+	active_config.bits.gddr_therm_trip = enabled ? 1U : 0U;
```
jgrowdenTT, line 269: Sorry, where are you using this to disable/enable the feature?

alayander: On line 292 of `lib/tenstorrent/bh_arc/cat.c`, the if statement for communicating with DMC and triggering the therm trip is based on the active config.
(https://github.com/tenstorrent/tt-system-firmware/pull/1387/changes#diff-86c69b71fc837df2d517062782de16d8ba7ba8381e5d015a3185d49db2403f71R292)

SMC will still log the event via `LOG_ERR` if gddr therm trip action is disabled but the criteria for a trip was met

jgrowdenTT (approved):

lib/tenstorrent/bh_arc/telemetry.c
jgrowdenTT: Thanks, sorry I missed it!

projkovTT (approved):

PR #1389, opened 2026-07-30
boards: tenstorrent: Add second SPI mux for galaxy 2

projkovTT:
Add second SPI mux, `spi_mux1` for galaxy 2 which allows board to board (B2B) to access the flash.
mux0 (`spi_mux`) is the same as on blackhole PCIe cards.

MUX truth table (0 = low, active)
```
MUX0 | MUX1 | Route
0    |    x | flash → BH
1    |    0 | DMC → flash
1    |    1 | B2B → flash
```

Note that the gpio hog (pre-init) on blackhole PCIe cards (and on galaxy 2) is physical HIGH, ie. connects DMC to flash.
`main.c::main` then connects flash to SMC.

### DIFF
--- a/include/tenstorrent/bh_chip.h
+++ b/include/tenstorrent/bh_chip.h
@@ -28,6 +28,7 @@ struct bh_chip_config {
 	struct gpio_dt_spec asic_reset;
 	struct gpio_dt_spec spi_reset;
 	struct gpio_dt_spec spi_mux;
+	struct gpio_dt_spec spi_mux1;
 	struct gpio_dt_spec pgood;
 	struct gpio_dt_spec therm_trip;
 	const struct device *flash;
@@ -147,6 +148,9 @@ extern struct bh_chip BH_CHIPS[BH_CHIP_COUNT];
 			.spi_mux = GPIO_DT_SPEC_GET(                                               \
 				DT_PHANDLE_OR_CHILD(DT_PHANDLE_BY_IDX(n, prop, idx), spi_mux),     \
 				gpios),                                                            \
+			.spi_mux1 = GPIO_DT_SPEC_GET_OR(                                           \
+				DT_PHANDLE_OR_CHILD(DT_PHANDLE_BY_IDX(n, prop, idx), spi_mux1),    \
+				gpios, {0}),                                                       \
 			.pgood = GPIO_DT_SPEC_GET(                                                 \
 				DT_PHANDLE_OR_CHILD(DT_PHANDLE_BY_IDX(n, prop, idx), pgood),       \
 				gpios),                                                            \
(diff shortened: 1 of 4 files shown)


lib/tenstorrent/bh_chip/Kconfig
```diff
@@ -18,6 +18,13 @@ config TT_I2C_STRAP_INIT_PRIORITY
 	  to initialize before the I2C bus, so that GPIO expanders used
 	  for strapping can be properly configured during system boot
 
+config TT_BH_CHIP_SPI_MUX1
```
jgrowdenTT: mmmm... I thought DTS should be able to feed this.

jgrowdenTT: Something like` DT_NODE_HAS_STATUS(DT_ALIAS(spi_mux1), okay)`

projkovTT: Updated in https://github.com/tenstorrent/tt-system-firmware/pull/1389/commits/3e8989439ee682b5b08fe76173990a4ad09a7c18. I couldn't conditionally add spi_mux1 to the struct/init/main.c in precompilation in a way that I found satisfying (but I also don't know so much about zephyr's macros). Instead, spi_mux1 is now always present in the config struct but initialized to {0} if the node is not present in the DT.
Let me know if this is a better design, I'm not too sure.

jgrowdenTT (dismissed):

include/tenstorrent/bh_chip.h
```diff
@@ -147,6 +149,12 @@ extern struct bh_chip BH_CHIPS[BH_CHIP_COUNT];
 			.spi_mux = GPIO_DT_SPEC_GET(                                               \
 				DT_PHANDLE_OR_CHILD(DT_PHANDLE_BY_IDX(n, prop, idx), spi_mux),     \
 				gpios),                                                            \
+			.spi_mux1 = COND_CODE_1(                                                   \
+		HAS_DT_PHANDLE_OR_CHILD(DT_PHANDLE_BY_IDX(n, prop, idx), spi_mux1),        \
+		(GPIO_DT_SPEC_GET(                                                         \
+			DT_PHANDLE_OR_CHILD(DT_PHANDLE_BY_IDX(n, prop, idx), spi_mux1),        \
+			gpios)),                                                               \
+		({ 0 })),                                                                       \
```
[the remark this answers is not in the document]
projkovTT: {0} is the right way to do it. But I've just recently discovered `GPIO_DT_SPEC_GET_OR` which allows default assignment without using `COND_CODE_1` so I'll update that.

alexapostolu (approved):

jgrowdenTT (approved):

PR #1393, opened 2026-08-04
drivers: pldm: tt_oem

jgrowdenTT:
Add pldm OEM API. Add TT implementation for the OEM API which opens up zephyr shell capabilities


drivers/pmci/pldm/Kconfig
```diff
@@ -20,3 +20,13 @@ config PMCI_PLDM_BASE
 	depends on PLDM
 	help
 	  Enable PLDM BASE support
+
+config TT_PMCI_PLDM_SHELL
+	bool "Enable PLDM OEM shell bridge"
+	depends on DT_HAS_TENSTORRENT_PLDM_OEM_HANDLER_ENABLED
+	depends on SHELL
+	select SHELL_BACKEND_DUMMY
+	help
```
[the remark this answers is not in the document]
jgrowdenTT: I'll add OEM. BASE drags in more than I'd like right now so I'm going to leave it out.

drivers/pmci/pldm/tt_pldm_oem.c
```diff
@@ -0,0 +1,401 @@
...
+#define TT_PLDM_SHELL_MAX_OUTPUT_LEN   256U
+#define TT_PLDM_SHELL_MAX_CHUNK_LEN    40U
+
+struct tt_pldm_shell_context {
+	struct k_mutex lock;
+	struct k_work_q work_q;
+	struct k_work work;
+
+	K_KERNEL_STACK_MEMBER(work_q_stack, TT_PLDM_SHELL_WORKQ_STACK_SIZE);
+	bool initialized;
```
alayander: nit: Would this bool be already covered by checking `device_is_ready()` on the driver?

drivers/pmci/pldm/tt_pldm_oem.c
```diff
@@ -0,0 +1,401 @@
...
+				       const struct pldm_msg *req_msg, size_t req_payload_len,
+				       struct pldm_msg *resp_msg, size_t *resp_pldm_len)
+{
+	const uint8_t *payload = req_msg->payload;
+	uint8_t *resp_payload = resp_msg->payload;
+	uint16_t request_id;
+	uint8_t cc = PLDM_SUCCESS;
+	int rc;
+
+	ARG_UNUSED(req_msg);
```
alayander: nit: This arg is used above when defining the `payload`

alayander (dismissed):

alayander (dismissed):

include/zephyr/drivers/pmci/pldm/pldm_oem_handler.h
```diff
@@ -0,0 +1,119 @@
...
+	 */
+	int (*build_response)(const struct device *dev, const struct pldm_header_info *req_hdr,
+			      const struct pldm_msg *req_msg, size_t req_payload_len,
+			      struct pldm_msg *resp_msg, size_t *resp_pldm_len);
+};
+
+static inline const ver32_t *pldm_oem_handler_get_versions(const struct device *dev,
+							   size_t *versions_size)
+{
+	if (dev == NULL || !device_is_ready(dev)) {
```
alexapostolu: We don't need `dev == NULL` since `DEVICE_API_GET` already asserts that, https://docs.zephyrproject.org/latest/doxygen/html/device_8h.html#a9907e8ee5e7be8c3b5d52b13bd33970a

include/zephyr/drivers/pmci/pldm/pldm_oem_handler.h
```diff
@@ -0,0 +1,119 @@
...
+
+static inline const bitfield8_t *pldm_oem_handler_get_commands(const struct device *dev)
+{
+	if (dev == NULL || !device_is_ready(dev)) {
+		return NULL;
+	}
+
+	const struct pldm_oem_handler_driver_api *api = DEVICE_API_GET(pldm_oem_handler, dev);
+
+	if (api == NULL || api->get_commands == NULL) {
```
alexapostolu: I think `api == NULL` is redundant since DEVICE_API_GET won't work if API is NULL, https://docs.zephyrproject.org/latest/doxygen/html/device_8h.html#a48c6030c2e7d1d05ace7c708dda11949

include/zephyr/drivers/pmci/pldm/pldm_oem_handler.h
jgrowdenTT: Sure, but because `device_is_ready` already checks it.

Note the ASSERT is not compiled in release builds.  I'd rather an ENOSYS return than a nullptr deref.

include/zephyr/drivers/pmci/pldm/pldm_oem_handler.h
jgrowdenTT: I'll remove it, but again I think the guards you're referring to are only present in debug builds. If we're staring at a nullptr deref in prod I think I'd rather the enosys return. But precedent upstream is as you say, it appears.

alexapostolu: Ah sorry, I just remembered we had this conversation before.
But if I recall last time the reason we wanted to return an error was cause there was potential that we would serialize the API and that way the parameters could change and runtime and we would want an non-fatal error if the parameter is wrong. Is that true for here as well?

drivers/pmci/pldm/tt_pldm_oem.c
```diff
@@ -0,0 +1,401 @@
...
+	const uint8_t *payload = req_msg->payload;
+	uint8_t *resp_payload = resp_msg->payload;
+	uint16_t request_id;
+	uint8_t cmd_len;
+	uint8_t cc = PLDM_SUCCESS;
+	int rc;
+
+	if (req_payload_len < 3U) {
+		return pldm_cc_only_response(req_hdr, PLDM_ERROR_INVALID_LENGTH, resp_msg,
+					     resp_pldm_len);
```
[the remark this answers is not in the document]
projkovTT: Is this a valid concern?

projkovTT (approved):

alayander (approved):

drivers/pmci/pldm/tt_pldm_oem.c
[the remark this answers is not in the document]
jgrowdenTT: I mean if it doesn't link ya can't run anything :P 

More of a build help than a runtime fix.

PR #1397, opened 2026-08-05
lib: bh_arc: use zephyr reset controller driver

projkovTT:
### Changes:
Update Global, Tensix, DDR, ETH, L2CPU resets to use `reset_tt_bh.c` driver which implements the zephyr reset controller `zephyr/drivers/reset.h` as well as BH specific `zephyr/drivers/reset/reset_tt_bh.h`.

BH specific reset behaviour is for controlling multiple resets lines at the same time, eg. `reset_tt_bh_lines_deassert`, which the Zephyr driver does not support. SYS-1553 asks to add this to upstream API - I am mixed on doing this.

Move base addresses, masks, other configuration used for resets from `bh_arc/init.h` into device tree `tt_blackhole_smc.dtsi`.

Create emul reset driver for native sim.

Add smoke `test_reset_unit_post_init` which verifies reset register values after init.

### Reason for change:
Potability: The driver resolves configuration via the DTS rather than hard coded #defines.
Overall code cleanliness: We have a reset driver, might as well use it instead of writing out the whole RMW every time we want to reset something.

### Testing:
Smoke on p150a. Stress dirty reset on p150a. 
Moderate risk since resets are kind of important and this PR touches them all.
New `test_reset_unit_post_init` to check reset register values after init.
The other reset tests appear to have good coverage. Room for future work to add L2CPU reset test.

### Ticket:
SYS-1553


include/zephyr/dt-bindings/reset/tt-bh-reset.h
projkovTT, line 1: Questionable addition by Claude. We need this information somewhere but I don't think this is the right place.

projkovTT: So I think this is fine actually

include/zephyr/dt-bindings/reset/tt-bh-reset.h
```diff
@@ -0,0 +1,38 @@
...
+#define TT_BH_L2CPU_TILE_ID(n) ((n) + 0)
+#define TT_BH_L2CPU_RISC_ID(n) ((n) + 4)
+
+/* global_reset @ 0x80030000 bit positions (reset-mask 0x2383) */
+#define TT_BH_GLOBAL_SYSTEM_RESET_ID  0
+#define TT_BH_GLOBAL_NOC_RESET_ID     1
+#define TT_BH_GLOBAL_REFCLK_CNT_EN_ID 7
+#define TT_BH_GLOBAL_PCIE0_RESET_ID   8
+#define TT_BH_GLOBAL_PCIE1_RESET_ID   9
+#define TT_BH_GLOBAL_PTP_RESET_ID     13
```
projkovTT, line 36: I'm honestly not a fan of moving the bitfield definitions out of the structs and into these defines.

The reason to do this is that the zephyr reset driver exposes per-line assert/deassert rather than mask.

However, we prefer to just set the whole field at once and to that end use our multi-line reset anyways which takes a mask as input. Individually defining these lines is contradictory to how we use them, ie. see how we construct the masks wtih #defines at the top of reset.c.

projkovTT: tbh the struct is really only better for the global reset register, I think that what we have here is good for the others. I'm open to suggestions / comments

lib/tenstorrent/bh_arc/reset.c
```diff
@@ -38,12 +38,50 @@
...
+	(BIT(TT_BH_GLOBAL_SYSTEM_RESET_ID) | BIT(TT_BH_GLOBAL_NOC_RESET_ID) |                      \
+	 BIT(TT_BH_GLOBAL_REFCLK_CNT_EN_ID) | BIT(TT_BH_GLOBAL_PCIE0_RESET_ID) |                   \
+	 BIT(TT_BH_GLOBAL_PCIE1_RESET_ID) | BIT(TT_BH_GLOBAL_PTP_RESET_ID))
+
+#define TT_BH_ETH_TILE_MASK   GENMASK(13, 0)
+#define TT_BH_ETH_RISC_MASK   GENMASK(29, 16)
+#define TT_BH_DDR_TILE_MASK   GENMASK(7, 0)
+#define TT_BH_DDR_RISC_MASK   GENMASK(31, 8)
+#define TT_BH_L2CPU_TILE_MASK GENMASK(3, 0)
+#endif /* CONFIG_RESET_TT_BH */
```
alexapostolu: I think instead of wrapping this with `#ifdef CONFIG_RESET_TT_BH`, you can actually remove that ifdef and define the reset node in the native sim overlay `tests/lib/tenstorrent/bh_arc/app.overlay`

lib/tenstorrent/bh_arc/reset.c
```diff
@@ -38,12 +38,50 @@
...
 #define PLL_DEVICE_INIT(inst) DEVICE_DT_INST_GET(inst),
 
 static const struct device *const pll_devs[] = {DT_INST_FOREACH_STATUS_OKAY(PLL_DEVICE_INIT)};
 
+#if defined(CONFIG_RESET_TT_BH)
+static const struct device *const global_reset_dev = DEVICE_DT_GET(DT_NODELABEL(global_reset));
+static const struct device *const eth_reset_dev = DEVICE_DT_GET(DT_NODELABEL(eth_reset));
+static const struct device *const ddr_reset_dev = DEVICE_DT_GET(DT_NODELABEL(ddr_reset));
+static const struct device *const l2cpu_reset_dev = DEVICE_DT_GET(DT_NODELABEL(l2cpu_reset));
+static const struct device *const tensix_reset_devs[] = {
```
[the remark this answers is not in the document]
projkovTT: I put `CONFIG_RESET_TT_BH` in `tt_blackhole_tt_blackhole_smc_defconfig` (whats up with that name btw) alongside other critical configuration defs for this reason. It _should_ be included on any build for real hardware. I don't know if this is the right way to handle this inclusion though - open to suggestions.

include/zephyr/dt-bindings/reset/tt-bh-reset.h
alexapostolu: I think that's fine for now

lib/tenstorrent/bh_arc/reset.c
```diff
@@ -361,37 +395,21 @@ static int DeassertTileResets(void)
...
-
-	eth_reset.f.eth_reset_n = 0x3fff;
-	WriteReg(RESET_UNIT_ETH_RESET_REG_ADDR, eth_reset.val);
-
-	RESET_UNIT_TENSIX_RESET_reg_u tensix_reset = {.val = RESET_UNIT_TENSIX_RESET_REG_DEFAULT};
+	/* Always deassert NOC, system, and PCIe resets - needed for ARC-PCIe communication.
+	 * Absolute writes for dual-field / sparse banks force a known full-word state
+	 * (tile half released, RISC half still held). Full tensix banks use line RMW.
+	 */
+#if defined(CONFIG_RESET_TT_BH)
```
projkovTT: Full word writes here instead of one by one line deasssert derisk this PR by maintaining functional equivalence to what we were doing before. I think that line-by-line _should_ be fine (and is the Zephyr way since that is what their driver defines), but I don't see a good reason to change.

dts/vendor/tenstorrent/tt_blackhole_smc.dtsi
```diff
@@ -656,6 +659,70 @@
...
+			status = "okay";
+		};
+
+		tensix_risc_reset4: reset@80030050 {
+			compatible = "tenstorrent,bh-reset";
+			reg = <0x80030050 0x4>;
+			#reset-cells = <1>;
+			nresets = <32>;
+			status = "okay";
+		};
```
alexapostolu: Instead of having a separate node for each reset controller, you can group them into one node with an offset of `0x4`,

```
/* Tensix RISC reset controllers 0-7 */
tensix_risc_reset: reset@80030040 {
    reg = <0x80030040 0x4>
    ...
};
```

This would require a small change in the driver to take into account the offset, but in the end it should cut down on the number of lines in the DTS files

alexapostolu (commented):
1. For the emulated reset driver -

lib/tenstorrent/bh_arc/eth.c
```diff
@@ -677,16 +682,24 @@ static uint8_t toggle_eth_reset_handler(const union request *req, struct respons
...
+			continue;
+		}
 
-	/* Assert tile and risc reset */
-	eth_reset.f.eth_reset_n &= ~mask;
-	eth_reset.f.eth_risc_reset_n &= ~mask;
-	WriteReg(RESET_UNIT_ETH_RESET_REG_ADDR, eth_reset.val);
+		(void)reset_line_assert(eth_reset_dev, TT_BH_ETH_TILE_ID(eth_inst));
+		(void)reset_line_assert(eth_reset_dev, TT_BH_ETH_RISC_ID(eth_inst));
+	}
```
alexapostolu, line 695: Is the logic here correct? Co-pilot said,

> toggle_eth_reset_handler currently asserts/deasserts reset lines per instance (multiple RMWs), which changes behavior vs the prior single-register write and can introduce transient states (e.g., tile asserted while RISC still deasserted). Since this register is a dual-field bank, prefer single-RMW multi-line operations to keep the reset update atomic and reduce MMIO traffic.
	/* Assert tile and risc reset for each requested instance */
	for (uint8_t eth_inst = 0; eth_inst < MAX_ETH_INSTANCES; eth_inst++) {
		if (!IS_BIT_SET(mask, eth_inst)) {
			continue;
		}

projkovTT: The logic is correct, but the concern about dual-field bank is valid. I had considered atomicity from the perspective of toggling multiple ETH resets in sequence (which is fine) but not toggling tile/risc for a single one which is what co-pilot is concerned about. 
Personally, I think that should also be fine and it hasn't failed any tests to make me think otherwise. If its not, that should definitely be documented!
However, to de-risk I'll change this to a single write setting tile and risc fields.

dts/vendor/tenstorrent/tt_blackhole_smc.dtsi
projkovTT: This was James's suggestion as well, will implement

drivers/reset/reset_tt_bh_emul.c
```diff
@@ -0,0 +1,206 @@
...
+
+#define TT_BH_RESET_EMUL_MASK_FROM_NRESETS(_n)                                                     \
+	((uint32_t)BIT64_MASK(DT_INST_PROP_OR(_n, nresets, 0)))
+#define TT_BH_RESET_EMUL_MASK(_n) DT_INST_PROP_OR(_n, reset_mask, 0)
+#define TT_BH_RESET_EMUL_NUM_SPECIFIERS(_n)                                                        \
+	(!!DT_INST_PROP_OR(_n, nresets, 0) + !!DT_INST_PROP_OR(_n, reset_mask, 0))
+#define TT_BH_RESET_EMUL_N_BANKS(_n) DT_INST_PROP_OR(_n, n_banks, 1)
+
+#define TT_BH_RESET_EMUL_MASK_DEFINE(_n)                                                           \
+	((TT_BH_RESET_EMUL_MASK(_n) | TT_BH_RESET_EMUL_MASK_FROM_NRESETS(_n)) ?: UINT32_MAX)
```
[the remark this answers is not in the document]
jgrowdenTT: Could you look into this one?

app/smc/pytest/e2e_smoke.py
```diff
@@ -597,6 +621,69 @@ def test_boot_status(arc_chip_dut, asic_id):
...
+    """
+    Validates reset-unit register state after SMC init
+    """
+    # Re-init so register state is independent of prior tests
+    assert smi_reset_test(asic_id), "tt-smi reset failed before reset-unit check"
+    arc_chip = pyluwen.detect_chips()[asic_id]
+
+    err = arc_chip.axi_read32(ERROR_STATUS0)
+    cable_fault = bool(err & (1 << INIT_STAGE_CABLE_FAULT))
+    if cable_fault:
```
jgrowdenTT: Is this necessary? If cable fault is high this entire test suite is gonna fail.

drivers/reset/reset_tt_bh.c
```diff
@@ -128,7 +182,7 @@ static DEVICE_API(reset, tt_bh_reset_api) = {
 	(!!DT_INST_PROP_OR(_n, nresets, 0) + !!DT_INST_PROP_OR(_n, reset_mask, 0))
 
 #define TT_BH_RESET_MASK_DEFINE(_n)                                                                \
-	((TT_BH_RESET_MASK(_n) | TT_BH_RESET_MASK_FROM_NRESETS(_n)) || (-1))
+	((TT_BH_RESET_MASK(_n) | TT_BH_RESET_MASK_FROM_NRESETS(_n)) ?: UINT32_MAX)
```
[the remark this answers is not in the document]
jgrowdenTT: yes pls

app/dmc/boards/dmc_common.dtsi
```diff
@@ -6,6 +6,7 @@
 
 #include <zephyr/dt-bindings/gpio/gpio.h>
 #include <zephyr/dt-bindings/i2c/i2c.h>
+#include <tenstorrent/tt_bh_reset_emul.dtsi>
```
jgrowdenTT, line 9: I'm not entirely sure if this works. I know the DTS supports C preprocessor directives.

So I'm thinking:
```
#define TT_RST_COMPAT "tenstorrent,bh-reset-emul"
#include <tenstorrent/tt_bh_reset.dtsi>
```

Where inside you can just `ifndef `in the real thing presuming the source didn't set the compat
```
#ifndef TT_RST_COMPAT
#define TT_RST_COMPAT "tenstorrent,bh-reset"
#endif
```
And then you use `TT_RST_COMPAT` in place of the hardcoded strings in each dtsi file.

Which then means you can have one dtsi file.

lib/tenstorrent/bh_arc/init.h
```diff
@@ -10,87 +10,7 @@
...
-#define RESET_UNIT_TENSIX_RESET_4_REG_ADDR 0x80030030
-#define RESET_UNIT_TENSIX_RESET_5_REG_ADDR 0x80030034
-#define RESET_UNIT_TENSIX_RESET_6_REG_ADDR 0x80030038
-#define RESET_UNIT_TENSIX_RESET_7_REG_ADDR 0x8003003C
-
-#define RESET_UNIT_TENSIX_RISC_RESET_0_REG_ADDR 0x80030040
-#define SCRATCHPAD_SIZE                         CONFIG_TT_BH_ARC_SCRATCHPAD_SIZE
-
-typedef struct {
-	uint32_t system_reset_n: 1;
```
jgrowdenTT, line 31: Its tough to tell at a glance this was all macro-ized succesfully. 

Depending on how thorough our native sim testing is I guess can show how much risk there was to the refactor. 

I also can't immediately tell why we wanted the macros visible to the DTS. Probably just blind.

jgrowdenTT (dismissed):
Nothing major. Cleanups suggested can be done now or later.

app/smc/pytest/e2e_smoke.py
projkovTT: If I understand correctly that getting back INIT_STAGE_CABLE_FAULT indicates SMC failed to init, then yea this is redundant. There's no point in continuing the test with different expected reset bank values

app/dmc/boards/dmc_common.dtsi
projkovTT: I see, that makes sense. That's cuts down on the copied dts stuff. The emul multi-bank reset nodes for tensix have an `n-banks` (num banks) property since they lack the reg property in which that's encoded. But there's probably another better way to bring that information to emul.

drivers/reset/reset_tt_bh_emul.c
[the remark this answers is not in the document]
projkovTT: The elvis operator is cute (and compiles everywhere so far) but Copilot is right here - it's not C standard. I'll update to regular ternary. Good knowledge to file away is that the previous implementation had a bug from using logical || which casts to bool type, ie. 0xf...f was turned into 0x1.

app/dmc/boards/dmc_common.dtsi
projkovTT: Looking at what Cursor built it's not really cleaner. Perhaps it just didn't do a good job, but I don't have enough zephyr/dts knowhow to do a better job of it myself.

jgrowdenTT (approved):

alexapostolu (approved):

alayander (approved):

PR #1400, opened 2026-08-06
Revert multi bootfs prebuilt

alexapostolu:
We decided to go a different path with how MCUBoot and Recovery images
are generated, and will instead build them from CMake.

This also caused a regression in manufacturing in p100: https://github.com/tenstorrent/tt-system-firmware/actions/runs/30979590866/job/92657624649

Tests: Smoke and Stress


boards/tenstorrent/tt_blackhole/bootfs/preflash-bootfs.yaml
alayander, line 1: The changes in this file are probably what caused p100a manufacturing test failures. I presume the motherboard for that runner is one of the ones which is more strict on `num serdes` the recovery image is built with

projkovTT:
Will this cause CI to take longer? Also with these changes is `west blobs fetch tt-system-firmware` no longer required for repo getting started? [tt-sys-fw getting started](https://docs.tenstorrent.com/tt-system-firmware/develop/getting_started/index.html)

alexapostolu:
@projkovTT I'm not sure about CI times, but since this is reverting a recent change, I think it's fine.
We would still need `west blobs fetch tt-system-firmware` since we have other blobs in the repo

projkovTT (approved):

alewycky-tenstorrent (approved):

alayander (approved):

jgrowdenTT (approved):

PR #1408, opened 2026-08-10
lib: bh_arc: source msgqueue info register from Devicetree

projkovTT:
## Overview
Move MSG_Q_INFO scratch word address out of the hardcoded Blackhole `status_reg.h` map and into `tenstorrent,scratch-reg` DT node.
Boards can supply different absolute addresses in their overlays.
native_sim and DMC emul overlays keep the BH ABI address so `msgqueue.c` still builds.

Addresses [SYS-4955](https://tenstorrent.atlassian.net/browse/SYS-4955)

## Tests
Smoke on p150a
Mimir smc build + emul test

### DIFF
--- a/dts/vendor/tenstorrent/tt_grendel_smc.dtsi
+++ b/dts/vendor/tenstorrent/tt_grendel_smc.dtsi
@@ -333,5 +333,12 @@
 			channel-pairs = <32>;
 			#mbox-cells = <1>;
 		};
+
+		/* To be overridden by board */
+		msgqueue_info: scratch-reg@8003042c {
+			compatible = "tenstorrent,scratch-reg";
+			reg = <0x0 0x8003042c 0x0 0x4>;
+			status = "disabled";
+		};
 	};
 };
--- a/app/dmc/boards/dmc_common.dtsi
+++ b/app/dmc/boards/dmc_common.dtsi
@@ -58,6 +58,12 @@
 		status = "okay";
 	};
 
+	msgqueue_info: scratch-reg@8003042c {
+		compatible = "tenstorrent,scratch-reg";
+		reg = <0x8003042c 0x4>;
+		status = "okay";
+	};
+
 	chip0: chip0 {
 		compatible = "tenstorrent,bh-chip";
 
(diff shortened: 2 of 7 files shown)


jgrowdenTT:
I think probably better is if the binding is just a generic SCRATCH binding. No different except for the name.

then we have a node for msgqueue which is an instance of said SCRATCH binding.

That means we can re-use this compatible for (e.g.) telemetry posting, vuart posting, etc.

thoughts?

projkovTT:
> I think probably better is if the binding is just a generic SCRATCH binding. No different except for the name.
> 
> then we have a node for msgqueue which is an instance of said SCRATCH binding.
> 
> That means we can re-use this compatible for (e.g.) telemetry posting, vuart posting, etc.
> 
> thoughts?

Agreed, done.

dts/vendor/tenstorrent/tt_grendel_smc.dtsi
```diff
@@ -333,5 +333,12 @@
 			channel-pairs = <32>;
 			#mbox-cells = <1>;
 		};
+
+		/* To be overridden by board */
+		msgqueue_info: scratch-reg@8003042c {
+			compatible = "tenstorrent,scratch-reg";
+			reg = <0x0 0x8003042c 0x0 0x4>;
+			status = "disabled";
+		};
```
alexapostolu, line 342: What's the point of this if it's going to get overriden?

alexapostolu (approved):

dts/vendor/tenstorrent/tt_grendel_smc.dtsi
projkovTT: I think overridden is kind of misleading. This is to indicate that I'm not sure what the actual address we use on Grendel smcs are. If we end up using the same one for both then can just enable this node.
Whatever happens this will get updated when I add message queue for Grendel so its fine for now.

jgrowdenTT: SCRATCH base is completely different so grendel is different. 

I'm pretty sure I can use this plumbing as is to implement grendel side. But I'll keep you all posted.

jgrowdenTT (approved):

PR #1410, opened 2026-08-10
scripts: ci: stability improvements

jgrowdenTT:
### Overview

Try increasing the flash timeout to see if we can address the spotty CI issues this way.

Change the msg test to not bother verifying the post code. The post code can be updated by internal mechanisms by the time the host checks the scratch register, so the assertion is not guaranteed by our SW behaviour.

Don't run the ci-stats job except on pushes to main (like when we publish docs) since the contents of the ci-stats are always the same per PR since they only take into account stats on main. (Future effort could be made to remove it if we think the newer CI dashboard replaces this.)

### tests
CI here


projkovTT (dismissed):

jgrowdenTT:
I need docs to build. looking into it. Done.

alexapostolu (approved):

PR #1412, opened 2026-08-11
lib: tenstorrent: Move messagequeue out of bh_arc and into new fw_common lib

projkovTT:
## Overview
Create `fw_common` library for firmware common to Blackhole, Grendel SMC
Move `reg.h`, `reg_stub.c`, `post_code.c` into `fw_common`
Move `msgqueue.c` into `fw_common`

Logic behind moving `reg.h` is because `msgqueue.c` is a consumer of `reg.h`. Generalist `fw_common` shouldn't call funcs defined in specialist `bh_arc`.

## Future Work
Update `msgqueue.c` to use Mbox IRQ callback instead of MSI.
Create Grendel test app which inits msgqueue, and test on m/k.

## Tests
p150a smoke

hello world build / emul for Mimir:
- Change msgqueue nodes in `tt_grendel_smc.dtsi` to status okay
- Build with `-DCONFIG_TT_FW_COMMON=y`
- `msgqueue` symbols appear in `zephyr.elf`
- Test with zrun (pass! Doesn't really matter though, we don't even init msgqueue)

## Note about ReadReg/WriteReg for Grendel
These funcs throw: `warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]` when building for 64 bit
- I vote for changing param `uint32_t addr` to `uintptr_t addr`
- Not too important for this PR since `fw_common` isn't yet included in Grendel builds


jgrowdenTT (changes requested):
rebase pls

dts/vendor/tenstorrent/tt_grendel_smc.dtsi
```diff
@@ -340,5 +340,17 @@
 			reg = <0x0 0x8003042c 0x0 0x4>;
 			status = "disabled";
 		};
+
+		arc_misc_cntl: arc-misc-cntl@80030100 {
```
jgrowdenTT: Remove.

lib/tenstorrent/fw_common/post_code.c
```diff
@@ -5,14 +5,19 @@
  */
 
 #include <stdint.h>
+
 #include <tenstorrent/post_code.h>
+
 #include "reg.h"
+
+#if defined(CONFIG_TT_POST_CODE)
 #include "status_reg.h"
```
projkovTT, line 15: Thinking to the future, perhaps we can use KConfigs to include `status_reg_arc.h`, `status_reg_mimir.h` and so on for different boards? Open to suggestions. This does kind of violate my "rule" of `fw_common` not consuming board specific files.

jgrowdenTT: Impl can go to SOC specific. status reg defs can go there also. 

out of scope of this.

jgrowdenTT (approved):

alexapostolu (approved):

PR #1416, opened 2026-08-12
github: ci: Galaxy flash fix 

jgrowdenTT:
Target the galaxy revC runner, which should have the new label.

Tests will not work on the old one, as tt-flash will correctly determine the old galaxy FW bundle is not acceptable for the revC, and therefore tests will run on whatever was flashed before.

Remove the old board for now since we don't have a runner for it.

Can't run the upgrade-from-version test on the revc galaxy because the FW for that board didn't exist at the time.

### Tests:

stress here passed everything (except the upgrade test)

### DIFF
--- a/.github/ci_boards.json
+++ b/.github/ci_boards.json
@@ -50,8 +50,9 @@
       "metal-options": "--cap-add SYS_MODULE --device /dev/tenstorrent --user=root",
       "kmd_build": "2.10.0"
     },
+
     {
-      "board": "bh-galaxy",
+      "board": "bh-galaxy-revc",
       "metal-image": "ghcr.io/tenstorrent/tt-metal/upstream-tests-bh-glx:v0.74.0",
       "metal-target": "blackhole_glx",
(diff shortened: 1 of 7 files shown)


.github/ci_boards.json
```diff
@@ -50,8 +50,9 @@
       "metal-options": "--cap-add SYS_MODULE --device /dev/tenstorrent --user=root",
       "kmd_build": "2.10.0"
     },
+
     {
-      "board": "bh-galaxy",
+      "board": "bh-galaxy-revc",
```
projkovTT, line 55: Intentional to change to `bh-galaxy` to `bh-galaxy-revc` instead of adding? We don't have a runner for `bh-galaxy` on CI anymore?

projkovTT: Oh yes, you said in the description.

projkovTT (approved):

alayander (approved):

PR #1419, opened 2026-08-12
MK DMC: a shell on the USB-C port, alongside the UART one

andrewvuTT:
The MK's USB-C port goes straight to the STM32 (PA11 = D-, PA12 = D+), but nothing claimed it, so the only way into the DMC shell was uart4 out through an FT4232H. This exposes the port as CDC ACM and runs a shell on it, so the board needs one cable and no adapter to be driven.

The board already had the USB node enabled with the right pinctrl and clk_hsi48 on -- byte-identical to upstream's nucleo_u385rg_q, which is the same SoC -- so only a CDC ACM function node and the app's Kconfig were missing.

Both shells at once, rather than moving the shell
-------------------------------------------------
Upstream's cdc-acm-console snippet repoints chosen zephyr,shell-uart at the CDC ACM, which *moves* the shell to USB and takes it off uart4. That would have broken every host tool that drives this DMC over the FTDI, so it is not what this does. Zephyr's serial shell backend is a single instance bound to zephyr,shell-uart (subsys/shell/backends/shell_uart.c), so src/usb_shell.c declares a second transport and shell against the same public SHELL_UART_DEFINE API. chosen zephyr,shell-uart is untouched and uart4 keeps its shell.

The second instance uses CONFIG_SHELL_PROMPT_UART deliberately, so both transports present the same prompt: host tools find the end of a reply by matching it, and they should not have to care which wire they came in on. It is registered with log_backend = false, because uart4's shell is already the log backend and two would duplicate every message -- logs stay on the FTDI, which also keeps them readable while USB itself is being brought up.

Log levels are not incidental here
----------------------------------
dm_test_app runs the whole log subsystem at DBG. Left there, UDC_DRIVER alone sprays a line per IN transfer ("DataIn ep 0x%02x" in udc_stm32.c) into the uart4 console -- which is also the shell backend, so driving the USB shell corrupted replies on the FTDI one. A partial line ("ataIn ep 0x80") does not match the host's log-line filter and lands inside a command's output. USBD, USBD_CDC_ACM and UDC_DRIVER are therefore pinned to WRN; real faults still show up, and the image is 4.7 KB smaller for losing the DBG strings.

Hardware notes for anyone repeating this
----------------------------------------
VDDUSB is a separate supply on this part (3.0-3.6 V, "external independent power supply for USB transceivers" -- DS14861 3.8.1) and the USB FS block sits in that domain. Zephyr sets PWR_SVMCR.USV via LL_PWR_EnableVDDUSB(), but that only declares an external supply valid; it does not create one. With that rail down the PHY cannot drive the D+ pull-up and the host never sees a device.

BOOT0/PH3 must be LOW. High boots the system bootloader, which enumerates as an ST DFU device rather than this CDC ACM shell.

There is no CRS node in Zephyr's U3 DTS, so HSI48 runs on its factory trim rather than being trimmed against USB SOF. Same as the nucleo board, and it enumerates reliably here, but it is the first thing to suspect if it ever goes flaky across temperature.

Tested on MK board 3 (STM32U375RGT6): enumerates as a full-speed CDC ACM, 2fe3:0004 "MK DMC shell", bound to /dev/ttyACM1. Both shells were driven concurrently, each doing real I2C through its own session; a full functional sweep of every reachable component returned identical results over USB and UART, and round-trip latency is 5.4 ms against 6.0 ms over the FTDI.

CONFIG_CDC_ACM_SERIAL_VID/_PID are still Zephyr's defaults (2fe3:0004), which is why lsusb labels the board "NordicSemiconductor". That wants a Tenstorrent VID/PID before it goes anywhere near production.

### DIFF
--- a/app/dm_test_app/prj.conf
+++ b/app/dm_test_app/prj.conf
@@ -31,6 +31,34 @@ CONFIG_LOG_BUFFER_SIZE=32768
 # create build/zephyr/zephyr.lst
 CONFIG_OUTPUT_DISASSEMBLY=y
 
+# USB device: the board's USB-C port goes straight to the STM32 (PA11/PA12), so
+# expose it as a CDC ACM serial port and run a second shell on it. The UART shell
+# on uart4 (FTDI) is untouched -- both work at once, see usb_shell.c.
+#
+# INITIALIZE_AT_BOOT builds the whole USBD context (descriptors, configuration,
+# usbd_init + usbd_enable) around the first zephyr,cdc-acm-uart under
+# zephyr_udc0, so the app needs no USB setup code of its own.
+CONFIG_USB_DEVICE_STACK_NEXT=y
+CONFIG_USBD_CDC_ACM_CLASS=y
+CONFIG_CDC_ACM_SERIAL_INITIALIZE_AT_BOOT=y
+CONFIG_CDC_ACM_SERIAL_PRODUCT_STRING="MK DMC shell"
+CONFIG_CDC_ACM_SERIAL_MANUFACTURER_STRING="Tenstorrent"
+CONFIG_UART_LINE_CTRL=y
+
+# This app runs the whole log subsystem at DBG (below). Left there, USB
+# enumeration alone emits hundreds of lines per plug-in over the 115200 console,
+# which is slow enough to disturb the enumeration it is reporting on. Pin the USB
+# modules to warnings so real faults still show up.
+#
+# UDC_DRIVER matters as much as the other two: udc_stm32.c logs "DataIn ep 0x%02x"
+# at DBG on *every* IN transfer, and the console is uart4 -- which is also the
+# shell backend. Left at DBG, driving the USB shell sprays log lines into the
+# FTDI shell's replies, where a partial line ("ataIn ep 0x80") does not match the
+# host's log-line filter and lands inside a command's output instead.
+CONFIG_USBD_LOG_LEVEL_WRN=y
+CONFIG_USBD_CDC_ACM_LOG_LEVEL_WRN=y
+CONFIG_UDC_DRIVER_LOG_LEVEL_WRN=y
+
 # Shell configuration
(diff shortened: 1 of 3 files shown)


app/dm_test_app/prj.conf
```diff
@@ -31,6 +31,34 @@ CONFIG_LOG_BUFFER_SIZE=32768
 # create build/zephyr/zephyr.lst
 CONFIG_OUTPUT_DISASSEMBLY=y
 
+# USB device: the board's USB-C port goes straight to the STM32 (PA11/PA12), so
```
jgrowdenTT, line 34: I guess this is fine since copilot didn't add anything.

I do wonder if it needs to go in a board specific conf file to keep the nucleo build alive. (or, yknow, we add the usb DTS goo required so it works there also) But we can deal with that after merge.

app/dm_test_app/src/usb_shell.c
```diff
@@ -0,0 +1,72 @@
...
+#include <zephyr/kernel.h>
+#include <zephyr/shell/shell.h>
+#include <zephyr/shell/shell_uart.h>
+
+#define USB_SHELL_NODE DT_NODELABEL(cdc_acm_uart0)
+
+#if DT_NODE_HAS_STATUS_OKAY(USB_SHELL_NODE)
+
+SHELL_UART_DEFINE(shell_transport_usb);
+SHELL_DEFINE(shell_usb, CONFIG_SHELL_PROMPT_UART, &shell_transport_usb,
```
jgrowdenTT, line 40: Appreciate if you can run clang-format and squash in the changes. Thanks Andrew!

projkovTT (approved):

alayander (approved):

alayander:
@andrewvuTT - the compliance checks probably need to be addressed before we can merge, could you amend the commit message line lengths and add the missing signed-off-by line 🙏

PR #1421, opened 2026-08-13
app: dmc: MCTP/PLDM initial commit

jgrowdenTT:
### Overview:

Add initial MCTP endpoint support. Enable it on just the nucleo board.

Add in a test for the DMC which operates in loopback configuration. This is only to be enabled on the nucleo when used to self test. 

Same DTS for target side is used for the DMC app for nucleo (native sim lacks I3C emulation so we can't hook the test or the DM App nucleo builds into it, though there is an upstream PR for i3c emulation we could steal for our fork.)

### tests:
Loopback nucleo tests locally
CI here should run all PLDM native sim tests

resolves: SYS-4685


app/dmc/boards/nucleo_u385rg_q.overlay
```diff
@@ -8,3 +8,72 @@
 #define DMC_FLASH_DEV <&{/soc/flash-controller@40022000}>
 
 #include "dmc_common.dtsi"
+#include "../../../tests/drivers/pmci/pldm/include/tt_pldm_sensor_ids.h"
+
```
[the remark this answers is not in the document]
jgrowdenTT: This is intentional to make the "test" more of an integration test rather than just a unit test of the PLDM block. But yes, this test can never be upstreamed.

app/dmc/boards/nucleo_u385rg_q.overlay
```diff
@@ -8,3 +8,72 @@
 #define DMC_FLASH_DEV <&{/soc/flash-controller@40022000}>
 
 #include "dmc_common.dtsi"
+#include "../../../tests/drivers/pmci/pldm/include/tt_pldm_sensor_ids.h"
```
alexapostolu: It seems weird to include a file from the tests folder, would it make more sense if `tt_pldm_sensor_ids.h` was defined somewhere else?

projkovTT (dismissed):
Looks good. I'd like other approvals though since it's a big PR. Feel good about it though given the depth of the unit tests. Do you have coverage?

app/dmc/boards/nucleo_u385rg_q.overlay
```diff
@@ -8,3 +8,72 @@
...
+			sensor-id = <BH_PLDM_SENSOR_ID_VCORE>;
+		};
+
+		vreg_temp: vreg_temp {
+			compatible = "tenstorrent,pldm-numeric-sensor";
+			sensor = <&smc_sensor>;
+			channel-type = <12>;
+			channel-index = <0>;
+			sensor-id = <BH_PLDM_SENSOR_ID_VREG_TEMP>;
+			unit-modifier = <(-1)>;
```
alayander: There are duplicate pairs of channel-type and channel-index, are these placeholder values?

jgrowdenTT:
i don't have coverage on STM, no. I think it is possible to generate cov if we throw enough memory at the problem on device but I haven't looked into how that works.

app/dmc/boards/nucleo_u385rg_q.overlay
jgrowdenTT: The nucleo board running the DMC app is a test configuration for the test setup; It's got BH_ARC_EMUL on and etc. 

I wouldn't mind it if we used a real sensor IDs file but we'd have to define it. I didn't want to be in a position where this work was blocking the release (like carving out the SENSORS for ranges for K/M/Q/BH/DMC....) The IDs I use in the test are at least temporary.

app/dmc/boards/nucleo_u385rg_q.overlay
jgrowdenTT: Thanks, let me remove them.  The only thing properly emulated on ARC side is temp IIRC. (And TDP is needed.)

tests/drivers/pmci/pldm/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# SPDX-License-Identifier: Apache-2.0
+
+cmake_minimum_required(VERSION 3.20.0)
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(dmc_pldm_mctp_transport_test)
+
+target_include_directories(app PRIVATE
+    ${CMAKE_CURRENT_LIST_DIR}/../../../lib/tenstorrent/bh_arc
+    ${CMAKE_CURRENT_LIST_DIR}/include
+)
```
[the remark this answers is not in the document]
jgrowdenTT: Right, this was needed when I didn't have a proper way of stubbing telem. Now I don't need this, which is why it's compiling fine (CMake doesn't audit if an inc dir exists, it just adds it to the -I even if it's useless.)

drivers/pmci/pldm/pldm_mctp_responder.c
```diff
@@ -0,0 +1,275 @@
...
+	const struct device *dev = cb_data;
+	struct tt_pldm_mctp_responder_data *data = dev->data;
+	const uint8_t *rx_buf = msg;
+	struct tt_pldm_rx_work_item item;
+
+	ARG_UNUSED(tag_owner);
+
+	if (len < 1U || len > sizeof(item.buf)) {
+		return;
+	}
```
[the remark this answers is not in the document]
jgrowdenTT: Fine for now.

projkovTT (approved):

alayander (approved):

PR #1422, opened 2026-08-13
boards: tenstorrent: mmk board

jgrowdenTT:
### Overview
Add in the keraunos SOC and the MMK board definitions. 

### Tests
This will **_not_** compile as is. The sival drop is needed in order to compile, and the drop is not release ready; tenstorrent developers can source it from internal repo in the interim.

With the sival drop, I have working PCIE initialization, shell, and msgqueue.

resolves: SYS-5011


projkovTT (dismissed):
typo that Copilot suggested, otherwise looks good.
I can add blackhole `post_code.c` into soc once this and #1412 are in as a follow up.

projkovTT (approved):

alexapostolu (approved):

PR #1423, opened 2026-08-13
lib: tenstorrent: bh_arc: process error logs for gddr therm trip

alayander:
On logs with error-level, flush buffer to host immediately when processed.
Also force the log thread to to process queued logs on a gddr therm trip so that log can be captured by host before therm trip is triggered.

**Tests**:
- Local spot check on p150a system (dmesg captures log of therm trip before ASIC is gets tripped)

**Resolves**: SYS-4905

### DIFF
--- a/lib/tenstorrent/bh_arc/cat.c
+++ b/lib/tenstorrent/bh_arc/cat.c
@@ -23,6 +23,7 @@
 #include <zephyr/drivers/sensor.h>
 #include <zephyr/drivers/sensor/tenstorrent/pvt_tt_bh.h>
 #include <zephyr/logging/log.h>
+#include <zephyr/logging/log_ctrl.h>
 
 LOG_MODULE_REGISTER(cat);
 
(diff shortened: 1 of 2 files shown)


lib/tenstorrent/bh_arc/cat.c
```diff
@@ -290,6 +291,10 @@ static void gddr_therm_trip_work_handler(struct k_work *work)
 	int max_temp = MonitorGddrThermTrip(k_uptime_get(), GetTelemetryTag(TAG_MAX_GDDR_TEMP));
 
 	if (max_temp && GetActiveFeatures().gddr_therm_trip) {
+		/* Force log thread to process so the over-temp error above
+		 * reaches the host (KMD) before the therm trip resets the SMC.
+		 */
+		log_flush();
```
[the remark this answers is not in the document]
alayander: This is intentional to ensure the log goes through before the trip triggers

projkovTT, line 297: If the SMC is getting reset anyways, can we use log_panic()?

lib/tenstorrent/bh_arc/tt_pcie_log.c
```diff
@@ -384,6 +384,12 @@ static void backend_process(const struct log_backend *const backend, union log_m
 	uint16_t total_size_u16 = (uint16_t)total_size;
 
 	memcpy(framed_log_buffer + entry_start_pos, &total_size_u16, sizeof(total_size_u16));
+
+	/* Push error-level logs to host immediately */
+	if (entry_header.log_level == LOG_LEVEL_ERR) {
+		flush_buffer_to_host();
+	}
```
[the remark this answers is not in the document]
alayander: I believe on KMD side, it will consume logs every MSI and return ownership to FW

lib/tenstorrent/bh_arc/cat.c
alayander: Yeah I think it would be fine to call `log_panic()` in this situation. Functionally, either function will reach our goal of pushing the error log to be processed before therm trip. If preferred I can switch the call to `log_panic()`

projkovTT: My though process was that if there are other logs afterwards we want to get through the panic makes sure of that. But really this is fine

projkovTT (approved):

alexapostolu (approved):

PR #1424, opened 2026-08-14
lib: tenstorrent: bh_arc: Use sys init instead of init.c

projkovTT:
Adresses [SYS-5030](https://tenstorrent.atlassian.net/browse/SYS-5030)

## Overview
Moves the following `bh_arc` inits out of `init.c` and uses Zephyr `SYS_INIT` instead (order preserved):
- DVFS
- msgqueue
- telemetry
- fan_cntrl
- Telemetry timer start
- DVFS timer start
- GDDR therm trip monitor

After all this, `init.c::bh_arc_init_end` is still called to perform:
- Post `POST_CODE_ZEPHYR_INIT_DONE`
- printk version
- `Dm2CmReadyRequest`

Only functional difference is that post code and printk used to happen before all the rest. I actually think it makes much more sense to do them after.

## Tests
Last time we did something like this we ran into dirty reset test failures. With recent fixes to host side of dirty reset that should be fixed.

Smoke on p150a
Stress dirty reset on p150a (0/1000 failures)
stress "super dirty reset" (remove the 0.5s sleep between resets) 0/1000 failures

TODO: Stress this on gh-03 since it fails the dirty reset in CI.

### DIFF
--- a/include/tenstorrent/msgqueue.h
+++ b/include/tenstorrent/msgqueue.h
@@ -1146,7 +1146,7 @@ int msgqueue_request_push(uint32_t msgqueue_id, const union request *request);
 int msgqueue_request_pop(uint32_t msgqueue_id, union request *request);
 int msgqueue_response_push(uint32_t msgqueue_id, const struct response *response);
 int msgqueue_response_pop(uint32_t msgqueue_id, struct response *response);
-void init_msgqueue(void);
+int init_msgqueue(void);
 
 #ifdef __cplusplus
 }
--- a/include/tenstorrent/sys_init_defines.h
+++ b/include/tenstorrent/sys_init_defines.h
@@ -34,7 +34,14 @@
 #define gddr_training_PRIO                    111
 #define CATInit_PRIO                          112
 #define publish_runtime_telemetry_PRIO        113
-#define bh_arc_init_end_PRIO                  114
+#define InitDVFS_PRIO                         114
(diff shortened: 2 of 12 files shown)


lib/tenstorrent/bh_arc/aiclk_ppm.c
```diff
@@ -281,10 +283,13 @@ uint8_t ForceAiclk(uint32_t freq)
 		return 1;
 	}
 
+#ifdef CONFIG_TT_BH_ARC_DVFS
```
projkovTT: The #ifdef spam is unfortunate. I have another branch which just adds stubs to the header when the config is off.
Truthfully not sure what I prefer - the ifdefs make it pretty clear when we're calling funcs that may have missing source but are gross in the callee. Stubs could make misconfigured builds fail non-obviously at runtime.
Open to suggestions.

projkovTT: Just removed these kconfigs always build the source (like we were already doing). Only difference with main is the init bit. If we want to select these modules with Kconfig that's future work.

alexapostolu (dismissed):

alayander (dismissed):

jgrowdenTT (dismissed):

projkovTT:
rebased

jgrowdenTT (dismissed):

alexapostolu (approved):

PR #1430, opened 2026-08-17
lib: bh_arc: add characterization msg to set telemetry interval

kaiqihuTT:
Add TT_SUB_MSG_SET_TELEMETRY_UPDATE_INTERVAL (0x5) under TT_SMC_MSG_CHARACTERISATION so the host can change the periodic telemetry update interval at runtime. The new interval restarts the telemetry timer immediately and is published in TAG_UPDATE_TELEM_SPEED. A payload of 0 restores the default.

The accepted range is hardcoded in telemetry.h as
TELEM_UPDATE_INTERVAL_MIN_MS (10) to TELEM_UPDATE_INTERVAL_MAX_MS (1000), with a default of 100 ms. Values outside that range are rejected with a warning naming the bounds and leave the running interval unchanged.

Tested with twister on native_sim, including two new cases in tests/lib/tenstorrent/bh_arc/src/msgqueue.c covering valid values, both range boundaries, restore-to-default, and rejection of out-of-range values without clobbering the current interval. Verified on hardware via 'tt msg 0x5c6 <ms>

### DIFF
--- a/lib/tenstorrent/bh_arc/aiclk_ppm.c
+++ b/lib/tenstorrent/bh_arc/aiclk_ppm.c
@@ -542,6 +542,12 @@ static uint8_t characterisation_handler(const union request *request, struct res
 		return CatSetGddrThermTripEnabled(
 			request->characterisation_msg.submsg_data.gddr_therm_trip_enabled.enabled);
 
+#ifdef CONFIG_BH_FWTABLE
+	case TT_SUB_MSG_SET_TELEMETRY_UPDATE_INTERVAL:
+		return TelemetrySetUpdateInterval(
+			request->characterisation_msg.submsg_data.telemetry_interval.interval_ms);
+#endif
+
 	default:
 		LOG_WRN("Unknown characterization submessage ID: 0x%02x",
 			request->characterisation_msg.submsg_ID);
--- a/lib/tenstorrent/bh_arc/telemetry.h
+++ b/lib/tenstorrent/bh_arc/telemetry.h
@@ -496,10 +496,32 @@ typedef union {
 /* Telemetry tags are at offset `tag` in the telemetry buffer */
 #define TELEM_OFFSET(tag) (tag)
 
+/** @brief Default periodic telemetry update interval in milliseconds. */
+#define TELEM_UPDATE_INTERVAL_DEFAULT_MS 100
+
+/** @brief Shortest accepted telemetry update interval in milliseconds.
+ *
+ * The telemetry pass shares the system work queue with the 1 ms DVFS loop, so the interval is
+ * floored well above 1 ms to keep telemetry from crowding out frequency and voltage control.
+ */
+#define TELEM_UPDATE_INTERVAL_MIN_MS 10
+
+/** @brief Longest accepted telemetry update interval in milliseconds. */
+#define TELEM_UPDATE_INTERVAL_MAX_MS 1000
+
 void init_telemetry(uint32_t app_version);
 uint32_t ConvertFloatToTelemetry(float value);
 float ConvertTelemetryToFloat(int32_t value);
 void StartTelemetryTimer(void);
+
+/** @brief Set the periodic telemetry update interval.
+ * @param interval_ms 0 restores @ref TELEM_UPDATE_INTERVAL_DEFAULT_MS, otherwise the interval in
+ *                    milliseconds, which must be within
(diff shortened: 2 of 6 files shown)


lib/tenstorrent/bh_arc/telemetry.h
```diff
@@ -496,10 +496,32 @@ typedef union {
...
 
+/** @brief Default periodic telemetry update interval in milliseconds. */
+#define TELEM_UPDATE_INTERVAL_DEFAULT_MS 100
+
+/** @brief Shortest accepted telemetry update interval in milliseconds.
+ *
+ * The telemetry pass shares the system work queue with the 1 ms DVFS loop, so the interval is
+ * floored well above 1 ms to keep telemetry from crowding out frequency and voltage control.
+ */
+#define TELEM_UPDATE_INTERVAL_MIN_MS 10
```
afongTT, line 507: Let it go down to 1.

lib/tenstorrent/bh_arc/aiclk_ppm.c
```diff
@@ -542,6 +542,10 @@ static uint8_t characterisation_handler(const union request *request, struct res
 		return CatSetGddrThermTripEnabled(
 			request->characterisation_msg.submsg_data.gddr_therm_trip_enabled.enabled);
 
+	case TT_SUB_MSG_SET_TELEMETRY_UPDATE_INTERVAL:
+		return TelemetrySetUpdateInterval(
+			request->characterisation_msg.submsg_data.telemetry_interval.interval_ms);
```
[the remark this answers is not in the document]
projkovTT: I was checking this cause I think its a valid suggestion, but not sure its needed: `aiclk_ppm.c` is only compiled in with `!CONFIG_TT_SMC_RECOVERY` which I'm pretty sure always means `CONFIG_BH_FWTABLE`.
That said, probably fine to add this gate anyways.

lib/tenstorrent/bh_arc/telemetry.h
```diff
@@ -496,10 +496,32 @@ typedef union {
...
+
+/** @brief Longest accepted telemetry update interval in milliseconds. */
+#define TELEM_UPDATE_INTERVAL_MAX_MS 1000
```
alexapostolu, line 510: What's the point of a max interval? Why shouldn't someone set it to 1.5 seconds for example?

kaiqihuTT: Since we haven't tested for the smallest and largest interval that can be set to ensure healthy timing and CPU load, it is good to first set a guaranteed safe bound so that users do not make the interval too small or too large to mess things up

lib/tenstorrent/bh_arc/telemetry.h
kaiqihuTT: I think it is safer to do that after we do some testing with those numbers and look at how much timing error it causes and whether that is acceptable.

lib/tenstorrent/bh_arc/telemetry.h
alexapostolu: I get why the min bound is needed, but is the max bound needed?

lib/tenstorrent/bh_arc/telemetry.h
alexapostolu: Is it possible to add a python "stress" test that sets the interval to the minimum, runs tt-burnin while logging telemetry, and see if anything fails?

kaiqihuTT: that sounds like a good idea. In that case I'll just need to come up with some metrics for this test

lib/tenstorrent/bh_arc/telemetry.h
kaiqihuTT: cause it doesn't make sense to set it too high ;). Defies the purpose of telemetry update

alexapostolu (approved):

lib/tenstorrent/bh_arc/telemetry.h
projkovTT: @kaiqihuTT  did we collect some statistics on if 1 ms is stable?
If you want to just merge the 10ms and then make a follow up to test 1 ms then I think that'd be fine too.

projkovTT (approved):

PR #1435, opened 2026-08-18
lib: tenstorrent: fw_common: Add mbox support for msgqueue

projkovTT:
## Overview
Use mbox IRQs instead of MSI IRQs for message handling.

Add msgqueue_mbox nodes to dts, which inform compilation of mbox IRQ vs MSI IRQ.

## Tests
Emul test app `hello_msgqueue` runs sucesfully, receives message from mbox self write.
Test app: https://github.com/projkovTT/tt-system-firmware/blob/hello_msgqueue/app/hello_msgqueue/src/main.c

### DIFF
--- a/boards/tenstorrent/tt_mmk/tt_mmk_tt_keraunos_smc.dts
+++ b/boards/tenstorrent/tt_mmk/tt_mmk_tt_keraunos_smc.dts
@@ -48,3 +48,12 @@
 		reg = <0x0 0x88fc 0x1fe8fa15>;
 	};
 };
+
+&{/soc} {
+	/* Cold Scratch 3 */
+	msgqueue_info: scratch-reg@c000280c {
+		compatible = "tenstorrent,scratch-reg";
+		reg = <0x0 0xc000280c 0x0 0x4>;
+		status = "okay";
+	};
+};
--- a/boards/tenstorrent/tt_mmk/tt_mmk_tt_keraunos_smc.dts
+++ b/boards/tenstorrent/tt_mmk/tt_mmk_tt_keraunos_smc.dts
@@ -27,6 +27,13 @@
 		zephyr,console = &vuart0;
 		zephyr,shell-uart = &vuart0;
 	};
+
+	msgqueue_mbox: msgqueue_mbox {
+		compatible = "tenstorrent,msgqueue-mbox";
+		mboxes = <&mbox0 1>;
+		mbox-names = "doorbell";
+		status = "okay";
+	};
 };
 
 &i3c1 {
--- a/dts/bindings/misc/tenstorrent,msgqueue-mbox.yaml
+++ b/dts/bindings/misc/tenstorrent,msgqueue-mbox.yaml
@@ -1,0 +1,21 @@
+# Copyright (c) 2026 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+  Tenstorrent message queue mbox doorbell channel.
+
+  Links a Grendel mbox controller channel to the portable SMC message
+  queue as its inbound IRQ/doorbell. Presence of a node with this
+  compatible (status okay) selects the mbox IRQ backend in msgqueue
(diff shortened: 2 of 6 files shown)


boards/tenstorrent/tt_mmk/tt_mmk_tt_keraunos_smc.dts
```diff
@@ -41,3 +41,10 @@
 		reg = <0x0 0x88fc 0x1fe8fa15>;
 	};
 };
+
+/* Cold Scratch 2 */
+msgqueue_info: scratch-reg@0x8002808 {
```
jgrowdenTT: On top of the  compliance issue, the 0x8000 base address for Grendel will never be correct. 

The one you had in the board overlay is fine for now.

But this should just be changed in the tt_Grendel_smc.dtsi so we don't have to respecify. That way KMD doesn't have to pick a different scratch register per K/M/Q. They all have the same definitions.

lib/tenstorrent/fw_common/Kconfig
```diff
@@ -36,4 +36,13 @@ config TT_MSGQUEUE_NUM_MSG_CODES
 	help
 	  Size of the message handler table (message code range).
 
+config TT_MSGQUEUE_MBOX_IRQ
```
jgrowdenTT: I think maybe better is to look at the DT for existance of a node `msgqueue_mbox` (which would need a compatible, `tenstorrent,mbox_channel` or something like that.)

That node in the DTS will link mbox0 (phandle) channel X (int) to msgqueue. Then in source we automatically compile in mbox support into msgqueue once we see that node. 

My thought here is that all our MBOX allocations, per board/app, can eventually be kept in one `dtsi` file per board/app and we don't have to update this KCONFIG with every new board. We can also probably write compile-time verification that are allocations are sound and not stomping each other.

boards/tenstorrent/tt_mmk/tt_mmk_tt_keraunos_smc.dts
jgrowdenTT: I'm kinda flip flopping on tt_grendel_smc.dtsi placement FYI. 

Some apps might not have msgqueue and thus this belongs app level. 

Some apps (X-BL1?) Might not have msgqueue.

So arguably the placement goes app level.

projkovTT: > On top of the compliance issue, the 0x8000 base address for Grendel will never be correct.

Oops, I think I was looking at the Keraunos SOC map (which has SMC starting at 0x800_0000). 

We could use cold scratch 3 at 0x280C for both I think.

> I'm kinda flip flopping on tt_grendel_smc.dtsi placement FYI.

I think it's fine to have the node in the dts even if the msgqueue isn't compiled? I appreciate getting the reg address from dts.

projkovTT:
Need https://github.com/tenstorrent/tt-system-firmware/pull/1437 in first

dts/bindings/misc/tenstorrent,msgqueue-mbox.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+compatible: "tenstorrent,msgqueue-mbox"
+
+include: base.yaml
+
+properties:
+  mboxes:
+    required: true
+
+  mbox-names:
```
jgrowdenTT, line 20: I think we could probably drop the name. The DT nodelabel or whatever is enough, right

jgrowdenTT (approved):

alayander (approved):

dts/bindings/misc/tenstorrent,msgqueue-mbox.yaml
projkovTT: I like the name, lets you use Zephyr's helpers `DT_MBOX_CTLR/CHANNEL_BY_NAME` to get the mbox phandle. I think it follows the convention that Zephyr uses

PR #1442, opened 2026-08-21
drivers: misc: mmk Bun2 loader

jgrowdenTT:
### Overview

Impement some common driver code for bundle management.

Implement specific functionality of MMK's K-SMC-BL0p5  to perorm the BUN2 parsing and execution of K-SMC-BL1

Implement the BL0p5 overlay/configuration files so we can build BL0p5.

### Tests
```west build -b tt_mmk/tt_keraunos/smc ./app/smc -- -DFILE_SUFFIX=bl0p5```
(With some additions to main to hack out the WDT update.)

I have a host side application build on the prelim tt-kmd ioctls which drive the SCRATCH state, where this host acts as both SEP and Host.  Confirmed using this applciation boots K-SMC-BL1 if the BUNDLE is placed in SRAM by host.

### DIFF
--- a/drivers/misc/tt_bundle_loader/tt_mmk_bun2_loader.c
+++ b/drivers/misc/tt_bundle_loader/tt_mmk_bun2_loader.c
@@ -1,0 +1,119 @@
+/*
+ * Copyright (c) 2026 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#define DT_DRV_COMPAT tenstorrent_mmk_bun2_loader
+
+#include <errno.h>
+#include <string.h>
+
+#include <zephyr/device.h>
+#include <zephyr/devicetree.h>
+#include <zephyr/drivers/misc/tt_bundle_loader.h>
+#include <zephyr/init.h>
+#include <zephyr/sys/sys_io.h>
+#include <zephyr/kernel.h>
+#include <zephyr/logging/log.h>
+#include <zephyr/sys/util.h>
+
+#include "tt_bundle_loader_priv.h"
+
+LOG_MODULE_REGISTER(tt_mmk_bun2_loader, CONFIG_TT_MMK_BUN2_LOADER_LOG_LEVEL);
+
+/* Host <-> SMC boot handshake states, written/polled via the host-boot-state scratch register */
+#define HOST_BOOT_STATE_WAIT_FOR_BUNDLE (1U)
+#define HOST_BOOT_STATE_BUNDLE_STAGED   (2U)
+
+/* Bundle validation handshake bits, in the bundle-validation scratch register */
+#define BUNDLE_READY_FOR_VALIDATION_BIT BIT(0)
+#define BUNDLE_VALIDATED_BIT            BIT(1)
+
+/* SMC_CPU_SMC_CPU_CTRL_RESET_VECTOR_0 local address */
+#define SMC_RESET_VECTOR0_ADDR 0xC0010000UL
+
+struct bun2_loader_config {
+
+	uintptr_t host_boot_state_addr;
+	uintptr_t bundle_validation_addr;
+	uintptr_t staging_area;
+	size_t staging_area_size;
+};
+
+static int bun2_loader_init(const struct device *dev)
+{
+	const struct bun2_loader_config *cfg = dev->config;
+
+	/* A) Tell the host we are ready to receive a bundle */
+	sys_write32(HOST_BOOT_STATE_WAIT_FOR_BUNDLE, cfg->host_boot_state_addr);
+
+	/* B) Wait for the host to stage the bundle in the staging area */
+	LOG_INF("Waiting for BUN2 staging");
+	while (sys_read32(cfg->host_boot_state_addr) != HOST_BOOT_STATE_BUNDLE_STAGED) {
+		k_busy_wait(100);
+	}
+
+	/* Close host access to the staging area; only the SMC touches it from here on */
+
+	/* D) Request validation of the staged bundle */
+	sys_write32(BUNDLE_READY_FOR_VALIDATION_BIT, cfg->bundle_validation_addr);
+
+	/* E) Wait for validation to complete */
+	LOG_INF("Waiting for BUN2 validation");
+	while (!(sys_read32(cfg->bundle_validation_addr) & BUNDLE_VALIDATED_BIT)) {
+		k_busy_wait(100);
+	}
+
+	const struct fw_bundle_manifest *manifest =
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: drivers/misc/tt_bundle_loader/tt_mmk_bun2_loader.c [hunk 1]
```diff
@@ -0,0 +1,119 @@
...
+static int bun2_loader_init(const struct device *dev)
+{
+	const struct bun2_loader_config *cfg = dev->config;
+
+	/* A) Tell the host we are ready to receive a bundle */
+	sys_write32(HOST_BOOT_STATE_WAIT_FOR_BUNDLE, cfg->host_boot_state_addr);
+
+	/* B) Wait for the host to stage the bundle in the staging area */
+	LOG_INF("Waiting for BUN2 staging");
+	while (sys_read32(cfg->host_boot_state_addr) != HOST_BOOT_STATE_BUNDLE_STAGED) {
```
projkovTT, line 52: Suggest:
```
while (sys_read32(cfg->host_boot_state_addr) & 0xf != HOST_BOOT_STATE_BUNDLE_STAGED) {
```
AND with `HOST_BOOT_STATE` mask [3:0] because `HOST_BUNDLE_SEQ` [7:4] may be non-zero (I think that in this case it _is_ 0 but I'm not 100% sure).
On a related note, do we want to make a bitfield struct for this scratch

projkovTT: Note "BUN1/BUN2/BUN3 remain payload/manifest labels only". Bun sequence numbers shall be defined in  bundle-sequence table which I don't think we have yet defined.

drivers/misc/tt_bundle_loader/tt_mmk_bun2_loader.c
```diff
@@ -0,0 +1,119 @@
...
+	uintptr_t staging_area;
+	size_t staging_area_size;
+};
+
+static int bun2_loader_init(const struct device *dev)
+{
+	const struct bun2_loader_config *cfg = dev->config;
+
+	/* A) Tell the host we are ready to receive a bundle */
+	sys_write32(HOST_BOOT_STATE_WAIT_FOR_BUNDLE, cfg->host_boot_state_addr);
```
projkovTT, line 48: I think that we should add code or comment to explicitly state the value assigned to `HOST_BUNDLE_SEQ`. At this step, K-SMC should fill `HOST_BUNDLE_SEQ` with the bundle sequence number that SEP shall stage. Sec 8.10.2.

dts/bindings/misc/tenstorrent,mmk-bun2-loader.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+description: |
+  Tenstorrent MMK BUN2 (host-staged) FW bundle loader.
+
+  Drives the host handshake used to stage and validate the FW bundle used to boot the SMC:
+    1. Set the host boot state register to WAIT_FOR_BUNDLE.
+    2. Poll the host boot state register for BUNDLE_STAGED.
+    3. Close host access to the staging region.
+    4. Set the bundle validation register's BUNDLE_READY_FOR_VALIDATION bit.
+    5. Poll the bundle validation register for its BUNDLE_VALIDATED bit.
```
projkovTT, line 12: I think steps 1-5 are bundle agnostic with the exception of `HOST_BUNDLE_SEQ` write number which would be in step 1. Could this be a generic bundle loader `k-smc-bun-loader`? I'm not sure how to do step 6 though

drivers/misc/tt_bundle_loader/tt_mmk_bun2_loader.c
```diff
@@ -0,0 +1,119 @@
...
+
+	if (bl1_entry == NULL || bl1_entry->type != FW_BUNDLE_IMG_TYPE_SMC_BL1) {
+		LOG_ERR("No SMC BL1 image found at TOC index 0 in staged bundle");
+		return -ENOENT;
+	}
+
+	/* F) Copy SMC BL1 from the staged bundle to its execute location */
+	memcpy((void *)bl1_entry->load_addr,
+	       (const void *)(cfg->staging_area + manifest->payload_offset + bl1_entry->offset),
+	       bl1_entry->length);
```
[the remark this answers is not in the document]
projkovTT: I think this is a good idea. Our contract for the bundle size and staging area is pretty well defined but this could help prevent surprises

include/zephyr/drivers/misc/tt_bundle_loader.h
```diff
@@ -0,0 +1,460 @@
...
+	 * Load address for this image. Interpreted on a per-image type basis. May not apply to all
+	 * image types.
+	 */
+	uint64_t load_addr;
+
+	/** @brief Entry point for this image.
+	 * Interpreted on a per-image type basis, may be in different address space than load_addr.
+	 * May not apply to all images.
+	 */
+	uint64_t entry_point;
```
projkovTT, line 392: Improve comment for `load_addr` and `entry_point`. It's not clear to me if `entry_point` is used as an absolute address or offset from `load_addr`. From my understanding, it's up to the consumer?

May want to wrap in some of the comment from boot manifest repo specs doc `yaml-config-schema.md `:

> `load_addr` / `entry_point` are per-image and load-bearing for consumers that place and enter images from the TOC (e.g. a boot ROM copying `BLSTAGE1` to its `load_addr` and entering at `load_addr + entry_point`). Leaving them unset emits 0, which such a consumer reads as "load at address 0, enter at offset 0" — set them explicitly for any image that is executed.

jgrowdenTT:
Closing this; I think I need to rename and reorg. But I'll deal with the PR feedback.

include/zephyr/drivers/misc/tt_bundle_loader.h
jgrowdenTT: These comments should be straight from the boot doc. Any feedback you have should be directed there.

dts/bindings/misc/tenstorrent,mmk-bun2-loader.yaml
jgrowdenTT: Maybe? I didn't look that far ahead. If there are bundle generic staging sequences, that's sort of the point of the generic bundle C source helper.

Probably can be addressed later.

drivers/misc/tt_bundle_loader/tt_mmk_bun2_loader.c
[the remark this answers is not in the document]
jgrowdenTT: My thought was SEP will verify. SMC will not.

PR #1444, opened 2026-08-24
app: K-SMC-bl0p5

jgrowdenTT:
### Overview
Create the application for the K-SMC-BL 0.5. This application:

A) Boots PCIE (TODO)
B) BUN2 loading handshake via host (KMD-side out of repo/TODO) And then close host access (TODO) 
C) BUN2 Validation through SEP (SEP-side TODO)
D) Load and jump K-SMC-BL1

### Tests:
In ttsim this application works to boot K-SMC-BL1 from host. 
This will not compile in main because we are dependent on the SIVAL drop to build MMK applications.

resolves: SYS-5082


projkovTT:
Looks good to me, I agree with the organization.
Typos/compliance that copilot noted.

app/bl0p5/src/main.c
```diff
@@ -0,0 +1,100 @@
...
+	       (const void *)(STAGING_AREA_ADDR + manifest->payload_offset + bl1_entry->offset),
+	       bl1_entry->length);
+
+	__asm__ volatile("fence\nfence.i" ::: "memory");
+
+	LOG_INF("Jumping to SMC BL1 at load_addr 0x%llx", bl1_entry->load_addr);
+
+	/* G) Jump to SMC BL1. Does not return; BL1 owns the core from here on. */
+	/* mret gives BL1 a clean entry: M-mode, MIE=0, no BL0P5 mtvec leaking through */
+	sys_write32((uint32_t)bl1_entry->entry_point, SMC_RESET_VECTOR0_ADDR);
```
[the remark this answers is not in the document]
jgrowdenTT: I think Validation like this should belong to SEP...

app/bl0p5/src/main.c
```diff
@@ -0,0 +1,100 @@
...
+
+	const struct fw_bundle_manifest *manifest =
+		(const struct fw_bundle_manifest *)STAGING_AREA_ADDR;
+	const struct fw_bundle_toc *toc =
+		(const struct fw_bundle_toc *)(STAGING_AREA_ADDR + manifest->payload_offset);
+
+	/* Bundles always place the K-SMC BL1 image at TOC index 0 */
+	const struct fw_bundle_toc_entry *bl1_entry =
+		(toc->image_count > 0) ? &toc->entries[0] : NULL;
+
```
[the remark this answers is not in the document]
jgrowdenTT: negative bundle should be supported by this.... well, other than 0xC0067000 being to "low"

app/bl0p5/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+	/* Close host access to the staging area; only the SMC touches it from here on */
+
+	/* D) Request validation of the staged bundle */
+	sys_write32(BUNDLE_READY_FOR_VALIDATION_BIT, BUNDLE_VALIDATION_ADDR);
+
+	/* E) Wait for validation to complete */
+	LOG_INF("Waiting for BUN2 validation");
+	while (!(sys_read32(BUNDLE_VALIDATION_ADDR) & BUNDLE_VALIDATED_BIT)) {
+		k_busy_wait(100);
+	}
```
alexapostolu, line 53: Could you add a timeout to this so it doesn't have the probability to hang?

app/bl0p5/prj.conf
```diff
@@ -0,0 +1,24 @@
...
+
+# BL0P5 jumps to BL1 in main() and never returns; the PVT sensor
+# driver and MCUboot flash-confirm logic are unused in this variant.
+CONFIG_SENSOR=n
+CONFIG_FLASH=n
+
+# No float/double formatting needed in this variant's printk/printf usage, but
+# 64-bit load addresses are logged so long long integer support must stay.
+CONFIG_CBPRINTF_FP_SUPPORT=n
+CONFIG_PICOLIBC_IO_LONG_LONG=n
```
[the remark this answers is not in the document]
alexapostolu: This seems valid, why did you resolve it @jgrowdenTT ?

app/bl0p5/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+
+int main(void)
+{
+	bl1_launcher_init();
+
+	/* Unreachable unless BL1 returns, which means the handoff failed and the core's
+	 * state (stack, vector table, PMP, ...) can no longer be trusted.
+	 */
+	k_panic();
+	return 0;
```
alexapostolu, line 103: What if `bl1_launcher_init1` returns `-ENOENT`, then that is a different error case and we shouldn't `k_panic()` right?

app/bl0p5/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+#define STAGING_AREA_ADDR      0xc0067000UL
+
+static int bl1_launcher_init(void)
+{
+	/* A) Tell the host we are ready to receive a bundle */
+	sys_write32(HOST_BOOT_STATE_WAIT_FOR_BUNDLE, HOST_BOOT_STATE_ADDR);
+
+	/* B) Wait for the host to stage the bundle in the staging area */
+	LOG_INF("Waiting for BUN2 staging");
+	while (sys_read32(HOST_BOOT_STATE_ADDR) != HOST_BOOT_STATE_BUNDLE_STAGED) {
```
projkovTT: mask to check just the boot state field
`sys_read32(HOST_BOOT_STATE_ADDR) & 0xF`

app/bl0p5/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+
+	if (bl1_entry == NULL || bl1_entry->type != FW_BUNDLE_IMG_TYPE_SMC_BL1) {
+		LOG_ERR("No SMC BL1 image found at TOC index 0 in staged bundle");
+		return -ENOENT;
+	}
+
+	/* F) Copy SMC BL1 from the staged bundle to its execute location */
+	memcpy((void *)bl1_entry->load_addr,
+	       (const void *)(STAGING_AREA_ADDR + manifest->payload_offset + bl1_entry->offset),
+	       bl1_entry->length);
```
alewycky-tenstorrent: Can source and dest overlap? If so use memmove instead of memcpy.

app/bl0p5/src/main.c
jgrowdenTT: I don't think I can, no. not until we get a better idea of how long SEP should take. For now infinite is just more robust than any other option, I think.

app/bl0p5/src/main.c
jgrowdenTT: What would you like to do instead?

app/bl0p5/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+	sys_write32(HOST_BOOT_STATE_WAIT_FOR_BUNDLE, HOST_BOOT_STATE_ADDR);
+
+	/* B) Wait for the host to stage the bundle in the staging area */
+	LOG_INF("Waiting for BUN2 staging");
+	while ((sys_read32(HOST_BOOT_STATE_ADDR) & 0xF) != HOST_BOOT_STATE_BUNDLE_STAGED) {
+		k_busy_wait(100);
+	}
+
+	/* Close host access to the staging area; only the SMC touches it from here on */
+
```
alewycky-tenstorrent, line 45: My reading of [Grendel Boot Specification 8.8.4. K-SMC-BL0P5 to K-SMC-BL1 Handoff](https://arch-spec.local.tenstorrent.com/grendel-boot#ksmc_bl0p5_bl1_handoff) is that you should set `HOST_BOOT_STATE = BUNDLE_CONSUMED` at this point.

alewycky-tenstorrent:
In the commit title, please say this is Keraunos SMC

app/bl0p5/src/main.c
```diff
@@ -0,0 +1,101 @@
...
+	       (const void *)(STAGING_AREA_ADDR + manifest->payload_offset + bl1_entry->offset),
+	       bl1_entry->length);
+
+	__asm__ volatile("fence\nfence.i" ::: "memory");
+
+	LOG_INF("Jumping to SMC BL1 at load_addr 0x%llx", bl1_entry->load_addr);
+
+	/* G) Jump to SMC BL1. Does not return; BL1 owns the core from here on. */
+	/* mret gives BL1 a clean entry: M-mode, MIE=0, no BL0P5 mtvec leaking through */
+	sys_write32((uint32_t)bl1_entry->entry_point, SMC_RESET_VECTOR0_ADDR);
```
alewycky-tenstorrent: SMC_RESET_VECTOR0 is a 56-bit value, I'd prefer sys_write64.

app/bl0p5/src/main.c
jgrowdenTT: The current doc I'm on has it after jump into BL1, but thinking about it, it's probably better to set it here. 

A) A lot of zephyr/early init hook stuff is probably going to fire before we get a BL1 app hook to set the scratch, so we'll have to shoehorn in something in, and even so that hook will happen "later" than the diagram might suggest
B) The window to the stage is functionally closed at the line you're saying. 

I've made a note to discuss this with Ale.

image

jgrowdenTT: I've added it after validation.

app/bl0p5/src/main.c
```diff
@@ -0,0 +1,104 @@
...
+/* SMC_CPU_SMC_CPU_CTRL_RESET_VECTOR_0 local address */
+#define SMC_RESET_VECTOR0_ADDR 0xC0010000UL
+
+#define HOST_BOOT_STATE_ADDR   ((uintptr_t)DT_REG_ADDR(DT_NODELABEL(host_boot_state)))
+#define BUNDLE_VALIDATION_ADDR ((uintptr_t)DT_REG_ADDR(DT_NODELABEL(bundle_validation)))
+
+static int bl1_launcher_init(void)
+{
+	/* A) Tell the host we are ready to receive a bundle */
+	sys_write32(HOST_BOOT_STATE_WAIT_FOR_BUNDLE, HOST_BOOT_STATE_ADDR);
```
projkovTT, line 36: Write `HOST_BUNDLE_SEQ=0`

Sec 12.6

Image

Initialized to zero so doesn't matter just a nit

projkovTT (approved):

app/bl0p5/src/main.c
projkovTT: Ok actually reading the bootspec closer, I think that `HOST_BUNDLE_SEQ` shall be set to 0 _only_ when `HOST_BOOST_STATE` is set to `RESET_OR_UNINITIALIZED`, so the current implementation is correct

alewycky-tenstorrent (approved):

PR #1446, opened 2026-08-25
Version finalization v19.14

kaiqihuTT:


alayander (approved):

projkovTT:
@kaiqihuTT You add the 19.14 migration guide?

alexapostolu:
Honestly should we stop publishing that migration guide? We haven't updated it since 19.0.0 and maybe any migration requirements would be better communicated in the release notes so it's in one place for everyone

projkovTT:
> Honestly should we stop publishing that migration guide? We haven't updated it since 19.0.0 and maybe any migration requirements would be better communicated in the release notes so it's in one place for everyone

I'd be good with that

PR #1447, opened 2026-08-25
drivers: mbox: allow for remote instances

jgrowdenTT:
### Overview
Allow mbox_tt_grendel to register for remote instances e.g. on the mimir board, map via SPA the K mbox.

### Tests:
Using this for bl1 remote boot handshake.

### DIFF
--- a/dts/bindings/mbox/tenstorrent,grendel-mbox.yaml
+++ b/dts/bindings/mbox/tenstorrent,grendel-mbox.yaml
@@ -12,7 +12,10 @@ properties:
     required: true
 
   interrupts:
-    required: true
+    description: |
+      Required for instances that receive on inbound channels. Instances that
+      only ever send (e.g. a remote view of another die's mailbox registers)
+      may omit interrupts entirely.
 
   channel-pairs:
     type: int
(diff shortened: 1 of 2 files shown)


dts/bindings/mbox/tenstorrent,grendel-mbox.yaml
```diff
@@ -12,7 +12,10 @@ properties:
     required: true
 
   interrupts:
-    required: true
+    description: |
+      Required for instances that receive on inbound channels. Instances that
+      only ever send (e.g. a remote view of another die's mailbox registers)
+      may omit interrupts entirely.
```
projkovTT, line 18: Could using interrupts = NULL as a proxy for "this is a remote instance" break down in the future? Ie. might we later have local mailboxes wihtout interrupts. If that's likely I think I'd prefer adding a `remote` field.
If it's not likely then leave as is I think.

jgrowdenTT: ...But then I need to fill in the required interrupts field with something not-null even though it's nonsense since there is none...

jgrowdenTT: Or sorry, are you suggesting Allow NULL interrupts but also have a `remote-mbox` flag? I dunno what else that flag would do though I think we can revisit in the future if need be.

projkovTT: Yea the latter. If you don't know what else the flag would do then we don't need it. This is good

projkovTT (approved):

PR #1448, opened 2026-08-25
lib: bh_arc: add DVFS counters to characterize telemetry interval

kaiqihuTT:
The telemetry pass shares the system work queue with the 1 ms DVFS loop, so shortening the telemetry update interval eats into the time available for frequency and voltage control. This adds the instrumentation needed to find how short that interval can safely get.

Three values are exposed through a new DVFS counter bank (COUNTER_BANK_DVFS under TT_SMC_MSG_COUNTER): 
1. ticks dropped because the previous pass was still queued
2. the longest gap between passes
3. the longest pass duration. 

The interval itself is now settable at runtime via TT_SUB_MSG_SET_TELEMETRY_UPDATE_INTERVAL (0x5), so a sweep needs no reflash. 'tt counter' and 'tt counter_clr' print a single decimal value rather than the full response dump.

Tested with twister on native_sim (83/83 passing) for the interval message, and manually on a p150b through tt-fw-terminal: setting intervals with 'tt msg 0x5c6 <ms>', reading them back with 'tt telem 5', and clearing and reading the counters across runs.

### DIFF
--- a/include/tenstorrent/msgqueue.h
+++ b/include/tenstorrent/msgqueue.h
@@ -76,6 +76,35 @@ enum counter_cmd {
 enum counter_bank {
 	/** @brief Throttler counters indexed by @ref aiclk_arb_max */
 	COUNTER_BANK_THROTTLERS,
+	/** @brief DVFS loop counters indexed by @ref dvfs_counter_index */
+	COUNTER_BANK_DVFS,
+};
+
+/** @brief Counter indices within @ref COUNTER_BANK_DVFS */
+enum dvfs_counter_index {
+	/** @brief DVFS timer ticks that expired while the previous pass was still queued.
+	 *
+	 * Such a tick is coalesced into the pending one rather than running on its own, so the
+	 * DVFS loop skips a cycle. A non-zero value means something is occupying the system work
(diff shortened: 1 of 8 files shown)


projkovTT:
What results did characterization yield? Is 1 ms feasible

jgrowdenTT (dismissed):

jgrowdenTT:
@alayander could you review? Thanks.

alayander (dismissed):
@kaiqihuTT IIRC were you doing some validation of this with PPA right? Has that activity happened/been successful?

jgrowdenTT:
@kaiqihuTT could you rebase and resolve the conflicts please? Thanks man.

kaiqihuTT:
> @kaiqihuTT IIRC were you doing some validation of this with PPA right? Has that activity happened/been successful?

I did a sync with Mai on this and I guess she will use this test when the PPA team needs to further reduce the interval. I think it was 20ms the last time I heard from them.

kaiqihuTT:
> @kaiqihuTT could you rebase and resolve the conflicts please? Thanks man.

yessir. I'm on it

alayander (approved):

PR #1449, opened 2026-08-25
D2d fw load

alexapostolu:
Bring a D2D tile's Rocket CPU up on a supplied image:

  1) Hold the D2D core in reset
  2) Clear SRAM
  3) DMA the D2D FW to SRAM, falling back to CPU stores
  4) Verify the bytes
  5) Write the configuration block

The firmware image comes from the SiVal drop and is closed source, so
it is not in this repository.

---

Tests:

Ran on Mimir emulation


tests/drivers/tt_d2d/src/chip_init.c
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright (c) 2026 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+ * Mirrors the bring-up the emulation tests perform before touching a D2D tile:
```
jgrowdenTT, line 8: Nit, can probably do without the first line. 

Chippy could/should be sival drop since that's the blob we'll be unpacking into the build folder

drivers/misc/tt_d2d/d2d_fw_blob.S
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) 2026 Tenstorrent AI ULC
```
jgrowdenTT, line 2: I think this needs to go and get replaced with something that just acts as a ptr provided. I'm thinking CB, dts phandle, something of the ilk.

anything in this folder is "prod" and "prod" shouldn't embed the D2D FW inside the SMC applciation image.

alexapostolu:
Sorry I didn't realize I made a branch on tt-sys-fw instead of my fork, here's a new PR with James's suggestions applied https://github.com/tenstorrent/tt-system-firmware/pull/1453/

PR #1453, opened 2026-08-27
D2d fw load

alexapostolu:
Bring a D2D tile's Rocket CPU up on a supplied image:

  1) Hold the D2D core in reset
  2) Clear SRAM
  3) DMA the D2D FW to SRAM, falling back to CPU stores
  4) Verify the bytes
  5) Write the configuration block

The firmware image comes from the SiVal drop and is closed source, so
it is not in this repository.

---

Tests:

Ran on Mimir emulation

### DIFF
--- a/drivers/misc/tt_d2d/tt_d2d.c
+++ b/drivers/misc/tt_d2d/tt_d2d.c
@@ -1,0 +1,462 @@
+/*
+ * Copyright (c) 2026 Tenstorrent AI ULC
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <zephyr/device.h>
+#include <zephyr/drivers/dma.h>
+#include <zephyr/kernel.h>
+#include <zephyr/sys/byteorder.h>
+#include <zephyr/sys/sys_io.h>
+
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(tt_d2d, CONFIG_TT_D2D_LOG_LEVEL);
+
+#include <d2d_api_driver.h>
+#include <d2d_api_general_definitions.h>
+#include <platform.h>
+
+#include <zephyr/drivers/misc/tt_d2d.h>
+
+#define DT_DRV_COMPAT tenstorrent_grendel_d2d
+
+#ifdef CONFIG_DMA
+BUILD_ASSERT(CONFIG_TT_D2D_INIT_PRIO > CONFIG_DMA_INIT_PRIORITY,
+	     "TT_D2D_INIT_PRIO must be higher than DMA_INIT_PRIORITY so the DMA "
+	     "controller is ready when a tile checks for it");
+#endif
+
+/*
+ * Offsets within a D2D tile's register map.
+ */
+#define TT_D2D_OFFSET(addr) ((addr) - D2D_0_REG_MAP_BASE_ADDR)
+
+#define TT_D2D_STRAP_RESET_OFFSET TT_D2D_OFFSET(D2D_0_STRAP_RESET_REG_ADDR)
+#define TT_D2D_CPU_CTRL_OFFSET    TT_D2D_OFFSET(D2D_0_D2D_D2D_SS_ASYNC_CPU_CTRL_REG_ADDR)
+#define TT_D2D_SRAM_OFFSET        TT_D2D_OFFSET(D2D_0_D2D_D2D_ROCKET_REG_MAP_BASE_ADDR)
+
+BUILD_ASSERT(D2D_1_STRAP_RESET_REG_ADDR - D2D_1_REG_MAP_BASE_ADDR == TT_D2D_STRAP_RESET_OFFSET &&
+		     D2D_1_D2D_D2D_SS_ASYNC_CPU_CTRL_REG_ADDR - D2D_1_REG_MAP_BASE_ADDR ==
+			     TT_D2D_CPU_CTRL_OFFSET &&
+		     D2D_1_D2D_D2D_ROCKET_REG_MAP_BASE_ADDR - D2D_1_REG_MAP_BASE_ADDR ==
+			     TT_D2D_SRAM_OFFSET,
+	     "every D2D tile must share one register layout for base-relative offsets to hold");
+
+/*
+ * The firmware's side of the same window, from the drop's D2D API headers: the
+ * addresses there are tile-relative and start at the SRAM, so the two views
+ * have to agree on where that is before any of them can be used as offsets.
+ */
+BUILD_ASSERT(D2D_MEMORY_BASE == TT_D2D_SRAM_OFFSET,
+	     "the D2D API memory map and the register map disagree on the SRAM window");
+
+/*
+ * The config space is an SRAM region for firmware to read once at statup.
+ */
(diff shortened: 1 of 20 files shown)


dts/bindings/misc/tenstorrent,grendel-d2d.yaml
```diff
@@ -0,0 +1,95 @@
...
+      Optional DMA channel used to zero the SRAM window and copy the image
+      into it. Without it the driver writes both with CPU stores, which is
+      correct but issues one bus transaction per word: a 64 KB window plus a
+      27 KB image is around 23000 of them, and under emulation that is slow
+      enough to matter. The reference bring-up uses DMA for exactly this.
+
+  dma-names:
+    description: Name for the channel in dmas. Only the first entry is used.
+
+  fw-max-size:
```
jgrowdenTT: I guess it's fine this is a DT binding (rather than a KCONFIG or even just a regular old define) But, just to ask, are we expecting different blocks to have different FW sizes? My current understanding was that all D2D blocks are using the same FW.

include/zephyr/drivers/misc/tt_d2d.h
```diff
@@ -0,0 +1,89 @@
...
+ * be sequenced against work this driver does not own:
+ *
+ *   1. tt_d2d_reset_release() - deassert the tile's subsystem resets. Must
+ *      happen before the CCE clock switches to the PLL, because the D2D logic
+ *      samples its straps at the clock rate in effect at the time.
+ *   2. tt_d2d_load_fw()       - put firmware in the Rocket's SRAM and fill in
+ *      its configuration block. Leaves the Rocket in reset.
+ *   3. tt_d2d_start()         - release the Rocket so the firmware runs.
+ *
+ * Step 3 is deliberately not folded into step 2. Both ends of a link must be
```
jgrowdenTT, line 24: ... This is maybe going to be a problem for us. I'm still confused that without D2D how K is supposed to handshake with M

_A_ solution is I3C but it's not specified in the boot doc. The specified way (M writes back to K scratch or mbox) doesn't work because D2D aint up. Or I'm missing something fundamental, which is honestly likely, that saves us. 

So maybe we just lean on the fact that this happens early and M side just loads, and K side just OCCPed M to start executing. And that's "close enough"?

IDK.

tests/drivers/tt_d2d/src/chip_init.c
```diff
@@ -0,0 +1,83 @@
...
+#define FILTER_CFG_READ_EN   BIT64(0)
+#define FILTER_CFG_WRITE_EN  BIT64(1)
+#define FILTER_CFG_ADDR_MODE BIT64(4)
+#define FILTER_CFG_ALLOW_NS  BIT64(8)
+#define FILTER_CFG_BURST     BIT64(24)
+
+/* 7 selects the 128-bit data bus. */
+#define FILTER_CFG_DATA_WIDTH_128 (7ULL << 12)
+
+static void filter_open(uintptr_t filter_base, bool allow_non_secure)
```
jgrowdenTT: @projkovTT Is this related ot the filter driver you made?

Prefer one or the other go in, and the second merge uses the zephyr driver for the work

jgrowdenTT (dismissed):
Looks good overall. Some cleanup work and probably some synchronization needed in conjunction with @projkovTT's PR for the filters (unless I'm wrong about that ;)

@projkovTT appreciate if you could work with @alexapostolu to make sure driver-side logic is using the correct APIs. I'll wait for your green light before merge. Thanks.

tests/drivers/tt_d2d/src/chip_init.c
projkovTT: Yea I'm adding a filter driver in https://github.com/tenstorrent/tt-system-firmware/pull/1458, which I think can be reused in this context

dts/bindings/misc/tenstorrent,grendel-d2d.yaml
```diff
@@ -0,0 +1,95 @@
...
+    type: phandle-array
+    description: |
+      Optional DMA channel used to zero the SRAM window and copy the image
+      into it. Without it the driver writes both with CPU stores, which is
+      correct but issues one bus transaction per word: a 64 KB window plus a
+      27 KB image is around 23000 of them, and under emulation that is slow
+      enough to matter. The reference bring-up uses DMA for exactly this.
+
+  dma-names:
+    description: Name for the channel in dmas. Only the first entry is used.
```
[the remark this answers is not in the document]
alexapostolu: type is inherited

dts/bindings/misc/tenstorrent,grendel-d2d.yaml
alexapostolu: Yea I think you're right, I renamed it to sram size so it describes the tile rather than the image

alexapostolu:
Talked to @projkovTT , will hold off on the filters for now until we determine whether to go with SiVal's filter implementation (function for specific filters) or ours (functions for arbitrary filters, you can pass in any address range)

tests/drivers/tt_d2d/src/chip_init.c
projkovTT: I'm redoing my filter driver to make more use of the SiVal implementation - essentially just a wrapper with some init funcs. So you won't be able to use my driver without SiVal.

It's fairly benign duplication in just the test. Once it's in, SiVal funcs should work in native sim since at that end of the day it just calls `write64_reg`.

drivers/misc/tt_d2d/tt_d2d.c
```diff
@@ -0,0 +1,456 @@
...
+#endif
+
+/*
+ * Offsets within a D2D tile's register map.
+ *
+ * The Rocket's SRAM is a window inside the tile rather than a separate region,
+ * and the firmware configuration block is not a register block at all: it is
+ * an area near the top of that SRAM which the firmware reads once at startup.
+ * Keeping the configuration offset SRAM-relative is what makes the collision
+ * with the image (see TT_D2D_IMAGE_MAX below) obvious rather than a surprise.
```
projkovTT: I'm a bit confused by this comment.
First nit I think SMC, K/M-SMC if possible is better to use than Rocket.
This is describing the SMC SRAM state - that D2D FW lives in SRAM beside the BL1 image?
I'm confused by "the firmware configuration block is not a register block at all", is it just describing the D2D config map?

drivers/misc/tt_d2d/tt_d2d.c
```diff
@@ -0,0 +1,456 @@
...
+ */
+#define TT_D2D_STRAP_RESET_OFFSET 0x100000U
+#define TT_D2D_CPU_CTRL_OFFSET    0x1800U
+#define TT_D2D_SRAM_OFFSET        0x2000U
+#define TT_D2D_CFG_OFFSET         0xFCC0U
+
+/* Anything at or past the configuration block would be overwritten by it. */
+#define TT_D2D_IMAGE_MAX TT_D2D_CFG_OFFSET
+
+/* Word indices within the configuration block. */
```
projkovTT: I think these defines are redundant, could instead use SiVal enum tCONFIG_PARM included in `d2d_api_driver.h`

projkovTT (dismissed):

jgrowdenTT (dismissed):

jgrowdenTT (approved):

projkovTT (approved):

PR #1460, opened 2026-08-28
github: ci: bump metal-test images to v0.76.0

projkovTT:
Update upstream-tests-bh* to v0.76.0 so metal CI runs against a Metal release that includes the current BH test trees (including the deployment suite a follow-on job will reuse).

Prepatory for https://github.com/tenstorrent/tt-system-firmware/pull/1432

## Tests
metal CI on this PR

### DIFF
--- a/.github/ci_boards.json
+++ b/.github/ci_boards.json
@@ -10,7 +10,7 @@
     {
       "board": "p100a",
       "supports_smoke": true,
-      "metal-image": "ghcr.io/tenstorrent/tt-metal/upstream-tests-bh:v0.74.0",
+      "metal-image": "ghcr.io/tenstorrent/tt-metal/upstream-tests-bh:v0.78.0-dev20260827-35-g31a9a54154d",
       "metal-target": "blackhole",
       "arch-name": "blackhole",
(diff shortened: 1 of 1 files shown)


jgrowdenTT:
It's a no from me, til stuff's green.

alexapostolu:
Could you grab containers by the "release-latest" tag instead of grabbing it my numbers? I don't think Metal guarantees that a number like 0.76 is tested and working, but they guarantee that the tag "release-latest" is working.

alexapostolu (approved):

jgrowdenTT:
Thanks @alexapostolu 


@projkovTT  - What's up with the p300/galaxy failures? Are these regressions or do we see these nightly?

projkovTT:
> Thanks @alexapostolu
> 
> @projkovTT - What's up with the p300/galaxy failures? Are these regressions or do we see these nightly?

~GCC / glibc version mismatches I'm looking at a more robust fix than our current just upgrade to gcc-13, cause it looks like it errors out on minor version differences too~
edit: the minor version is just a warning, problem is glibc version is wrong for objtool

PR #1399, opened 2026-08-05
boards: tt_blackhole: add galaxy_bin6 board revision

kaiqihuTT:
Add the galaxy_bin6 revision for bin6 UBB parts, which permit one harvested DRAM instance. The SPI config tables are copied from galaxy_revc with dram_disable_count set to 1 and a distinct board_id (0x47000000000).

Registers the revision with the board, the CI board list and rev2board.sh, adds the SMC overlay that drops the DMC bootfs nodes and a GDDR parameter blob, and extends the recovery chip-info UBB and serdes-count cases.

PR #1463, opened 2026-08-31
boards: tt_blackhole: add galaxy_bin6 board revision

alayander:
Continuation of #1399.

Add the galaxy_bin6 revision for bin6 UBB parts, which permit one harvested DRAM instance. The SPI config tables are copied from galaxy_revc with dram_disable_count set to 1.

Registers the revision with the board, the CI board list, rev2board.sh, the sysbuild galaxy check, the tt_boot_fs bmfw exclusion and the CI DMC signature check, and extends the recovery chip-info UBB and serdes-count cases. SMC overlay reuses revc blobs.

### DIFF
--- a/include/zephyr/drivers/misc/bh_fwtable.h
+++ b/include/zephyr/drivers/misc/bh_fwtable.h
@@ -40,6 +40,7 @@ const struct _ReadOnly *tt_bh_fwtable_get_read_only_table(const struct device *d
 #define BOARDTYPE_P300A 0x45
 #define BOARDTYPE_P300C 0x46
 #define BOARDTYPE_UBB   0x47
+#define BOARDTYPE_GALAXY_BIN6 0x02 /* 00-00202-1-xxxxxxxx */
 
 typedef enum {
 	PcbTypeOrionSLT = 0,
--- a/.github/boards.json
+++ b/.github/boards.json
@@ -9,5 +9,6 @@
   "galaxy",
   "galaxy_2",
   "galaxy_revc",
+  "galaxy_bin6",
   "orion_slt"
 ]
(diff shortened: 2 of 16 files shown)


alayander:
Confirmed that a built bundle from this branch behaves as expected:

1. Flashed on galaxy system with `bootstrap.py` and new board ID
2. On reset all chips in the system boot with 7 GDDR tiles enabled
3. Can use tt-flash to flash to another bundle with these changes: https://github.com/tenstorrent/tt-flash/pull/113

include/zephyr/drivers/misc/bh_fwtable.h
```diff
@@ -40,6 +40,7 @@ const struct _ReadOnly *tt_bh_fwtable_get_read_only_table(const struct device *d
 #define BOARDTYPE_P300A 0x45
 #define BOARDTYPE_P300C 0x46
 #define BOARDTYPE_UBB   0x47
+#define BOARDTYPE_GALAXY_BIN6 0x02 /* 00-00202-1-xxxxxxxx */
```
afongTT, line 43: Sorry, I'm a little late, but this should be 0x202. And you'd need to increase the return type of `get_board_type`. Technically it can be 5 hex digits, so really should be `uint32_t`

projkovTT:
Cursor noted missing `boards/tenstorrent/tt_blackhole/tt_blackhole_tt_blackhole_smc_galaxy_bin6.yaml` for Twister. Fine now but I think we'll need it for testing once the board is in

projkovTT (approved):

include/zephyr/drivers/misc/bh_fwtable.h
alayander: I'll create a follow up PR

PR #1472, opened 2026-09-02
drivers: memc: add mimir gddr memc driver

alayander:
Add driver for mimir gddr memc as well as simple testsuite to ensure gddr init is successful. Building and testing this image requires having the sival sdk tar locally.

Tests:

- Testsuite passes in both emulation and simulation platforms

Resolves: SYS-4940


drivers/memc/gddr_env.c
alayander, line 1: Should we instead be compiling source files given by the drop?

jgrowdenTT: Anything you need a symbol for.... But, those objects should be compiled inside another .a though, not by compiling the source file for the symbols here inside the driver.

Reason being, other drivers/libs might need those symbols, and we don't wanna run into a muldefs issue if they also recompile it.

jgrowdenTT: Probably a good spot is the gr_smc lib actually.

alayander:
Will keep this in draft until sim runs finish

drivers/memc/memc_tt_mimir.c
```diff
@@ -0,0 +1,81 @@
...
+			     fw_params_t *params)
+{
+	if (dev == NULL || backend == NULL || backend->get_blob == NULL || params == NULL) {
+		return -EINVAL;
+	}
+
+	struct memc_tt_mimir_data *data = dev->data;
+
+	params->chiplet.mimir_smc.gddr.gddr_tile_mask = MEMC_GDDR_TILE_MASK;
+
```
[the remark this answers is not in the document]
alayander: Should be 1 driver per mimir SMC

lib/tenstorrent/gr_smc/grendel_utils.c
```diff
@@ -0,0 +1,40 @@
...
+ * (read_cycle, wait_loop_itr) the prebuilt driver archives (gddr, eth)
+ * reference but do not define.
+ */
+
+#include <stdint.h>
+
+#include <zephyr/kernel.h>
+
+uint64_t read_cycle(void);
+void wait_loop_itr(uint32_t iterations);
```
projkovTT, line 15: What's this library for? It doesn't seem to get used. For future work?

projkovTT: Oops didn't read the comment, embarassing

drivers/memc/memc_tt_mimir.c
```diff
@@ -0,0 +1,81 @@
...
+		return -EIO;
+	}
+
+	LOG_INF("GDDR bring-up complete (tile_mask=0x%x)", (unsigned int)gddr_get_tile_en());
+	return 0;
+}
+
+static struct memc_tt_mimir_data memc_tt_mimir_data0;
+
+DEVICE_DT_INST_DEFINE(0, memc_tt_mimir_init, NULL, &memc_tt_mimir_data0, NULL, POST_KERNEL,
```
projkovTT, line 80: For my understanding, we will need another init hook which calls `memc_tt_mimir_set_config` after filling in the config struct (like in tests `mimir_gddr_config.c`). Where will we get the finalized config?

projkovTT: For when that comes, [deferred init](https://docs.zephyrproject.org/latest/kernel/drivers/index.html#deferred-initialization) also gives us the option to init in application if that is what we need.

alayander: Yup, right now that hook doesn't exist - for the production flow, the params will be loaded into SRAM at a fixed address via D2D. Once that is done (potentially notified via an mbox prod), we would use an init hook that will fill in the config struct.

projkovTT (approved):

alexapostolu (approved):

PR #1474, opened 2026-09-04
boards: tenstorrent: mmk: explicit sival_shim kconfig

jgrowdenTT:
### Overview:

Select the SIVAL_SHIM explicitly with a KCONFIG that is (for now) default true at the board level.

This allows tt_mmk/tt_mimir to run with the sival shim also.

### Tests:

This builds my sysbuild with the shim.


alayander:
Can be applied here as well: https://github.com/tenstorrent/tt-system-firmware/blob/main/drivers/gpio/gpio_tt_grendel.c#L20

alayander (approved):

alexapostolu (approved):

projkovTT (approved):

PR #1477, opened 2026-09-04
soc: tt_keraunos: Move SiVal register aliases into tt_grendel_shim.h

projkovTT:
## Overview
Move the SiVal typedef shims into `tt_grendel/tt_keraunos/tt_shim.h`. Configure the inclusion with `CONFIG_SIVAL_SHIM`

## How the register includes work:
- `#include platform.h` includes SiVal `platform.h`
- SiVal `registers/platform.h` is multiplexed via `PLATFORM_KER_SMC`, `PLATFORM_MIMIR_SMC`, ... to include keraunos, mimir, ... registers `ker_platform.h`, `mimir_platform.h`, ...
- Then `#include tt_grendel_shim.h` which just typedef aliases SiVal registers

The reason we need the shims for Keraunos is because `tt_mimir_tt_mimir` builds against the old SiVal drop `grendel-sival-sources.tar.gz` which uses the shortened names, whereas Keraunos (and `tt_mmk_tt_mimir`) uses the new SiVal drop. The aliasing is needed for code portability.

## Tests
Build sysbuild, bl0p5, bl1


jgrowdenTT (changes requested):
Keraunos and Mimir for mmk should both be shimming out. 

Platform.h selection based on include order is straight hacks lol. Though platform include selection based on directory is less hacks. 

Net net, I think we should take this offline. We need to find something that works for everybody. I wouldn't have shipped the shim if it didn't work for me locally but I guess that's causing issues for others.

drivers/dma/dma_tt_grendel_smc.c
```diff
@@ -13,6 +13,9 @@
 #include <zephyr/sys/sys_io.h>
 
 #include <platform.h>
+#ifdef CONFIG_SIVAL_SHIM
+#include "tt_grendel_shim.h"
```
jgrowdenTT, line 17: if you put the ifdef on the inside you don't need it on the outside everywhere...

jgrowdenTT (approved):

drivers/dma/dma_tt_grendel_smc.c
projkovTT: Yea holdover from when the shim was Keraunos only, I was worried about it not being on the include path for mimir builds

alexapostolu (approved):

PR #1478, opened 2026-09-04
drivers: misc: add Grendel AXI filter driver

projkovTT:
## Overview
Add filter driver for Grendel SMC
- Wrapper for SiVal firewall driver
Add init hooks, will be used in bl0p5

## Tests
Added ZTESTs, skipped in CI (sival drop). Run manually:
```
west build -p -b native_sim -d build-filter-test tests/drivers/misc/tt_grendel_filter
west build -d build-filter-test -t run
```

@jgrowdenTT  - ttsim MIS full launch plus the native sim test:
```
west build -b native_sim tests/drivers/misc/tt_grendel_filter
west build -t run
```

Emulation filter, pcie, hsio end-to-end in https://github.com/tenstorrent/tt-system-firmware/pull/1481


drivers/misc/tt_grendel_filter/tt_grendel_filter.c
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	fields.read_en = flags.read;
+	fields.write_en = flags.write;
+	fields.addr_mode = 1;
+	fields.allow_ns = flags.nonsecure;
+	fields.allow_burst = flags.burst;
+
+	err = filter_arm((filter_t)bank, entry, hsio_tile, start, end, fields);
+	switch (err) {
```
projkovTT, line 50: This error return code aliasing is a bit silly, I'll add a log instead

projkovTT:
Disabling the unit tests for CI (cause sival drop)

drivers/misc/tt_grendel_filter/tt_grendel_filter.c
projkovTT: Alternatively,
```
if (err) {
    LOG_ERR("tt_grendel_filter_enable: Error: %d (int)err);"
    return -EIO;
}
return 0;
```
I'm not sure what degree of aliasing we really want since there's not much that upstream can really do about it. 100% we want log visibility.

tests/drivers/misc/tt_grendel_filter/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})
+project(tt_grendel_filter)
+
+set(SIVAL_EXTDIR ${CMAKE_BINARY_DIR}/grendel-sival-sdk)
+set(SIVAL_BLOBS_TARBALL ${CMAKE_CURRENT_SOURCE_DIR}/../../../../zephyr/blobs/grendel-sival-sdk.tar.gz)
+set(SIVAL_SDK ${SIVAL_EXTDIR}/grendel_sival_sdk)
+set(SYSFW_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../../../..)
+
+file(MAKE_DIRECTORY ${SIVAL_EXTDIR})
+execute_process(COMMAND tar -xzf ${SIVAL_BLOBS_TARBALL} WORKING_DIRECTORY ${SIVAL_EXTDIR})
```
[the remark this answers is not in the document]
projkovTT: Its skipped by CI, would need the drop when run manually yes

tests/drivers/misc/tt_grendel_filter/CMakeLists.txt
jgrowdenTT, line 1: @projkovTT I'm going to rework this file so that we're not consuming tarball and self providing the inc dirs/sources to link; that should be done in `soc` (or at least "somewhere centralized" which is currenly `soc`) since everyone and their grandmother needs this boilerplate. No sense repeating it.

drivers/misc/tt_grendel_filter/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+zephyr_library_sources_ifdef(CONFIG_SOC_TT_KERAUNOS_SMC tt_grendel_filter_init.c)
+
+if(CONFIG_SOC_TT_KERAUNOS_SMC)
+  set(SIVAL_SDK ${CMAKE_BINARY_DIR}/grendel-sival-sdk/grendel_sival_sdk)
+  # scratch.h pulls fw_postcodes.h, which the SOC include paths do not cover.
+  zephyr_library_include_directories(
+    ${SIVAL_SDK}/common/include
+    ${SIVAL_SDK}/firmware/common/include
+    ${SIVAL_SDK}/firmware/drivers/firewall/include
+  )
```
jgrowdenTT: Going to remove this. This is present in the SOC Cmake

drivers/misc/tt_grendel_filter/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+  zephyr_library_sources(${SIVAL_SDK}/firmware/drivers/firewall/src/firewall.c)
```
jgrowdenTT: Having the source file we're wrapping part of the driver actually looks extremely clean. Unfortunately right now, lots of apps have need of `firewall.c` symbols without going thru the zephyr driver it looks like. So I'm going to put it into lib `gr_smc`

drivers/misc/tt_grendel_filter/Kconfig
```diff
@@ -0,0 +1,19 @@
...
+	bool "Grendel fabric filter (SiVal firewall wrapper)"
+	depends on SOC_TT_KERAUNOS_SMC
+	default y
+	select LOG
+	help
+	  Wrap SiVal firewall.c and program SMC inbound host windows
+	  (scratch, VUART, SEP-safe SRAM, BUN2 staging).
+
+if TT_GRENDEL_FILTER
+
```
jgrowdenTT, line 14: Under this, I'm adding a new KCONFIG for the `filter_init` C source file. That KCONFIG will only be selected in your test, and BL0p5 when we get around to integration with that. 

Reason being - we definitely do not want all Keraunos Grendel apps (mis, Bl1, random tests, etc) to re-init the filters.

drivers/misc/tt_grendel_filter/tt_grendel_filter_init.c
```diff
@@ -0,0 +1,149 @@
...
+	if (bun2_end < TT_BUN2_STAGING_AREA_ADDR) {
+		return 0;
+	}
+
+	/* Staging runs from the BUN2 base up to (not into) BL0P5's ram0 reservation. */
+	return enable_host_window(BUN2_FILTER_ENTRY, TT_BUN2_STAGING_AREA_ADDR, bun2_end,
+				  "BUN2 staging");
+}
+
+SYS_INIT(tt_grendel_smc_filter_pre_pcie_init, PRE_KERNEL_2, 0);
```
jgrowdenTT, line 154: Priorities look iffy but presumably this was all tested in a bunch. I won't worry too too much. Eventually I think we get KCONFIGged prios with build asserts on the order.

This is part of the BL0p5 integration and out of scope of this.

drivers/misc/tt_grendel_filter/tt_grendel_filter_init.c
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	return enable_host_window(SEP_SAFE_FILTER_ENTRY, SMC_CPU_SPM_MEMORY_MEM_BASE_ADDR + sep_off,
+				  SMC_CPU_SPM_MEMORY_MEM_BASE_ADDR + sep_off + sep_size - 1,
+				  "SEP-safe SRAM");
+}
+
+static int tt_grendel_smc_filter_post_pcie_init(void)
+{
+	uint64_t bun2_end = DT_REG_ADDR_U64(RAM0_NODE) - 1;
```
jgrowdenTT, line 142: We can't allow the window past  64KB from the end of SRAM, since it would stomp BL0p5. 

I _think_ BUN2 can take up space used by BL1, so long as the TOC entry overlapping with BL1's execution space is not BL1 itself... Unless we copy backwards, which I'm not really interested in since I wanna DMA.

jgrowdenTT: D'oh. this is good.

tests/drivers/misc/tt_grendel_filter/src/main.c
jgrowdenTT, line 1: This is really cool, thanks for doing this! 

I'm just going to rework this to use FFF rather than custom stub outs.

alewycky-tenstorrent:
If there's no vuart, tt_grendel_smc_filter_pre_pcie_init stops and doesn't set up the SEP-safe and scratch access windows. Is that intended?

tests/drivers/misc/tt_grendel_filter/CMakeLists.txt
jgrowdenTT: Nvm, this is a lie because SOC here is potentially native sim.

This points to a needed piece of improvement in our current build arch: to suppot native sim, the sival drop includes present in `soc` probably have to be driven by the `gr_smc` lib. I'm not yet sure how that marries with the fact that soc itself needs symbols for it's early init stuff.

So, I'm leaving this, and taking this as a piece of feedback to improve our build infrastructure. Thanks Phil <3

jgrowdenTT (dismissed):

drivers/misc/tt_grendel_filter/tt_grendel_filter_init.c
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	ret = enable_host_window(SCRATCH_FILTER_ENTRY, SCRATCH_REG_BASE, SMC_SCRATCH_END,
+				 "scratch");
+	if (ret < 0) {
+		return ret;
+	}
+
+	vuart_start = (uintptr_t)uart_tt_virt_get(vuart);
+	ret = enable_host_window(VUART_FILTER_ENTRY, vuart_start, vuart_start + VUART_SIZE - 1,
```
projkovTT, line 120: Re @alewycky-tenstorrent Yea I think that VUART device ready should only gate VUART filter and not fail the scratch and SEP windows as it does now. My bad, @jgrowdenTT

app/bl0p5/prj.conf
```diff
@@ -26,6 +26,7 @@ CONFIG_LOG_MODE_IMMEDIATE=y
 
 # Disable PMP to save padding space...
 CONFIG_RISCV_PMP=n
+CONFIG_TT_GR_SMC=y
```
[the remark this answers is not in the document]
jgrowdenTT: bl0p5 integration out of scope.

drivers/misc/tt_grendel_filter/tt_grendel_filter_init.c
projkovTT: I run west `build -d <builddir> initlevels` to verify that I get these right. An alternative to build asserts is to make a centralized priority list with all of them

drivers/misc/tt_grendel_filter/tt_grendel_filter_init.c
jgrowdenTT: I'll just base it on the KCONFIG.

We probably need to revist; Tying the vuart window setup to the filter init might not be wise since bl0p5 won't always up vuart and then no one will.

The scope of this initiative file isn't really defined/specced out, but it appears tied to BUN2 staging. So if the scope ends there, If VUART wants a window open, presumably it's on VUART to open the window. Lets figure it out later.

jgrowdenTT (approved):

projkovTT (commented):
LGTM! Thank you @jgrowdenTT

PR #1486, opened 2026-09-08
github: workflows: skip expired artifacts in ci-stats

projkovTT:
## Overview
Loudbox ci-stats has been failing due to expired artifacts. Eg.
https://github.com/tenstorrent/tt-system-firmware/actions/runs/34231674677/job/102079158083
So skip expired and just output a shorter stats csv.

Not unexpected since we haven't run loudbox stress in a while. Perhaps we should remove ci-stats for loudbox entirely? Unless we wish to resume testing loudbox - We may well end up with this job just generating empty CSVs.

Additional changes:
- Skip downloading non-main artifacts (we previously skipped after downloading)
- Add `ci-docs` label to run docs.yml workflow manually

## Tests
Run on this PR with `ci-docs`
Loudbox pass!
https://github.com/tenstorrent/tt-system-firmware/actions/runs/34239450084/job/102105599977?pr=1486

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -71,12 +71,27 @@ jobs:
             console.log(`Total artifacts found: ${artifacts.data.artifacts.length}`);
 
             for (const artifact of artifacts.data.artifacts) {
-              const download = await github.rest.actions.downloadArtifact({
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                artifact_id: artifact.id,
-                archive_format: 'zip'
-              });
+              if (artifact.expired) {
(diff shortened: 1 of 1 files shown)


jgrowdenTT (approved):

.github/workflows/docs.yml
```diff
@@ -6,7 +6,12 @@ name: Documentation
 on:
   push:
     branches: [main]
-  pull_request: {}
+  pull_request:
```
jgrowdenTT, line 9: Just revert this before merge, right?

jgrowdenTT: Oh no sorry, we're now skipping it unless labeled. Got it.

PR #1487, opened 2026-09-08
app: pytest: Reduce expected power delta

jgrowdenTT:
### Overview:

Reduce the expected power delta to 70W
We believe his is because the HIGH power state got lower as a result of the recent CG correction.

### Tests:

This PR


projkovTT (dismissed):

app/smc/pytest/e2e_smoke.py
projkovTT: Also update this comment :)

alayander (approved):

PR #1494, opened 2026-09-09
lib: bh_arc: add GDDR CA latch-and-retest [SYS-5042] 

akataniTT:
## Summary
Add a `gddrcal` SPI-flash partition and SMC latch-and-retest flow for GDDR CA settings. At boot, seed each MRISC params table from the table and enable the post-training margin check; after training, read MRISC telemetry and write changed settings back to flash.

## Why
MRISC 2.17 can take per-controller CA VREF, termination, and driver strength from the params table, check +/-N VREF margin after training, and sweep if that fails. Those results are only useful across reboot if the SMC persists them. One provisioned table image can be written to every ASIC; each SMC reads and updates only its own row (ASIC location mod 8).

## What changed
- New 4 KB `gddrcal` fixed partition at `0x31E000` (`tt_blackhole_fixed_partitions.dtsi`). No `binary-path`: starts erased and is written in the field (or provisioned from characterization).
- `gddr_cal.{c,h}`: v2 table (`GCAL` magic, CRC over `entries[8][8]`). Missing/corrupt/old (v1) tables are treated as empty so the flow re-latches.
- `CONFIG_TT_BH_ARC_GDDR_CA_LATCH` (default `y`): on MRISC params v7+, enable margin check + sweep fallback; if the local `gddrcal` row is valid, seed vref/termination (and driver strength on params v8+). After training, if telemetry `ca_settings_changed` is set, update the cached entry and write the table back.
- `MRISC_INIT_TIMEOUT` raised to 40 s when the config is on, to cover the full 15×4×8 sweep fallback (~28 s worst case on P300). Healthy boots still finish in well under a second.
- GDDR blobs updated to MRISC **2.17** (params v8, telemetry v3). Older MRISC ignores the extra fields.
Companion MRs:
- bh-mrisc: CA margin check + sweep in MRISC 2.17


jgrowdenTT (changes requested):
Build must be fixed before merge. Thanks @akataniTT !

lib/tenstorrent/bh_arc/gddr_cal.h
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * @brief Write the CA calibration table to flash.
+ *
+ * Fills in magic, version, and CRC on the caller's copy before writing, then
+ * reads back and verifies the write.
+ *
+ * @param [in,out] table Table to store; magic/version/crc fields are updated.
+ * @retval 0 on success, negative error code on failure.
+ */
+int gddr_cal_write(gddr_cal_table_t *table);
```
alewycky-tenstorrent: Please take `const gddr_cal_table_t *`

lib/tenstorrent/bh_arc/gddr_cal.h
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright (c) 2025 Tenstorrent AI ULC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _GDDR_CAL_H_
+#define _GDDR_CAL_H_
+
+#include <stdbool.h>
```
alewycky-tenstorrent: bool isn't used in this header.

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -72,6 +73,23 @@ static const struct device *const fwtable_dev = DEVICE_DT_GET(DT_NODELABEL(fwtab
...
+static gddr_cal_table_t gddr_cal_table;
+static bool gddr_cal_valid;
+
+/* Row of the gddrcal table belonging to this ASIC: 0 on single-ASIC boards, 0/1 on P300,
+ * SPI-provisioned ASIC location (mod table size) on Galaxy UBB. Allows one provisioned
+ * table image to be broadcast to every ASIC's flash.
+ */
+static uint8_t GddrCalAsicIndex(void)
+{
+	return tt_bh_fwtable_get_asic_location(fwtable_dev) % GDDR_CAL_NUM_ASICS;
```
alewycky-tenstorrent: asic_location is already UBB-local on galaxies.

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -609,6 +661,89 @@ static int CheckGddrHwTest(void)
...
+
+	/* ca_vrefc_offset, ca_termination_offset, ca_settings_source and ca_settings_changed
+	 * are adjacent uint8_t fields sharing one 32-bit word.
+	 */
+	uint32_t ca_word =
+		MriscL1Read32(gddr_inst, GDDR_TELEMETRY_TABLE_ADDR +
+						 offsetof(gddr_telemetry_table_t, ca_vrefc_offset));
+	uint8_t vrefc_offset = ca_word & 0xff;
+	uint8_t termination_offset = (ca_word >> 8) & 0xff;
+	uint8_t settings_changed = (ca_word >> 24) & 0xff;
```
alewycky-tenstorrent: Please use `FIELD_GET(GENMASK(7, 0), ca_mask)` & similar.

lib/tenstorrent/bh_arc/gddr.h
```diff
@@ -11,6 +11,19 @@
...
+ */
+#define GDDR_TELEMETRY_TABLE_T_VERSION_MIN 2
+
+/* gddr_params_table_t feature_bits fields (bh-mrisc feature_bit_t) */
+#define GDDR_FEATURE_AC_TIMING_OVERRIDE       (1u << 0)
+#define GDDR_FEATURE_CTRL_PHY_PARAMS_OVERRIDE (1u << 1)
+#define GDDR_FEATURE_ENABLE_INLINE_ECC        (1u << 2)
+#define GDDR_FEATURE_CA_SETTINGS_VALID        (1u << 3)
+#define GDDR_FEATURE_ENABLE_CA_MARGIN_CHECK   (1u << 4)
+#define GDDR_FEATURE_ENABLE_CA_SWEEP_FALLBACK (1u << 5)
```
alewycky-tenstorrent: Use `BIT(n)`.

lib/tenstorrent/bh_arc/gddr_cal.c
```diff
@@ -0,0 +1,118 @@
...
+
+	if (!device_is_ready(flash)) {
+		return -ENODEV;
+	}
+
+	table->magic = GDDR_CAL_MAGIC;
+	table->version = GDDR_CAL_VERSION;
+	table->crc = gddr_cal_crc(table);
+
+	rc = flash_erase(flash, GDDRCAL_OFFSET, GDDRCAL_SIZE);
```
alewycky-tenstorrent: The update sequence must be safe against power-loss.

We already do this for ccfgovr, see for example struct ccfgovr_bank_hdr and tt_bh_fwtable_apply_ccfgovr. (But they only implement the read side of the protocol, not the write side.)

alewycky-tenstorrent:
Please reference SYS-5042 (or another ticket) in the title.

projkovTT:
Could you add a description of how this was tested?

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -767,6 +935,10 @@ static uint8_t toggle_gddr_reset(const union request *req, struct response *rsp)
 		return 1;
 	}
 
+	if (LatchGddrCaSettings(gddr_inst)) {
+		WriteGddrCaSettings();
+	}
+
```
alewycky-tenstorrent: This is a test function that we use during stress tests. Do you have a specific need for `WriteGddrCaSettings()` here? Otherwise I'd rather not update the SPI settings here. If it's needed, condition it on the memory test passing.

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -638,10 +800,16 @@ static int gddr_training(void)
...
 				init_errors = true;
+			} else {
+				cal_dirty |= LatchGddrCaSettings(gddr_inst);
 			}
 		}
 	}
 
+	if (cal_dirty) {
+		WriteGddrCaSettings();
+	}
```
alewycky-tenstorrent: This writes CA settings before the memory test has been verified. Does the MRISC FW do enough verification that the memory test is irrelevant for CA?

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -72,6 +74,38 @@ static const struct device *const fwtable_dev = DEVICE_DT_GET(DT_NODELABEL(fwtab
...
 
+BUILD_ASSERT(GDDR_CAL_NUM_CONTROLLERS == NUM_GDDR, "gddrcal table entry count mismatch");
+
+/* Flash-backed per-ASIC, per-controller CA calibration table. Zero (all entries invalid)
+ * when the gddrcal partitions are empty or corrupt; entries get populated by the latch path.
+ */
+static gddr_cal_table_t gddr_cal_table;
+static bool gddr_cal_valid;
+/* This ASIC's row of gddr_cal_table, or -1 if the ASIC location is out of range. */
+static int gddr_cal_asic_row = -1;
```
alewycky-tenstorrent: Try replacing gddr_cal_asic_row with `static gddr_cal_entry_t *gddr_asic_cal` that is set to `&gddr_cal_table.entries[asic_location]` during init. You might also replace `gddr_cal_valid` with `gddr_asic_cal && gddr_asic_cal[gddr_inst]->valid`.

akataniTT:
@projkovTT Please review the following jira ticket: [https://tenstorrent.atlassian.net/browse/SYS-5042]
I uploaded the ticket with test results. 
Essentially we tested the changes on multiple galaxies and confirmed mrisc<-> zephyr changes work

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -72,6 +75,37 @@ static const struct device *const fwtable_dev = DEVICE_DT_GET(DT_NODELABEL(fwtab
...
+ */
+static int GddrCalAsicIndex(void)
+{
+	uint32_t asic_location = tt_bh_fwtable_get_asic_location(fwtable_dev);
+
+	if (asic_location >= GDDR_CAL_NUM_ASICS) {
+		LOG_WRN_ONCE("asic_location %u outside gddrcal table (%u rows); CA seed/latch "
+			     "disabled",
+			     asic_location, GDDR_CAL_NUM_ASICS);
+		return -EINVAL;
```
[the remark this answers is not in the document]
alewycky-tenstorrent: Copilot doesn't have complete context here. We control the bus number in galaxies.

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -225,6 +273,26 @@ static int LoadMriscFwCfg(uint8_t gddr_inst, uint8_t *buf, size_t buf_size, size
...
+		if (IS_ENABLED(CONFIG_TT_BH_ARC_GDDR_CA_LATCH) &&
+		    params_table->params_table_version >= 7) {
+			uint32_t feature_bits = params_table->feature_bits;
+			const gddr_cal_entry_t *entry =
+				gddr_asic_cal ? &gddr_asic_cal[gddr_inst] : NULL;
+
+			feature_bits |= GDDR_FEATURE_ENABLE_CA_MARGIN_CHECK |
+					GDDR_FEATURE_ENABLE_CA_SWEEP_FALLBACK;
+			if (entry != NULL && entry->valid) {
+				params_table->ca_vrefc_offset = entry->ca_vrefc_offset;
```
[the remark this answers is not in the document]
alewycky-tenstorrent: "Only valid is checked before copying a provisioned row into MRISC." That's the *final* check but just to get from SPI into gddr_cal_table in the first place the data had to pass CRC.

alewycky-tenstorrent (dismissed):

lib/tenstorrent/bh_arc/gddr_telemetry_table.h
```diff
@@ -45,6 +45,48 @@ typedef struct {
 	/* 1 if any uncorrected EDC errors on read since reset. */
 	uint8_t uncorr_edc_wr_error;
 	/* 1 if any uncorrected EDC errors on write since reset. */
+	uint8_t ca_vrefc_offset;
```
jgrowdenTT, line 48: Might have been wise to structure this per version so that we get compile time errors if we parse version -> cast to versioned structure type -> access something not present in the version. I don't really mind though, this can wait.

lib/tenstorrent/bh_arc/gddr_cal.c
```diff
@@ -0,0 +1,254 @@
...
+	active_bank = target;
+	active_seq = seq;
+	have_active = true;
+	LOG_DBG("Wrote gddrcal bank %u seq %u", target, seq);
+	return 0;
+}
+
+#else /* !(PARTITION_EXISTS(gddrcal_a) && PARTITION_EXISTS(gddrcal_b)) */
+
+int gddr_cal_read(gddr_cal_table_t *table)
```
jgrowdenTT, line 244: I would have thought we wanted to run into a compile time error if we're trying to call `gddr_cal_read`/`gddr_cal_write` without the partitions set up in the DTS if we turn on the KCONFIG? Or am I misinterpreting the intended scope of the KCONFIG?

lib/tenstorrent/bh_arc/gddr.c
```diff
@@ -609,6 +696,97 @@ static int CheckGddrHwTest(void)
...
+	LOG_INF("GDDR %d CA settings changed: vrefc_offset=0x%x termination_offset=%d "
+		"ocd_pulldown_offset=0x%x",
+		gddr_inst, vrefc_offset, termination_offset, ocd_pulldown_offset);
+	return true;
+}
+
+/* Write the cached cal table to the gddrcal flash partition. */
+static void WriteGddrCaSettings(void)
+{
+	int rc = gddr_cal_write(&gddr_cal_table);
```
jgrowdenTT, line 782: Should this write be compiled in/out based on `TT_BH_ARC_GDDR_CA_LATCH`?

tests/lib/tenstorrent/bh_arc/gddr_cal/src/main.c
jgrowdenTT, line 1: You rule <3

jgrowdenTT (dismissed):
Merge by EOD. Giving some time for additional comments. Thanks!

lib/tenstorrent/bh_arc/gddr_cal.c
akataniTT: The Kconfig is only the SMC seed/latch path, not “MRISC must have gddrcal partitions.” With it off, MRISC still does its own CA sweep and reports telemetry. SMC just doesn’t seed from flash or write back.
If the DTS nodes are missing, gddr_cal_read/gddr_cal_write are stubs that return -ENOSYS and never touch flash, so they can’t corrupt another partition.

lib/tenstorrent/bh_arc/gddr.c
akataniTT: agreed.

akataniTT: @jgrowdenTT Could you please check again? Tried wrapping the functionality around  TT_BH_ARC_GDDR_CA_LATCH

lib/tenstorrent/bh_arc/gddr_cal.c
jgrowdenTT: I'm still not following really. 

> KConfig is only the SMC seed/latch path 

So to rephrase, if the KCONFIG is on, it's doing the flash operations, if it's off it ain't. Aren't `gddr_cal_read`/`gddr_cal_write` the flash-facing operations? As coded it certainly appears their implementation is predicated on the existance of the partitions... So if you turn on the KCONFIG you _better_ have the partitions, otherwise the KCONFIG isn't going to do what you want.

Why are we arguing this is a runtime error vs "thou shalt not make a build with that configuration"? It smells like a misconfiguration to me.

> With it off, MRISC still does its own CA sweep and reports telemetry.

Yep, Agreed. 

> If the DTS nodes are missing, gddr_cal_read/gddr_cal_write are stubs that return -ENOSYS and never touch flash, so they can’t corrupt another partition.

Yeah... I'm not saying that's "bad"; I'm  saying if you turn on the KCONFIG saying "write cal data to flash" and you're not properly configured to do that, I'd rather see the build die with a clear `#error` vs "Let it run and just return some ECs and hope people looking at prints figure it out."

Sorry I'm slow. Thanks for your time explaining.

akataniTT: Okay I understand.  I had read the original comment as a safety question. Kconfig-on without gddrcal nodes is a misconfiguration. I can make this change as well. Thanks for reviewing

boards/tenstorrent/tt_blackhole/tt_blackhole_fixed_partitions.dtsi
```diff
@@ -241,6 +241,25 @@ partitions {
...
+	 */
+	gddrcal_a: partition@31e000 {
+		label = "gddrcala";
+		reg = <0x31e000 DT_SIZE_K(4)>;
+	};
+
+	gddrcal_b: partition@31f000 {
+		label = "gddrcalb";
+		reg = <0x31f000 DT_SIZE_K(4)>;
+	};
```
alewycky-tenstorrent, line 261: Please move these to 0xC10000 and 0xC11000. We're going to write-lock the SPI below 0xC00000.

alewycky-tenstorrent (approved):

PR #1498, opened 2026-09-09
app: bl1: change K_mbox SPA to be remapped.

jgrowdenTT:
### Overview

The former value assumed no remapping of SPA happened between the D2d link between K and M. However, I believe D2D ATT tables are programmed in the sival drop to make the mem region appear just like PCIE sees these SPA addrs.

### Tests:

Full boot in ttsim (with addr-remap hardcoded in the simulator.)


projkovTT (approved):
I shall believe 👍

PR #1503, opened 2026-09-14
BH P150 Qsfp-dd discovery

aperezvicente-TT:
## Summary
- DMC QSFP-DD management over CM2DM (`TT_SMC_MSG_QSFP_MGMT`): presence, inventory, LPMODE, reset, DOM, diagnostic pages.
- Telemetry tag `TAG_QSFP_STATUS`. Host bring-up CLI is intentionally not in this PR (local/internal).
-  is P150A-only. Not in the shared DMC conf, so P300 DMC does not probe cages on `i2c3` (chip1 ARC SMBus / ARP6). Galaxy cages still tbd.
## P150A hardware (management plane only)
STM32 DMC PA7/PB4 = hardware `i2c3` = MCU_I2C0, through FXMA2102 U1 (OE = PD2):
- TCA9554A `0x38–0x3b` = cages A–D sideband
- CMIS EEPROM `0x50` shared; ModSelL selects the cage
- Same bus: SMC SMBus `@0x0a`, board PCA9555 `@0x20`
- Did not auto-drop LPMODE on optics.
## Test plan
- [x] P150A DMC: `CONFIG_TT_QSFP=y`; live cages (optics CMIS 5.0 LowPwr, DACs CMIS 4.0 Ready)
- [x] P300A DMC configure: `CONFIG_TT_QSFP` unset
- [ ] Reviewer: confirm P300/Galaxy images are unchanged on `i2c3`


lib/tenstorrent/bh_arc/cm2dm_msg.c
```diff
@@ -292,14 +293,26 @@ REGISTER_MESSAGE(TT_SMC_MSG_SET_WDT_TIMEOUT, set_watchdog_timeout);
...
 #ifndef CONFIG_TT_SMC_RECOVERY
-	if (size != sizeof(dmStaticInfo)) {
+	/*
+	 * Accept both the current struct and the legacy layout that predates
+	 * qsfp_status. A DMC running older firmware sends the shorter struct;
+	 * still honor its fw versions / timing instead of dropping all DM
+	 * telemetry on a size mismatch. qsfp_status is only present in the
+	 * full-size message.
+	 */
+	bool has_qsfp = (size == sizeof(dmStaticInfo));
```
alewycky-tenstorrent: DMC and SMC FW need to be the same version. Maybe we need better handling of mismatches but we should address that comprehensively. Please don't try to change it now.

lib/tenstorrent/bh_arc/telemetry.h
```diff
@@ -494,12 +494,23 @@ typedef union {
...
+/** @brief QSFP-DD cage discovery status (P150a).
+ *
+ * One byte per cage, byte 0 = cage A ... byte 3 = cage D. Per-cage byte:
+ *   0x00        expander absent (no I2C response)
+ *   0x01        expander present, no module seated
+ *   0x02        module seated, CMIS 0x50 read failed
+ *   0x80 | id   module present; id = CMIS SFF-8024 identifier byte
+ * Populated by the DMC's qsfp_discover(); 0 if CONFIG_TT_QSFP_DISCOVERY is off.
+ */
+#define TAG_QSFP_STATUS 81
```
alewycky-tenstorrent: AI tells me that valid identifiers can have the high bit set, specifically:

> 00h–25h — assigned (0Ch QSFP … 18h QSFP-DD, 19h OSFP, 1Eh QSFP+ w/ CMIS, up to the CDFP/PCIe entries)
> 26h–7Fh — Reserved (never assigned in ~15 years of the spec; still empty)
> 80h–FFh — Vendor Specific

Please move status codes into the reserved space, maybe 0x70-0x7F. It's also reasonable to split ID & status telemetry wordsl

app/dmc/src/qsfp.c
```diff
@@ -0,0 +1,199 @@
...
+ * Per-cage status byte for the packed word returned by qsfp_discover() and
+ * forwarded to the SMC as telemetry TAG_QSFP_STATUS. See qsfp.h.
+ */
+#define QSFP_ST_EXPANDER_ABSENT 0x00
+#define QSFP_ST_NO_MODULE       0x01
+#define QSFP_ST_READ_FAILED     0x02
+#define QSFP_ST_PRESENT_FLAG    0x80 /* OR'd with the CMIS identifier byte */
+
+struct qsfp_cage {
+	const char *name;
```
alewycky-tenstorrent: Please change name to a single char.

scripts/qsfp_telem.py
```diff
@@ -0,0 +1,88 @@
...
+
+Usage: python3 scripts/qsfp_telem.py [--asic-id N] [--raw]
+"""
+import argparse
+import sys
+
+import pyluwen
+
+# RESET_UNIT_SCRATCH_RAM_REG_ADDR(12): publishes &telemetry[0] (see status_reg.h /
+# telemetry.c). Telemetry values are indexed by tag (TELEM_OFFSET(tag) == tag).
```
alewycky-tenstorrent: https://tenstorrent.atlassian.net/wiki/spaces/syseng/pages/666665152/Telemetry

They aren't indexed by tag. (That happens to work right now, but isn't part of the interface.) SCRATCH_RAM[13] -> struct telemetry_table. Then walk telemetry_table.tag_table until you find your tag. The matching entry contains an offset. Your data is at SCRATCH_RAM[12] + offset.

scripts/qsfp_telem.py
alewycky-tenstorrent: This file is almost completely deleted in the next commit. Do not rework changes within a PR - addition only. If you want to keep this commit separate, bring the library code forward into this commit.

lib/tenstorrent/fw_common/msgqueue.c
```diff
@@ -54,6 +54,8 @@
 #define MSG_ERROR_REPLY 0xff
 
 BUILD_ASSERT(sizeof(union request) <= (sizeof(uint32_t) * REQUEST_MSG_LEN));
+BUILD_ASSERT(TT_SMC_MSG_QSFP_MGMT < CONFIG_TT_MSGQUEUE_NUM_MSG_CODES,
```
jgrowdenTT: The precedent this sets isn't great. We add a new msg and either we add another static assert, or we don't and run into the same issue anyways on a different msg code. 

I'd say remove and we can make a ticket to improve the msg num codes in another PR. Open to other's opinions.

lib/tenstorrent/fw_common/Kconfig
```diff
@@ -28,9 +28,10 @@ config TT_MSGQUEUE
 
 config TT_MSGQUEUE_NUM_MSG_CODES
 	int "Number of message codes"
-	default 200
+	default 201
 	depends on TT_MSGQUEUE
 	help
-	  Size of the message handler table (message code range).
+	  Size of the message handler table. Must be greater than the
+	  highest TT_SMC_MSG_* command code (currently 0xC8 / 200).
```
jgrowdenTT: Comment doesn't really scale, I'd rather not update this every time we add a msg.

tests/lib/tenstorrent/bh_arc/src/smbus_target.c
jgrowdenTT, line 1: Awesome. Thanks <3

lib/tenstorrent/bh_chip/bh_chip.c
```diff
@@ -14,7 +14,6 @@
 #include <zephyr/logging/log.h>
 #include <zephyr/sys/clock.h>
 #include <zephyr/drivers/i2c.h>
```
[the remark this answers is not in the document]
jgrowdenTT: Why remove?

include/tenstorrent/tt_smbus_regs.h
```diff
@@ -298,6 +310,20 @@ enum CMFWSMBusReg {
...
+	 *
+	 * Input (28 bytes): a @ref qsfp_mgmt_response. The SMC validates its
+	 * token and returns it to the host request waiting on TT_SMC_MSG_QSFP_MGMT.
+	 */
+	CMFW_SMBUS_QSFP_MGMT_RESPONSE = 0x2B,
+	/**
+	 * @brief Per-cage QSFP-DD discovery status.
+	 *
+	 * Input (4 bytes, little-endian): packed TAG_QSFP_STATUS word (one
+	 * discovery byte per cage). Older SMCs do not implement this command.
```
jgrowdenTT: Good point, but "older" doesn't really help anyone looking at this without snooping git history.

I think it's fine for now. tt-sys-fw should have a follow up ticket to tabulate commands and the versions they were implemented in.

app/dmc/src/qsfp/qsfp.h
```diff
@@ -0,0 +1,81 @@
+/*
```
jgrowdenTT, line 1: I think we should consider making this a zephyr driver, or a bunch of zephyr drivers.

DTS proposal something like:
```
&i2c3 {
    expander_a: gpio-expander@38 {
        ...
    };

    qsfp_dd: qsfp-dd {
        compatible = "tenstorrent,qsfp-dd";
        enable-gpios = <&gpiod 2 GPIO_ACTIVE_HIGH>;

        cage_a: cage-a {
            modprsl-gpios = <&expander_a 0 GPIO_ACTIVE_LOW>;
            intl-gpios = <&expander_a 1 GPIO_ACTIVE_LOW>;
            lpmode-gpios = <&expander_a 2 GPIO_ACTIVE_HIGH>;
            resetl-gpios = <&expander_a 3 GPIO_ACTIVE_LOW>;
            modsel-gpios = <&expander_a 4 GPIO_ACTIVE_LOW>;
        };
...
    };
};
```

Why -

 - individual testing; `native_sim` tests should definitely be possible with this if you hijack the outputs on the `i2c` node you link it under. 
 - Less tying of this code specifically to the p150a; we can port it to different boards with different GPIO pin/i2c bus constraints, etc. just by tweaking the DTS, instead of the hardcodes to i2c3 in source code
 - Potential separation of the bus for the QSFP vs the expander.

app/dmc/src/qsfp/qsfp_bus.c
```diff
@@ -0,0 +1,213 @@
...
+void qsfp_enable_translator(void)
+{
+	const struct gpio_dt_spec oe = GPIO_DT_SPEC_GET(DT_NODELABEL(qsfp_i2c_en), gpios);
+	int ret = gpio_pin_configure_dt(&oe, GPIO_OUTPUT_ACTIVE);
+
+	if (ret != 0) {
+		LOG_ERR("QSFP: failed to enable MCU_CONN_I2C_EN: %d", ret);
+		return;
+	}
+	k_msleep(2);
```
jgrowdenTT, line 52: `2` is per spec, or just SWAG? Do we need this configruable or nah?

lib/tenstorrent/bh_arc/telemetry.h
```diff
@@ -494,12 +494,24 @@ typedef union {
  */
 #define TAG_FLASH_JEDEC_ID 80
 
+/** @brief QSFP-DD cage discovery status (P150a).
```
jgrowdenTT: Remove (P150a); this telemetry entry exists in all boards. 

Is the expectation for the other boards that it returns 0x00? Might want to mention that below. 

ID is the full 4 bits? 7 bits?

This documentation here is the source of truth from a compat standpoint; If we envision additions we might want to reserve other error values so that UMD callers know to be coded in a way so that if they're addressing future FW which gives em a `0x05` error code here, they don't blow up.

jgrowdenTT:
@alewycky-tenstorrent @afongTT appreciate if you could look. Thanks guys <3

lib/tenstorrent/fw_common/msgqueue.c
aperezvicente-TT: Agreed. I’ll drop this QSFP-specific assert. CONFIG_TT_MSGQUEUE_NUM_MSG_CODES still has to stay at 201 so 0xC8 actually registers.

lib/tenstorrent/fw_common/Kconfig
aperezvicente-TT: Fair. I’ll restore a generic help string and leave the default at 201

lib/tenstorrent/bh_chip/bh_chip.c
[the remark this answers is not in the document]
aperezvicente-TT: Not removing string.h, only the duplicate include and stddef.h. 
stddef.h was added solely for offsetof(dmStaticInfo, qsfp_status) when we briefly grew that struct to 28 bytes. 
Once QSFP status moved to CMFW_SMBUS_QSFP_STATUS, offsetof went away. string.h is still the first include

include/tenstorrent/tt_smbus_regs.h
aperezvicente-TT: I can drop “older SMCs” and say that a missing handler NACKs, or just drop it completely

app/dmc/src/qsfp/qsfp.h
aperezvicente-TT: This PR was just for P150A qsfp-dd bring-up. I agree that long term path and i can work on that either on this PR or a future one. A driver split (expander GPIOs vs CMIS vs translator OE) plus native_sim I2C hijack would also replace the CONFIG_TT_QSFP / overlay gate we added so Galaxy/P300 don’t compile this.

app/dmc/src/qsfp/qsfp_bus.c
aperezvicente-TT: SWAG, not CMIS.   2 ms is for the I2C translator, not the module.
Enabling MCU_CONN_I2C_EN takes U1 (FXMA2102) out of high-Z onto the cage bus. 2 ms is just a conservative wait before we talk to the expanders.

What i can do is drop the magic sleep  and treat expander NACK as “not up yet”, then retry qsfp_init_expanders() a few times.

from your previous comment https://github.com/tenstorrent/tt-system-firmware/pull/1503#discussion_r4038362221 DTS could be the right place later, as enable-gpios plus optional enable-delay-us  (enable-delay-us = <2000>) so another board can change it.

lib/tenstorrent/bh_arc/telemetry.h
aperezvicente-TT: I will update this comment as the compatibility contract:
```
- Tag is global; boards that never publish it stay 0.
- Bit 7 = present; bits 6:0 = SFF-8024 identifier (7 bits, not 4).
- 0x00–0x03 are the only defined not-present codes today.
- 0x04–0x7F reserved; hosts should treat unknown values as “not present / unknown”.
```

aperezvicente-TT: How about this?
```
/** @brief QSFP cage discovery status.
 *
 * Present on every board. Remains 0 if the DMC never publishes
 * CMFW_SMBUS_QSFP_STATUS.
 *
 * Packed little-endian: one byte per cage, byte 0 = cage A ... byte 3 = D.
 *
 * Per-cage byte:
 *   bit 7     present flag (QSFP_TELEM_PRESENT_FLAG)
 *   bits 6:0  SFF-8024 identifier when present (7 bits)
 *
 * Not-present codes (bit 7 clear):
 *   0x00  expander absent (no I2C ACK)
 *   0x01  expander present, no module seated
 *   0x02  module seated, identifier read at 0x50 failed
 *   0x03  bus stuck (all four bytes set)
 *   0x04–0x7F reserved. Hosts must treat unknown values as not present.
 **/
```

jgrowdenTT:
Appreciate also if you could get AI to squash the commit history into commits that are more easily digestible for human readers. As is there's a lot of rework PRs that make it hard to review on a per commit basis. Thanks!

jgrowdenTT (approved):

jgrowdenTT:
@alewycky-tenstorrent appreciate any feedback by tomorrow evening, if you have any thoughts. Thanks.

PR #1505, opened 2026-09-16
app: smc: pytest: Remove rescan in arc watchdog test

projkovTT:
Remove problematic rescan_pcie() in arc_watchdog_test which was causing CI stress failures.


jgrowdenTT (approved):

app/smc/pytest/e2e_stress.py
```diff
@@ -95,6 +96,8 @@ def test_arc_watchdog(arc_chip_dut, asic_id):
     Validates that the DMC firmware watchdog for the ARC will correctly
     reset the chip
     """
+    # Temporary fix for stress CI failure
+    test_arc_msg(arc_chip_dut, asic_id)
```
jgrowdenTT, line 100: Good find. Very curious result, but not the first time I've seen something like this and probably not the last.

we'll nee a ticket to follow up.

projkovTT: This is the current, [SYS-5210 Debug CI Stress arc watchdog failures](https://tenstorrent.atlassian.net/browse/SYS-5210), keep open for the debug?

PR #1506, opened 2026-09-16
app: bl0p5: Serdes FW from BUN1

jgrowdenTT:
### Overview
Get Serdes FW from BUN1

Move init seq around to put everything POST_KERNEL so:

A) we get logs
B) we can re-use the main stack

The design pattern of "provide the `extern`" is **not** cleared by sival and might change (e.g. new configuration structure needs to be passed pointers... we call a function to set the pointers... etc.) I expect that to change

point-`extern`-to-stage approach is one of the easier implemetnations to get the sival drop to do the right thing by just modifying structure and not so much logic/functional code.

### Tests:

- This FW builds and executes with a custom sival drop to allow us to provide the externs.
- This FW does the full sysbuild boot (with logs throughout the PCIE stage) in `ttsim`; with host app providing BUN1.TOC[2] entries. Important bits from BL0p5:

```
BL0.5 [00:00:00.009,188] <inf> tt_grendel_filter: scratch window [0x8010100, 0x801017f]
BL0.5 [00:00:00.021,583] <inf> tt_grendel_filter: VUART window [0x8158000, 0x815afff]
BL0.5 [00:00:00.034,361] <inf> tt_grendel_filter: SEP-safe SRAM window [0x8062400, 0x8156fff]
BL0.5 [00:00:00.340,801] <inf> tt_grendel_pcie: link up ltssm=17 gen=0 x0
BL0.5 [00:00:00.352,504] <inf> tt_grendel_filter: BUN2 staging window [0x8066400, 0x814ffff]
BL0.5 [00:00:00.365,726] <inf> tt_grendel_hsio: HSIO0 clocks and filters open
*** Booting Zephyr OS build 6006b4441942 ***
BL0.5 [00:00:00.382,602] <inf> main: Waiting for BUN2 staging
BL0.5 [00:00:06.081,654] <inf> main: Waiting for BUN2 validation
BL0.5 [00:00:06.401,860] <inf> main: Jumping to SMC BL1 at load_addr 0xc013b000
```

- Emulation test:
```
ubuntu@tt-emulation:~/tt-kmd$ sudo ./tools/keraunos_vuart_console k 0
Initial pointers: tx_head=0 tx_tail=705 rx_head=0 rx_tail=0 tx_oflow=0
Keraunos VUART: desc=0x1202158000 tx_cap=11228 rx_cap=1024 version=0x00000000
Ctrl-C to exit.
Press Ctrl-C or Ctrl-a,x to quit
BL0.5 [00:00:00.000,154] <inf> tt_grendel_filter: scratch window [0x8010100, 0x801017f]
BL0.5 [00:00:00.000,181] <inf> tt_grendel_filter: VUART window [0x8158000, 0x815afff]
BL0.5 [00:00:00.000,207] <err> tt_grendel_filter: Invalid SEP-safe SRAM range: offset 0x0 size 0x0
BL0.5 [00:00:00.000,766] <inf> tt_grendel_pcie: link up ltssm=17 gen=6 x4
BL0.5 [00:00:00.000,796] <inf> tt_grendel_filter: BUN2 staging window [0x8066400, 0x814ffff]
BL0.5 [00:00:00.000,825] <inf> tt_grendel_hsio: HSIO0 clocks and filters open
*** Booting Zephyr OS build v4.4.0-4617-g3d6af87a3085 ***
BL0.5 [00:00:00.000,868] <inf> main: Waiting for BUN2 staging
```

The SEP safe error is because BL0 did not execute to set it. (In ttsim, I let BL0 run, and Host-acting-as-SEP waits for SRAM init via scratch, which lets BL0 set the SEP-safe region.)

resolves: SYS-5230


drivers/pcie/tt_grendel_pcie.c
```diff
@@ -45,6 +46,27 @@ static int tt_grendel_pcie_init(const struct device *dev)
...
+
+	/* SERDES FW is always at TOC 2 */
+	const struct fw_bundle_toc_entry *serdes_fw_entry =
+		(toc->image_count > serdes_fw_toc_idx) ? &toc->entries[serdes_fw_toc_idx] : NULL;
+
+	if (serdes_fw_entry == NULL) {
+		LOG_ERR("No SERDES FW image found at TOC[2] in staged bundle %llu",
+			toc->image_count);
+		return -ENOENT;
+	}
```
alexapostolu: This could be refactored to use one less conditional,

```
if (serdes_fw_toc_idx >= toc->image_count) {
        LOG_ERR("No SERDES FW image found at TOC[2] in staged bundle %llu",
                toc->image_count); 
		return -ENOENT;
}

const struct fw_bundle_toc_entry *serdes_fw_entry = &toc->entries[serdes_fw_toc_idx];
```

alexapostolu (dismissed):

drivers/misc/tt_grendel_filter/tt_grendel_filter_init.c
```diff
@@ -151,6 +151,6 @@ static int tt_grendel_smc_filter_post_pcie_init(void)
 				  "BUN2 staging");
 }
 
-SYS_INIT(tt_grendel_smc_filter_pre_pcie_init, PRE_KERNEL_2, 0);
+SYS_INIT(tt_grendel_smc_filter_pre_pcie_init, POST_KERNEL, 10);
 
-SYS_INIT(tt_grendel_smc_filter_post_pcie_init, POST_KERNEL, 0);
+SYS_INIT(tt_grendel_smc_filter_post_pcie_init, POST_KERNEL, 30);
```
alayander, line 156: Maybe a later concern, but we could have these init priorities as Kconfigs or in a grendel equivalent of `include/tenstorrent/sys_init_defines.h`, also adding build asserts for the ordering

alayander (dismissed):

drivers/misc/tt_grendel_filter/tt_grendel_filter_init.c
jgrowdenTT: Good feedback, I'll make a ticket to address it.

drivers/pcie/tt_grendel_pcie.c
jgrowdenTT: Eventually we'll probably want more error checking, but yeah, I think for now simplifying makes sense.

drivers/pcie/tt_grendel_pcie.c
```diff
@@ -45,6 +47,27 @@ static int tt_grendel_pcie_init(const struct device *dev)
 #ifdef CONFIG_TT_GRENDEL_PCIE_HW_INIT
 	const struct tt_grendel_pcie_config *config = dev->config;
 	pcie_config_t cfg;
+	const uint32_t serdes_fw_toc_idx = 2U;
+
+	const struct fw_bundle_manifest *manifest =
+		(const struct fw_bundle_manifest *)TT_BUN1_STAGING_AREA_ADDR;
+	const struct fw_bundle_toc *toc = (const struct fw_bundle_toc *)(TT_BUN1_STAGING_AREA_ADDR +
+									 manifest->payload_offset);
```
projkovTT, line 55: To clarify, we're keeping BUN1 (minus bl0p5) in the same place it was loaded, no moving things around, the manifest/TOC stays in SRAM. I like this from complexity/speed standpoint. Vibes-wise I don't foresee issues with bl0p5 running out of space at runtime because its pretty small.
So I like this.

drivers/pcie/tt_grendel_pcie.c
```diff
@@ -45,6 +47,27 @@ static int tt_grendel_pcie_init(const struct device *dev)
...
+		return -ENOENT;
+	}
+
+	const struct fw_bundle_toc_entry *serdes_fw_entry = &toc->entries[serdes_fw_toc_idx];
+
+	_binary_pcie_serdes_iccm_fw_bin_start =
+		(uint32_t *)(TT_BUN1_STAGING_AREA_ADDR /*Base on SEP Safe start?*/
+			     + manifest->payload_offset + serdes_fw_entry->offset);
+
+	_binary_pcie_serdes_dccm_fw_bin_start = _binary_pcie_serdes_iccm_fw_bin_start + 16384;
```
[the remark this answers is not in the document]
projkovTT: 64 KiB bins is safe assumption, this is the size SiVal packs them to. I also checked the arithmetic cause the 16k confused me lol - the start/end pointers are uint32_t* so it checks out.

projkovTT (approved):

drivers/pcie/tt_grendel_pcie.c
jgrowdenTT: Yeah. Well, technically BL0p5 is still in BUN1 but it has to be copied to it's jump location by SEP-BL1.

For the serdes FW we read from the BUN1 TOC[2] entry and then sival drop can shove it into where it needs to go, as you say. Quick and not that complex, other than some bundle management required here. 

Kinda assumes SEP-BL0 leaves BUN1 around in SRAM, which I _think_ it does . But didn't test. If it doesn't, that would mean SEP-BL1 has to re-grab at least BL0p5 and the SERDES FW from flash, and we're slow again.

PR #1508, opened 2026-09-17
app: smc: pytest: add --skip-flash to skip DUT programming

afongTT:
## Summary
- Add a `--skip-flash` pytest option so SMC e2e smoke can run against firmware already on the card.
- Session bring-up (`_prepare_and_launch_dut`) skips `west flash` of mcuboot-bl2 and `tt-flash` / `dut.launch()`, then still waits for ARC and checks running FW versions.

## Test plan
Tested on p100a:
- [x] `pytest app/smc/pytest/e2e_smoke.py --help` shows `--skip-flash`
- [x] With firmware already flashed, `pytest app/smc/pytest/e2e_smoke.py --skip-flash` does not run `west flash` or `tt-flash` during session bring-up
- [x] Default (no flag) still flashes as before
- [x] Tests that flash as their own scenario (`test_mcuboot`, `test_ccfgovr_bh_mod`) still flash


jgrowdenTT:
Appreciate if you could verify that last test you have listed:

> Tests that flash as their own scenario (test_mcuboot, test_ccfgovr_bh_mod) still flash

afongTT:
> Appreciate if you could verify that last test you have listed:
> 
> > Tests that flash as their own scenario (test_mcuboot, test_ccfgovr_bh_mod) still flash

Sorry, it took longer than expected because I had an old `tt-smi` installed, but manually verified on my p100a.

`syseng-gh-03` p150a machine seemed to be having issues, but eventually passed?

jgrowdenTT (approved):

